PR #39, opened 2017-12-10
fix(backoffice): use ranged wix-storybook-utils version

argshook:
until wix-storybook-utils is version 1, the ^ does not include much

seems like this is the reason why you don't see correct link and import. Weird though that it still doesn't work for you when linking manually

@lbelinsk


argshook:
meh, i'll make it version `1.0.0`, there are more issues with 0.0.x

PR #46, opened 2017-12-11
Refactor packages

amiryonatan:


packages/wix-ui-backoffice/package.json
```diff
@@ -36,8 +36,8 @@
     "babel-polyfill": "^6.26.0",
     "babel-runtime": "^6.22.0",
     "lodash": "^4.17.4",
-    "wix-ui-core": "^1.0.8",
-    "wix-ui-theme": "^1.0.16"
+    "wix-ui-core": "latest",
```
lbelinsk: @amiryonatan we had an issue with Lerna when we use latest.

amiryonatan: I didn't encounter any issues ... What issues did u have ?

lbelinsk:
Can't remember. I think @shlomitc & @ranyitz saw that Lerna bootstrap didn't work well with latest

PR #51, opened 2017-12-13
ToggleSwitch - Adding rootWidth and rootHeight properties to theme

giladartzi:
Should not affect the component by default, lets us use percentage sizing inside other platforms (such as santa)


lbelinsk:
@shlomitc I think this is ok, but since this is a change in a core component, would like your opinion as well

lbelinsk:
Or @argshook

shlomitc:
Looks fine. Is there a difference between `auto` and `initial` css values? is it the same for all default values?

giladartzi:
According to this https://developer.mozilla.org/en-US/docs/Web/CSS/width, the default is the initial value, so they're basically the same. Will change to initial though in order to be more consistent

PR #58, opened 2017-12-14
Button component

D0vydas:


lbelinsk:
@D0vydas @shaulgo I tried to merge this pr but there were too many unresolved conflicts, which were not straight forwards. I had to create 3 different pr's for each package in order to be sure this is getting merged properly.

I pretty much changed only one thing, and it is not to render the Icons inside the `<UIText>`, but outside of it.

I am closing this one, thanks.
What about the second pr? is there something local?

PR #64, opened 2017-12-15
Auto testkit

aarnoldaas:
Best way to understand how DriverParser works probably is to run it in debugger mode and go step by step.
Use run-with-node.js (you need to be at /src/AutoTestKit)
change line 12 to see how it run with other mock testkits

To see errors remove catch: ComponentMetaInfoGetter/index.js line 188


argshook:
🎉  thanks!

i will review... a little later :)

PR #69, opened 2017-12-17
wix-ui-tpa: add README

lbelinsk:


packages/wix-ui-tpa/README.md
```diff
@@ -1 +1,95 @@
...
+          infoText="some information text should be here"
+          />
+        </UI.panelTabs>
+      </UI.appSettings>
+    );
+  }
+}
+```
+
+By looking at the [wix-ui-core docs] (TBD) we can see that Button's theme object contains `fonts`, `backgroundColor` and `color` properties.
```
lbelinsk: @argshook This should be updated when we will have a theme documentation for Button core.
I added this reminder to the guild week note in wix-style-react

image

PR #81, opened 2017-12-19
auto testkit fixes for diff components

argshook:
thanks to @vidmantas0010


shlomitc:
it's been out for a while now, still relevant?

argshook:
closing, PR is old but some parts are still relevant and still on TODO. i'll use what's needed by hand myself.

PR #82, opened 2017-12-20
Added hoist to reflect statics

amiryonatan:


packages/wix-ui-core/src/createHOC/index.tsx
```diff
@@ -1,6 +1,7 @@
 import * as React from 'react';
 import {string} from 'prop-types';
 import * as ReactDOM from 'react-dom';
+const hoistNonReactStatics = require('hoist-non-react-statics');
```
shlomitc: why not import?

packages/wix-ui-jss/src/withClasses.tsx
```diff
@@ -1,6 +1,7 @@
 import * as React from 'react';
 import * as uniqueId from 'lodash.uniqueid';
 import {generateClasses, detachStyleSheetFromDom} from './domStyleRenderer';
+const hoistNonReactStatics = require('hoist-non-react-statics');
```
shlomitc: why not import?

PR #83, opened 2017-12-20
Added popover

amiryonatan:


packages/wix-ui-core/src/components/Popover/Popover.driver.ts
```diff
@@ -0,0 +1,5 @@
+export const popoverDriverFactory = ({element}) => ({
+  exists: () => !!element,
+  isContentVisible: () => window.getComputedStyle(
+    element.querySelector('[data-hook="content"]')).display !== 'none'
```
shlomitc: do not render the popover if should not be displayed

packages/wix-ui-core/src/components/Popover/index.tsx
```diff
@@ -0,0 +1,42 @@
...
+    popoverShown: false
+  };
+
+  static Element = generateDefaultComponent('Popover.Element');
+  static Content = generateDefaultComponent('Popover.Content');
+
+  render() {
+    const {placement, popoverShown, children} = this.props;
+    const childrenObject = {Element: null, Content: null};
+    buildChildrenObject(children, childrenObject);
```
shlomitc: make this function typed instead of mutating the childrenObject

packages/wix-ui-core/src/utils/index.ts
```diff
@@ -0,0 +1,24 @@
...
+      return acc;
+    }
+
+    const name = child.type['displayName'].split('.').pop();
+    acc[name] = child;
+    return acc;
+  }, childrenObject || {});
+};
+
+export const  generateDefaultComponent = displayName => {
```
shlomitc: 1. extra space
1. a more specific name to describe how it extracts children

packages/wix-ui-core/src/components/Popover/Popover.driver.ts
amiryonatan: 👍

PR #85, opened 2017-12-20
ToggleSwitch - Changing inner structure

giladartzi:
Changed the inner structure of ToggleSwitch in order for it to be more customizable (with mainly resize in mind). All changes should be 100% transparent to all current consumers of this component.

Now it is supported for the knob to not be the same height as track - it can be smaller or even larger (needed to implement iOS style toggle switches)


lbelinsk:
Closing this one in favor of a new pr I will add soon

PR #86, opened 2017-12-21
AutoTestKit: nested methods and more

lxgreen:
This PR contains the following AutoTestKit improvements:

1. AutoTestKit parser now uses **babylon** parser internally instead of the default one (esprima). The reason: esprima throws 'Unexpected token' error on es6 property spread (e.g. { ...something })

2. Single export line driver is supported. For example, LanguagePicker.driver contains the only line:

_export {default} from '../IconWithOptions/IconWithOptions.driver';_

In such cases, the method table has additional column 'Origin' that displays the origin of the method.

3. Nested methods are supported. For example, the IconWithOptions.driver returns an object that contains multiple plain objects that nest the methods:

_return {
    driver,
    dropdownLayoutDriver: {...}
  };_

In this case, the method table displays the method names with prepended container name, e.g. "driver.mouseLeave"


shlomitc:
is this PR relevant?

argshook:
yes, leave it for me please

shlomitc:
it's been out for a while now, still relevant?

argshook:
closing, PR is old but some parts are still relevant and still on TODO. i'll use what's needed by hand myself.

PR #89, opened 2017-12-22
ToggleSwitch Accessibility

rokasmik:
https://github.com/wix/wix-style-react/issues/1235


lbelinsk:
@rokasmik missing tests

packages/wix-ui-core/src/components/ToggleSwitch/ToggleSwitch.tsx
```diff
@@ -44,14 +45,33 @@ class ToggleSwitch extends React.PureComponent<ToggleSwitchProps> {
...
+  componentDidMount() {
+    document.addEventListener('keydown', this.listenToSpace);
+  }
+
+  componentWillUnmount() {
+    document.removeEventListener('keydown', this.listenToSpace);
+  }
+
+  listenToSpace = e => {
+    if (e.keyCode === 32 && document.activeElement === this.toggle) {
```
lbelinsk: Create a constant please for readability (32)

packages/wix-ui-core/src/components/ToggleSwitch/ToggleSwitch.tsx
```diff
@@ -44,14 +45,33 @@ class ToggleSwitch extends React.PureComponent<ToggleSwitchProps> {
...
+      e.preventDefault();
+      this.props.onChange(e);
+    }
+  }
+
+  setToggleRef = ref => {
```
lbelinsk: I would drop this method  and will just do it inlined:
ref={ref => this.toggle = ref}

lbelinsk: <label htmlFor={id} className={classes.outerLabel} tabIndex={0} ref={this.setToggleRef}>

packages/wix-ui-core/src/components/ToggleSwitch/ToggleSwitch.tsx
rokasmik: yup, thought about this - will do!

packages/wix-ui-core/src/components/ToggleSwitch/ToggleSwitch.tsx
rokasmik: Arrow function creates a new reference of the function every time the component renders and therefore the ref is being set every render. If you pass the same ref function (as now), it is called only when initialised and as the ref does not change, the handler is not called later from react. generally it is a good practice no not pass arrow functions in jsx props because they will be garbage collected and will make a new instance every time the component renders.

lbelinsk: You are right. This is why I created `wix-ui-theme` in this repo, to make sure people don't pass `theme={theme({skin})}` for example.
Wierd that it is documented like that here: https://reactjs.org/docs/refs-and-the-dom.html
:)

Wo1v3r:
@lbelinsk 
Linking to the other PR: https://github.com/wix/wix-ui/pull/122

Wo1v3r:
- E2E for ToggleSwitch
- ToggleSwitch as Controlled Component in story
- Accessibility

PR #93, opened 2017-12-24
Add tooltip

amiryonatan:


packages/wix-ui-core/src/components/Popover/index.tsx
```diff
@@ -10,35 +11,45 @@ interface PopoverProps {
...
+  Content?: React.SFC;
+};
 
-  static Element = createComponentThatRendersItsChildren('Popover.Element');
-  static Content = createComponentThatRendersItsChildren('Popover.Content');
-
-  render() {
-    const {placement, popoverShown, children} = this.props;
+const Popover: PopoverType = props => {
+    const {placement, popoverShown, children} = props;
```
lbelinsk: rename popoverShown to shown

packages/wix-ui-core/src/components/Popover/index.tsx
```diff
@@ -10,35 +11,45 @@ interface PopoverProps {
...
         }
       </Manager>
-    );
-  }
-}
+  );
+};
+
+Popover.defaultProps = {
+  popoverShown: false
```
lbelinsk: not a must, since in your case undefined will work for you as well, so up to you

packages/wix-ui-core/src/components/Tooltip/Tooltip.spec.tsx
```diff
@@ -0,0 +1,46 @@
...
+  </Tooltip>;
+
+  it('should not display content by default', () => {
+    const driver = createDriver(createTooltip());
+    expect(driver.isContentExists()).toBeFalsy();
+  });
+
+  it('should display content on hover', () => {
+    const driver = createDriver(createTooltip());
+    driver.onMouseEnter();
```
lbelinsk: better name will be mouseEnter since this is a method

PR #103, opened 2017-12-26
ToggleSwitch - Adding styles prop

giladartzi:


lbelinsk:
@giladartzi Not merging since I am waiting for tomorrow to raise this missing feature in the Editor. Will see then after the meeting

PR #117, opened 2017-12-27
<Divider /> Component

liorzisman:
Implementation of a Divider


packages/wix-ui-core/src/components/Divider/Divider.spec.tsx
```diff
@@ -0,0 +1,16 @@
+import * as React from 'react';
+import {dividerDriverFactory} from './Divider.driver';
+import {createDriverFactory} from 'wix-ui-test-utils';
+
+import Divider from './index';
+
+describe.only('Divider', () => {
```
lbelinsk: you forgot the only here :)
You probably don't have the linter configured.
If you don't know how to configure it, we can do that tomorrow, but in the meantime, you can run locally:
`node_modules/.bin/haste lint`

liorzisman:
@lbelinsk Will do those things first thing in the morning

PR #120, opened 2017-12-28
Supporting component props

amiryonatan:


packages/wix-ui-jss/src/withClasses.tsx
```diff
@@ -33,12 +32,12 @@ export function withClasses(CoreComponent, styles) {
     }
 
     render() {
-      const {theme, ...coreProps} = this.props;
       return (
-        <CoreComponent {...coreProps} classes={this.state.classes}/>
+        <CoreComponent {...this.props} classes={this.state.classes}/>
```
lbelinsk: @amiryonatan why? You are now passing the theme inside the core component, he should not be aware of the theme

packages/wix-ui-core/src/components/Tooltip/Tooltip.tsx
```diff
@@ -19,7 +19,7 @@ class Tooltip extends React.PureComponent<TooltipProps, TooltipState> {
 
   static propTypes = {
     /** The location to display the content */
-    placement: string
+    placement: string.isRequired
```
lbelinsk: shouldn't we have default placement?
This should not be mandatory

packages/wix-ui-jss/package.json
```diff
@@ -50,6 +50,7 @@
     "jss": "^9.3.3",
     "jss-preset-default": "^4.0.1",
     "lodash.uniqueid": "^4.0.1",
+    "prop-types": "^15.6.0",
```
lbelinsk: why?

PR #137, opened 2018-01-02
<Input /> - Adding more props

danielstro:
Added some more native props, removed unneeded ones and added events.


packages/wix-ui-core/src/components/Input/Input.driver.ts
```diff
@@ -2,10 +2,18 @@ export const inputDriverFactory = ({element, eventTrigger}) => {
   return {
     exists: () => !!element,
```
lbelinsk: need to have comments here for the autodoc. I guess Amir missed it here so you can at least add to your new methods and Amir could take care of the rest

packages/wix-ui-core/src/components/Input/Input.spec.tsx
```diff
@@ -141,6 +129,80 @@ describe('Input', () => {
     });
   });
 
+  describe('autoFocus prop', () => {
```
lbelinsk: maybe add another test of "should not be focused by default"

packages/wix-ui-core/src/components/Input/Input.spec.tsx
```diff
@@ -141,6 +129,80 @@ describe('Input', () => {
...
+    it('should be called when the input is blurred', () => {
+      const onBlur = jest.fn();
+      const driver = createDriver(<Input onBlur={onBlur}/>);
+
+      driver.blur();
+      expect(onBlur).toBeCalled();
+    });
+  });
+
+  describe('onFocus attribute', () => {
```
lbelinsk: what happens when the Input is disabled?
Maybe we need to ask our UX, Ben Benhorin, but I think we shouldn't allow to get focused on the disabled input. wdyt?

This is like in Material UI: https://material-ui-next.com/demos/text-fields/

So I think it will be great to have a test for "should not get focused by when the input is disabled"

Same for other events such as onKeyDown, onClick...

danielstro: I think that this is a bit of a problem. The cbs will not be called if the comp is disabled however simulate will still call the events if the comp is disabled so we don't have a way to test this.

packages/wix-ui-core/src/components/Input/Input.spec.tsx
danielstro: Added a test.

PR #138, opened 2018-01-02
Added keyboard logic

amiryonatan:


packages/wix-ui-core/src/components/Dropdown/Dropdown.spec.tsx
```diff
@@ -9,7 +9,6 @@ describe ('Dropdown', () => {
   const options = [1, 2, 3, 4, 5].map(x => ({
     id: x,
     value: `value${x}`,
```
lbelinsk: what heppens if value is a custom value (some node with Icon) and it get's selected. What will be displayed in the Input? Maybe the displayname was a good approach, or you can have valueParser func that you get from the user to know waht to display on select

packages/wix-ui-core/src/components/Dropdown/Dropdown.tsx
```diff
@@ -3,18 +3,24 @@ import Popover, {SharedPopoverProps} from '../Popover';
...
+import {Option} from './DropdownContent/DropdownContent';
 import {CLICK, CLICK_TYPE, HOVER, HOVER_TYPE, SINGLE_SELECT, SINGLE_SELECT_TYPE, MULTI_SELECT, MULTI_SELECT_TYPE, SEPARATOR} from './constants';
 
+type DropdownClasses = {
+  targetElement: string;
+};
+
 interface DropdownProps {
-  children: () => React.ReactNode;
+  children: (onKeyDown) => React.ReactNode;
```
lbelinsk: we talked today that maybe having an object is better that only one param

packages/wix-ui-core/src/components/Dropdown/Dropdown.tsx
```diff
@@ -94,7 +105,37 @@ class Dropdown extends React.PureComponent<DropdownProps & SharedDropdownProps,
     }
   }
 
-  _onOptionClick(option, evt) {
+  onKeyDown(evt: React.KeyboardEvent<HTMLElement>) {
```
lbelinsk: Please remember to talk about it with Ben, to make sure we are synced

packages/wix-ui-core/src/components/Dropdown/Dropdown.spec.tsx
amiryonatan: 👍

PR #140, opened 2018-01-02
fix waitForVisibilityOf

yanivefraim:


lbelinsk:
thanks @yanivefraim !
Will merge after it passes. Will keep an eye on wix-style-react in case it breakes

PR #142, opened 2018-01-03
removes DomTestkit and fixes withClasses tests that relied on it

gavrielco:


packages/wix-ui-jss/test/withClasses.spec.tsx
```diff
@@ -39,48 +40,46 @@ describe('withClasses', () => {
...
     expect(wrapper.node.id).toBeDefined();
-    expect(domTestkit.getStyleElementByComponentId(wrapper.node.id)).toBe(getJssStyleElement());
+    expect(sheetMapper[wrapper.node.id].styleElement).toBe(getJssStyleElement());
   });
 
   it('should inject the correct style tag to the DOM', () => {
     wrapper = render(<StyledComponent/>);
-    const domTestkit = new DomTestkit({componentId: wrapper.node.id});
-    expect(domTestkit.getCssValue({className: 'someClass', property: 'color'})).toBe('green');
+    const element = ReactDOM.findDOMNode(wrapper.node);
```
lbelinsk: Since we are using enzyme here, there is no need to add reactDom. Yo can do `wrapper.getDOMNode()`

PR #149, opened 2018-01-05
wix-ui-icons-common: Add new icons 

ronenst:


argshook:
this can be merged but it should be named `wix-ui-icons-common`

argshook:
this package should contain both general and system icons, but system icons should be nested under `system`

general icon `import Icon from 'wix-ui-icons-common/Icon`
system icon `import SystemIcon from 'wix-ui-icons-common/system/SystemIcon';`

ronenst:
I've added the system icons, I was initially reluctant because this way does make it easily accessible and the hope was that by having them embedded internally users would maybe get the hint that they shouldn't be used.

PR #154, opened 2018-01-08
Added `replace-icons` transform

piratus:
Added `replace-icons` transform for `jscodeshift` and a CLI utility that does the sam using regular expressions for languages other than javascript.


shlomitc:
Closing it for now.
I don't think it should be a separate project, but a util inside wix-style-react.
We can keep it as a referance for when we need it.

argshook:
the idea of separate project was to have place for more transforms in future, e.g. codemods for migrating wsr components to wix-ui-backoffice, or migrating breaking component changes and such

PR #156, opened 2018-01-09
migrate all wix-ui to enzyme@3

eranshabi:


packages/wix-ui-test-utils/package.json
```diff
@@ -32,12 +32,15 @@
   "publishConfig": {
     "registry": "https://registry.npmjs.org/"
   },
+  "dependencies": {
```
lbelinsk: why not dev-dep?

packages/wix-ui-theme/package.json
```diff
@@ -68,7 +69,7 @@
       "<rootDir>/node_modules/"
     ],
     "testRegex": "/src/.*\\.spec\\.(ts|tsx)$",
-    "setupTestFrameworkScriptFile": "<rootDir>/jest-setup"
+    "setupTestFrameworkScriptFile": "wix-ui-test-utils/dist/src/jest-setup"
```
lbelinsk: The dependency in wix-ui-test-utils is only because of the jest-setup file. Isn't it better duplicate this file instead of adding a dependency for it?

packages/wix-ui-test-utils/package.json
eranshabi: Because dependent projects use this project jest-setup which uses enzyme-adapter-react-15, so we had to have it installed

packages/wix-ui-theme/package.json
eranshabi: this is how it's done in all other wi-ui projects

lbelinsk: But the other wix-ui-projects actually dependant on wix-ui-test-utils, not only because of the jset-setup file

packages/wix-ui-test-utils/src/testkit-helpers/enzyme.ts
```diff
@@ -10,7 +9,7 @@ export const enzymeTestkitFactoryCreator = driverFactory => ({wrapper, dataHook}
   return driverFactory({element, wrapper, eventTrigger});
 };
 
-export const isEnzymeTestkitExists = (Element, testkitFactory, options = {withoutDataHook: false}) => {
+export const isEnzymeTestkitExists = (Element, testkitFactory, mount, options = {withoutDataHook: false}) => {
```
shlomitc: 1. it's a breaking change so we need to keep a default mount in case we don't pass one. Because once we merge this PR - wix-style-react will fail.

2. following the previous point `mount` should be the last argument

eranshabi: 1. there is a pr for wsr for that
2. but the last argument is optional and mount isn't

lbelinsk: Maybe put it as the last parameter?
Then it won't be a breaking change for sure

eranshabi: and to import enzyme anyway and use its mount when mount wasn't passed? I ask because I deleted the import.

PR #167, opened 2018-01-15
Input with affixes

danielstro:


packages/wix-ui-core/src/components/InputWithAffixes/InputWithAffixes.driver.ts
```diff
@@ -0,0 +1,18 @@
+import {inputDriverFactory} from '../Input/Input.driver';
+
+function getInput(elem) {
+  return elem.getElementsByTagName('input')[0];
```
danielstro: Not sure if this is the best way to get the input but I didn't find a nicer one.

liorzisman: will `elem` have just one child of type `input` for sure?

packages/wix-ui-core/src/components/InputWithAffixes/InputWithAffixes.driver.ts
```diff
@@ -0,0 +1,18 @@
...
+}
+
+export const inputWithAffixesDriverFactory = ({element, eventTrigger}) => {
```
liorzisman: I liked

packages/wix-ui-core/src/components/InputWithAffixes/InputWithAffixes.spec.tsx
```diff
@@ -0,0 +1,52 @@
...
+    describe('No affixes', () => {
+        it('renders with no prefix or suffix', () => {
+            const driver = createDriver(<InputWithAffixes/>);
+
+            expect(driver.exists()).toBeTruthy();
+            expect(driver.hasPrefix()).toBeFalsy();
+            expect(driver.hasSuffix()).toBeFalsy();
+        });
+
+        it('renders the input with the correct props', () => {
```
liorzisman: in this case where you only check for `disabled`, might as well modify the description to `renders a disabled input` or something alike. your call

packages/wix-ui-core/src/components/InputWithAffixes/InputWithAffixes.spec.tsx
```diff
@@ -0,0 +1,52 @@
...
+    it('should allow adding a custom prefix and suffix components', () => {
+        const driver = createDriver(<InputWithAffixes prefix={<div data-automation-id="prefix"/>} suffix={<div data-automation-id="suffix"/>}/>);
+
+        expect(driver.hasSuffix()).toBeTruthy();
+        expect(driver.hasPrefix()).toBeTruthy();
+        expect(driver.getPrefix().querySelector('[data-automation-id="prefix"]')).toBeTruthy();
+        expect(driver.getSuffix().querySelector('[data-automation-id="suffix"]')).toBeTruthy();
+    });
+
+    it('renders the input with the correct props', () => {
```
liorzisman: should specify the input is the internal already implemented input

packages/wix-ui-core/src/components/InputWithAffixes/InputWithAffixes.spec.tsx
liorzisman: should specify the input is the internal already implemented input

packages/wix-ui-core/src/components/InputWithAffixes/InputWithAffixes.spec.tsx
```diff
@@ -0,0 +1,52 @@
...
+        const driver = createDriver(<InputWithAffixes prefix={<div data-automation-id="prefix"/>} suffix={<div data-automation-id="suffix"/>}/>);
+
+        expect(driver.hasSuffix()).toBeTruthy();
+        expect(driver.hasPrefix()).toBeTruthy();
+        expect(driver.getPrefix().querySelector('[data-automation-id="prefix"]')).toBeTruthy();
+        expect(driver.getSuffix().querySelector('[data-automation-id="suffix"]')).toBeTruthy();
+    });
+
+    it('renders the internal Input comp with the correct props', () => {
+        const driver = createDriver(<InputWithAffixes prefix={<div/>} readOnly/>);
```
shlomitc: why is this test needed?

packages/wix-ui-core/src/components/InputWithAffixes/InputWithAffixes.driver.ts
shlomitc: inside the `InputWithAffixes` component, put a `data-hook` attribute on the `Input` component. This is an internal data-hook so no harm is done.
Then, you can easily query it instead of looking for the `input` tag.

packages/wix-ui-core/src/components/InputWithAffixes/InputWithAffixes.driver.ts
```diff
@@ -0,0 +1,18 @@
...
+  const inputElement = getInput(element);
+  const inputDriver = inputDriverFactory({element: inputElement, eventTrigger});
+  const driver = {
+    hasPrefix: () => !!element.querySelector('[data-hook="input-prefix"]'),
+    getPrefix: () => element.querySelector('[data-hook="input-prefix"]'),
+    hasSuffix: () => !!element.querySelector('[data-hook="input-suffix"]'),
+    getSuffix: () => element.querySelector('[data-hook="input-suffix"]')
```
shlomitc: I'm not sure that users will need such functionality in their tests.
Usually, a suffix/prefix will be just an icon.
Your current implementation will return an element, which doesn't quite tell the user what it can do with it.

I prefer drivers to return either primitive values or other component's drivers.

The `has***fix()` functions are ok.

In `wix-style-react`, we supplied the following functionality:
`hasPrefix` `hasPrefixClass`, `prefixComponentExists`
`hasSuffix`, `hasSuffixClass`, `hasSuffixesClass`, `suffixComponentExists`
https://github.com/wix/wix-style-react/blob/master/src/Input/Input.driver.js#L23

Please investigate what is a reasonable behavior when coming to use this driver.

packages/wix-ui-core/src/components/InputWithAffixes/InputWithAffixes.spec.tsx
danielstro: It's not. It's a duplication

packages/wix-ui-core/src/components/InputWithAffixes/InputWithAffixes.driver.ts
danielstro: I agree, I wrote these just so I can test that the prefix is the element I provided to the component. I'm not sure if I have a way to test this otherwise.
Also, I'm not sure I understand the differences between these functions. Why do we need hasPrefixClass and what is the difference between hasSuffixClass and hasSuffixClasses?

PR #178, opened 2018-01-17
add automated storybook for some components (WIP)

shlomitc:


packages/wix-ui-core/src/components/Input/Input.tsx
```diff
@@ -87,14 +91,14 @@ class Input extends React.Component<InputProps> {
...
-    /** Inputs value */
-    value: oneOfType([string, number])
+    type: oneOf(['number', 'text'])
   };
 
   constructor(props) {
     super(props);
-    this.id = uniqueId('Input');
+    this.state = {
+      id: uniqueId('Input')
```
shlomitc: @argshook - For some reason, when the `id` was on the component instance, the AutoExample failed...

packages/wix-ui-core/src/components/Input/Input.tsx
```diff
@@ -87,14 +91,14 @@ class Input extends React.Component<InputProps> {
     /** Standard component tabIndex */
     tabIndex: number,
     /** The type of the input - number / text */
-    type: string,
-    /** Inputs value */
-    value: oneOfType([string, number])
```
shlomitc: @argshook - `oneOfType` is not supported in the AutoExample. Had to change to a regular `string` type

argshook: https://github.com/wix/wix-ui/issues/179

PR #223, opened 2018-02-01
Fixed divider

amiryonatan:


packages/wix-ui-core/src/baseComponents/DropdownOption/index.tsx
```diff
@@ -68,6 +68,14 @@ export const OptionFactory = {
       null,
       value ? () => <Divider>{value}</Divider> : () => <Divider/>);
   },
+  createCustomDivider(divider: React.SFC): Option {
```
shlomitc: should probably add a test for it

PR #225, opened 2018-02-04
Pagination: prioritize first and last page and show more pages during SSR

alisey:


packages/wix-ui-core/src/components/Pagination/Pagination.spec.tsx
```diff
@@ -152,6 +152,16 @@ describe('Pagination', () => {
       });
     });
 
+    it('adds error state to the input with an invalid numeric value after pressing ENTER and removes it on change', () => {
+      const onChange = jest.fn();
+      const pagination = createDriver(<Pagination paginationMode="input" totalPages={3} onChange={onChange} />);
+      pagination.changeInput('4');
+      pagination.commitInput();
+      expect(pagination.getPageInput().className).toMatch('error');
```
JoMarton: consider creating an `hasError` method in the driver to prevent tests relying on implementation details

packages/wix-ui-core/src/components/Pagination/Pagination.tsx
```diff
@@ -216,7 +222,10 @@ class Pagination extends React.Component<PaginationProps, PaginationState> {
         <input
           data-hook="page-input"
           type="number"
-          className={this.props.classes.pageInput}
+          className={classNames({
```
JoMarton: you already have `const {classes} = this.props;` so you can just use `classes.X`. also I think in classNames you can use the syntax `classNames('foo', { bar: true }); // => 'foo bar'`

PR #226, opened 2018-02-04
Hoist methods instead of hoist statics

guybenron:
In wix-ui-core (createHOC) and wix-ui-jss (withClasses)

To expose non-static methods as well.

### DIFF
--- a/packages/wix-ui-core/src/createHOC/createHOC.spec.tsx
+++ b/packages/wix-ui-core/src/createHOC/createHOC.spec.tsx
@@ -1,5 +1,4 @@
 import * as React from 'react';
-import * as ReactDOM from 'react-dom';
 import {createHOC} from './';
 import {mount} from 'enzyme';
 
--- a/packages/wix-ui-core/src/createHOC/createHOC.spec.tsx
+++ b/packages/wix-ui-core/src/createHOC/createHOC.spec.tsx
@@ -11,7 +11,7 @@ describe('createHOC function', () => {
     dataHook?: string
   };
 
-  const render = (Comp: any) => mount(Comp, { attachTo: document.createElement('div') });
+  const render = (Comp: any) => mount(Comp, {attachTo: document.createElement('div')});
 
   // Regular component with state
   class ChildComponent extends React.Component<ComponentProps, {id: string}> {
--- a/packages/wix-ui-core/src/createHOC/createHOC.spec.tsx
+++ b/packages/wix-ui-core/src/createHOC/createHOC.spec.tsx
@@ -64,25 +64,6 @@ describe('createHOC function', () => {
       const wrapper = renderIntoDocument(<HOCComponent dataClass="my-data-class"/>);
       expect(ReactDOM.findDOMNode(wrapper).getAttribute('data-class')).toBe('my-data-class');
     });
-
-    describe('hoisting', () => {
-      it('should hoist static methods', () => {
-        expect(HOCComponent.staticMethod).toBeDefined();
-        expect(HOCComponent.staticMethod()).toEqual('staticMethod');
-      });
-
-      it('should hoist static variables', () => {
-        expect(HOCComponent.staticVariable).toEqual('staticVariable');
-      });
-
-      it('should hoist prototype methods from child to HOC and bind them', () => {
-        const wrapper = renderIntoDocument(<HOCComponent id="some_id"/>);
-        expect(wrapper.unboundMethod).toBeDefined();
-        expect(wrapper.unboundMethod()).toEqual('unboundMethod');
-        expect(wrapper.boundMethod).toBeDefined();
-        expect(wrapper.boundMethod()).toEqual('some_id');
-      });
-    });
   });
 
   describe('Pure components', () => {
@@ -103,24 +84,6 @@ describe('createHOC function', () => {
       expect(ReactDOM.findDOMNode(wrapper).getAttribute('data-class')).toBe('my-data-class');
     });
 
-    describe('hoisting', () => {
-      it('should hoist static methods', () => {
-        expect(HOCComponent.staticMethod).toBeDefined();
-        expect(HOCComponent.staticMethod()).toEqual('staticMethod');
-      });
-
-      it('should hoist static variables', () => {
-        expect(HOCComponent.staticVariable).toEqual('staticVariable');
-      });
-
-      it('should hoist prototype methods from child to HOC and bind them', () => {
-        const wrapper = renderIntoDocument(<HOCComponent id="some_id" />);
-        expect(wrapper.unboundMethod).toBeDefined();
-        expect(wrapper.unboundMethod()).toEqual('unboundMethod');
-        expect(wrapper.boundMethod).toBeDefined();
-        expect(wrapper.boundMethod()).toEqual('some_id');
-      });
-    });
   });
 
   describe('Stateless components', () => {
--- a/packages/wix-ui-core/src/createHOC/createHOC.spec.tsx
+++ b/packages/wix-ui-core/src/createHOC/createHOC.spec.tsx
@@ -1,37 +1,146 @@
 import * as React from 'react';
+import * as ReactDOM from 'react-dom';
 import {createHOC} from './';
+import {renderIntoDocument, findRenderedComponentWithType} from 'react-dom/test-utils';
 import {mount} from 'enzyme';
 
 describe('createHOC function', () => {
 
   type ComponentProps = {
-    children?: any,
+    id?: string,
     dataClass?: string,
     dataHook?: string
   };
 
-  const Component: React.SFC<ComponentProps> = ({children}) => React.createElement('div', {children});
+  // Regular component with state
+  class ChildComponent extends React.Component<ComponentProps, {id: string}> {
+    constructor(props) {
+      super(props);
+      this.state = {id : props.id};
+      this.boundMethod = this.boundMethod.bind(this);
+    }
 
-  const HOCComponent = createHOC(Component);
+    static staticVariable = 'staticVariable';
+    static staticMethod() { return 'staticMethod'; }
+    unboundMethod() { return 'unboundMethod'; }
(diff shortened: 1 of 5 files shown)


packages/wix-ui-core/src/createHOC/index.tsx
```diff
@@ -30,9 +34,14 @@ export const createHOC = Component => {
     }
 
     render() {
-      return <Component {...this.props}/>;
+      // Can't pass refs to stateless components (and also there's nothing to hoist)
+      return isStatelessComponent(Component)
+        ? (<Component {...this.props}/>)
```
shlomitc: what about other static properties like `PropTypes`, `DisplayName`?
Take a look at the `stylable` wrapper and see how we maintain the properties
https://github.com/wix/wix-ui/blob/master/packages/wix-ui-core/src/withStylable/index.tsx#L51

guybenron: There is a specific line in both `createHOC` and `withClasses` that deals with `displayName`. This line existed before the change.

There is nothing else to treat React statics. This was the case before the change as well.

`hoist-non-react-methods` also hoists (non-React) statics.

packages/wix-ui-core/src/createHOC/createHOC.spec.tsx
```diff
@@ -1,37 +1,146 @@
 import * as React from 'react';
+import * as ReactDOM from 'react-dom';
 import {createHOC} from './';
+import {renderIntoDocument, findRenderedComponentWithType} from 'react-dom/test-utils';
```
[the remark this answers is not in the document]
guybenron: Enzyme testing returns ReactWrapper which does not have the hoisted methods. React test utils returns the actual component.
If you can find a way to make it work with Enzyme, please let me know, because I couldn't find any :/

packages/wix-ui-core/src/createHOC/createHOC.spec.tsx
```diff
@@ -1,37 +1,146 @@
...
-    expect(wrapper.html()).toBe('<div>Hello</div>');
+  // Stateless component
+  const StatelessChildComponent = () => (<div>stateless</div>);
+
+  describe('Regular components', () => {
+    const HOCComponent = createHOC(ChildComponent);
+
+    it('should render the wrapped component', () => {
+      const wrapper = renderIntoDocument(<HOCComponent/>);
+      expect(findRenderedComponentWithType(wrapper, ChildComponent)).toBeTruthy();
```
shlomitc: looks redundant test. the next test checks something on the specific component (data-hook), so why do you need it?

packages/wix-ui-core/src/createHOC/createHOC.spec.tsx
```diff
@@ -1,37 +1,146 @@
...
+        expect(HOCComponent.staticMethod()).toEqual('staticMethod');
+      });
+
+      it('should hoist static variables', () => {
+        expect(HOCComponent.staticVariable).toEqual('staticVariable');
+      });
+
+      it('should hoist prototype methods from child to HOC and bind them', () => {
+        const wrapper = renderIntoDocument(<HOCComponent id="some_id"/>);
+        expect(wrapper.unboundMethod).toBeDefined();
```
shlomitc: redundant

packages/wix-ui-core/src/createHOC/createHOC.spec.tsx
```diff
@@ -1,37 +1,146 @@
...
+        expect(wrapper.unboundMethod()).toEqual('unboundMethod');
+        expect(wrapper.boundMethod).toBeDefined();
```
shlomitc: redundant

packages/wix-ui-core/src/createHOC/createHOC.spec.tsx
```diff
@@ -1,37 +1,146 @@
...
   });
 
-  it('should place data-hook on the root of the component', () => {
-    wrapper = render(<HOCComponent dataHook="my-data">Hello</HOCComponent>);
-    expect(wrapper.getDOMNode().getAttribute('data-hook')).toBe('my-data');
+  describe('Pure components', () => {
+    const HOCComponent = createHOC(PureChildComponent);
+
+    it('should render the wrapped component', () => {
+      const wrapper = renderIntoDocument(<HOCComponent />);
```
shlomitc: redundant test

packages/wix-ui-core/src/createHOC/index.tsx
```diff
@@ -30,9 +34,14 @@ export const createHOC = Component => {
...
+        : (<Component ref={ref => this.wrappedComponentRef = ref} {...this.props}/>);
     }
   }
 
-  return hoistNonReactStatics(WixComponent, Component, {inner: true});
+  return isStatelessComponent(Component)
+    ? WixComponent
+    : hoistNonReactMethods(WixComponent, Component, {delegateTo: c => c.wrappedComponentRef, hoistStatics: true});
```
shlomitc: Another thing that is missing is a test for the original reason you did this change which is making sure you can use the `ref` of a hoisted component.

packages/wix-ui-core/src/createHOC/createHOC.spec.tsx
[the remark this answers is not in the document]
shlomitc: can this maybe help you?
http://airbnb.io/enzyme/docs/api/ReactWrapper/instance.html

packages/wix-ui-core/src/createHOC/index.tsx
shlomitc: ok, we can skip this for now

packages/wix-ui-core/src/createHOC/createHOC.spec.tsx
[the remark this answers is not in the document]
guybenron: It might. The tests work, so I don't see a reason to alter them now?

packages/wix-ui-core/src/createHOC/index.tsx
guybenron: That's not the original reason. I could access the wrapper component via `ref`, what I wanted was to access methods on the wrapped component like so:

```
const component = <Component/>;
component.func();
```

packages/wix-ui-core/src/createHOC/createHOC.spec.tsx
guybenron: Unclear which test(s) you mean are redundant. I am testing different hoists (static, bound, unbound) on different component types (regular, pure, stateless).

packages/wix-ui-core/src/createHOC/createHOC.spec.tsx
guybenron: Those three (render, data-hook, data-class) were original tests that I did not create myself.

packages/wix-ui-core/src/createHOC/createHOC.spec.tsx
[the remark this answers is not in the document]
shlomitc: Because you added another way to test the component which isn't necessary and makes it more complex to maintain

packages/wix-ui-core/src/createHOC/createHOC.spec.tsx
shlomitc: the render test did something different, it checked for the text value that was displayed:
https://github.com/wix/wix-ui/blob/master/packages/wix-ui-core/src/createHOC/createHOC.spec.tsx#L25

You are now just testing that a component is rendered, which is not a good way in my opinion.
In case that the component wasn't rendered, the test below it would also fail, so no need to keep it.

packages/wix-ui-core/src/createHOC/createHOC.spec.tsx
shlomitc: the expectation is redundant. if it wasn't defined, the line right after it will fail

packages/wix-ui-core/src/createHOC/index.tsx
shlomitc: ok, nevermind

packages/wix-ui-core/src/createHOC/createHOC.spec.tsx
guybenron: The old test checked that the wrapped component was being rendered, by looking for the text value (since it was a simple text component).
The new test checks for the wrapped component by looking for the actual component.

packages/wix-ui-core/src/createHOC/createHOC.spec.tsx
guybenron: Removed all redundant tests.

packages/wix-ui-core/src/createHOC/createHOC.spec.tsx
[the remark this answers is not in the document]
guybenron: Moved back to enzyme, removed react-test-utils

packages/wix-ui-core/src/createHOC/createHOC.spec.tsx
```diff
@@ -1,37 +1,144 @@
 import * as React from 'react';
+import * as ReactDOM from 'react-dom';
```
shlomitc: can be removed

packages/wix-ui-core/src/createHOC/index.tsx
```diff
@@ -1,15 +1,19 @@
 import * as React from 'react';
 import * as ReactDOM from 'react-dom';
```
shlomitc: can be removed

packages/wix-ui-core/src/createHOC/createHOC.spec.tsx
guybenron: Done

packages/wix-ui-core/src/createHOC/index.tsx
guybenron: Can't remove this, actually :/ using findDOMNode in this file

PR #230, opened 2018-02-06
wix-ui-test-utils: add puppeteer support

yanivefraim:


shlomitc:
Please export it also in the `index` file
https://github.com/wix/wix-ui/blob/master/packages/wix-ui-test-utils/src/index.ts

yanivefraim:
Yes, sure, already did... (pushing now)

packages/wix-ui-test-utils/src/testkit-helpers/puppetter.ts
```diff
@@ -0,0 +1,3 @@
+export function protractorTestkitFactoryCreator<T> (driverFactory: (e: any, page: any) => T) {
```
amiryonatan: does page has to be any ? 
is there no more specific type it can be ?

amiryonatan: same for `e`

yanivefraim: fixed, 10x

yanivefraim:
Fixed all issues. Plz mere...

PR #233, opened 2018-02-07
ToggleSwitch - Adding focus and blur methods

giladartzi:
Adding due to product request. This raises the question whether we need to support onFocus and onBlur events as well.

### DIFF
--- a/packages/wix-ui-core/src/components/ToggleSwitch/ToggleSwitch.driver.ts
+++ b/packages/wix-ui-core/src/components/ToggleSwitch/ToggleSwitch.driver.ts
@@ -25,6 +25,8 @@ export const toggleSwitchDriverFactory = ({element, eventTrigger}) => {
     getInnerLabelStyles: () => window.getComputedStyle(element.querySelector('.innerLabel')),
     /** Returns the computed styles object of the toggle icon component */
     getToggleIconStyles: () => window.getComputedStyle(element.querySelector('.toggleIcon')),
+    /** Returns true if the ToggleSwitch is currently focused */
+    isFocused: () => toggleSwitch === document.activeElement,
 
     /** returns elements innerHtml */
     styles: {
--- a/packages/wix-ui-core/src/components/ToggleSwitch/ToggleSwitch.driver.ts
+++ b/packages/wix-ui-core/src/components/ToggleSwitch/ToggleSwitch.driver.ts
@@ -1,4 +1,4 @@
-export const toggleSwitchDriverFactory = ({componentInstance, element, eventTrigger}) => {
+export const toggleSwitchDriverFactory = ({element, eventTrigger}) => {
   const toggleSwitch = element.querySelector('input');
 
   return {
@@ -25,12 +25,6 @@ export const toggleSwitchDriverFactory = ({componentInstance, element, eventTrig
     getInnerLabelStyles: () => window.getComputedStyle(element.querySelector('.innerLabel')),
     /** Returns the computed styles object of the toggle icon component */
     getToggleIconStyles: () => window.getComputedStyle(element.querySelector('.toggleIcon')),
-    /** Returns a boolean indicating whether the toggle switch is focused or not */
-    isFocused: () => toggleSwitch === document.activeElement,
-    /** Invokes the focus method on the toggle switch */
-    focus: () => componentInstance.focus(),
-    /** Invokes the blur method on the toggle switch */
-    blur: () => componentInstance.blur(),
 
     /** returns elements innerHtml */
     styles: {
--- a/packages/wix-ui-core/src/components/ToggleSwitch/ToggleSwitch.driver.ts
+++ b/packages/wix-ui-core/src/components/ToggleSwitch/ToggleSwitch.driver.ts
@@ -1,4 +1,4 @@
-export const toggleSwitchDriverFactory = ({element, eventTrigger}) => {
+export const toggleSwitchDriverFactory = ({componentInstance, element, eventTrigger}) => {
   const toggleSwitch = element.querySelector('input');
 
   return {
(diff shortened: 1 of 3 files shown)


packages/wix-ui-core/src/components/ToggleSwitch/ToggleSwitch.driver.ts
```diff
@@ -25,6 +25,12 @@ export const toggleSwitchDriverFactory = ({element, eventTrigger}) => {
     getInnerLabelStyles: () => window.getComputedStyle(element.querySelector('.innerLabel')),
     /** Returns the computed styles object of the toggle icon component */
     getToggleIconStyles: () => window.getComputedStyle(element.querySelector('.toggleIcon')),
+    /** Returns a boolean indicating whether the toggle switch is focused or not */
+    isFocused: () => element.querySelector('input') === document.activeElement,
```
lbelinsk: You have `toggleSwitch` variable which equals `element.querySelector('input')`, so better to use it

packages/wix-ui-core/src/components/ToggleSwitch/ToggleSwitch.driver.ts
```diff
@@ -25,6 +25,12 @@ export const toggleSwitchDriverFactory = ({element, eventTrigger}) => {
     getInnerLabelStyles: () => window.getComputedStyle(element.querySelector('.innerLabel')),
     /** Returns the computed styles object of the toggle icon component */
     getToggleIconStyles: () => window.getComputedStyle(element.querySelector('.toggleIcon')),
+    /** Returns a boolean indicating whether the toggle switch is focused or not */
+    isFocused: () => toggleSwitch === document.activeElement,
+    /** Invokes the focus method on the toggle switch */
+    focus: () => componentInstance.focus(),
```
shlomitc: I prefer not using `componentInstance` at all.
This function won't work for consumers of the library as we do not get the component instance (for example in the enzyme one).

Can't you just use just native `focus()` and `blur()`?

If not - I prefer that in the spec file you will wrap this component with a dummy component that invokes the direct call to the function (that will actually simulate the real case better).

packages/wix-ui-core/src/components/ToggleSwitch/ToggleSwitch.tsx
```diff
@@ -90,6 +99,7 @@ class ToggleSwitch extends React.PureComponent<ToggleSwitchProps> {
           checked={checked}
           disabled={disabled}
           onChange={e => this._handleChange(e)}
+          ref={el => this.inputRef = el}
```
shlomitc: Wouldn't it be better to expose the input ref and allow the consumer to do that from the outside instead of adding this functionality to the component?

giladartzi:
@lbelinsk - input elements should be able to be focused/blurred programatically

giladartzi:
@shlomitc - exposing the input ref means we expose our inner implementation to the consumer. I think the best practice is considered exposing a focus/blur method, which under the hood deals with the implementation

packages/wix-ui-core/src/components/ToggleSwitch/ToggleSwitch.spec.tsx
```diff
@@ -133,4 +147,30 @@ describe('ToggleSwitch', () => {
...
+
+    it('should expose a focus() method', () => {
+      const RefExpose = getRefExpose(ToggleSwitch, {onChange: () => {}}, refs);
+      const wrapper = mount(<RefExpose />);
+      const input = wrapper.find('input').getDOMNode();
+      expect(input).not.toBe(document.activeElement);
+      refs.childRef.focus();
+      expect(input).toBe(document.activeElement);
+    });
+
```
giladartzi: @shlomitc @lbelinsk what do you think about this kind of solution? calling a dummy component that exposes the ref. Calling the method directly from the wrapper's `componentDidMount` was not enough, because in the blur use case I needed to run focus, assert and then blur. There must be a nicer way though (Maybe get this inside the driver itself?)

shlomitc: I like the Dummy component for testing `ref`s.
Maybe instead of passing a `refs` object, you can just expose a function on the dummy component that allows you to trigger it.

I'm not sure I like the fact that you query the document in the test. Looks like something that should sit in the component's driver in `isFocus()`

packages/wix-ui-core/src/components/ToggleSwitch/ToggleSwitch.spec.tsx
```diff
@@ -133,4 +151,31 @@ describe('ToggleSwitch', () => {
...
+    const focus = () => wrapper.instance().focus();
+    const blur = () => wrapper.instance().blur();
+
+    beforeEach(() => {
+      const RefExpose = getRefExpose(ToggleSwitch, {onChange: () => {}});
+      wrapper = mount(<RefExpose />);
+      const element = wrapper.getDOMNode();
+      driver = toggleSwitchDriverFactory({element, eventTrigger: {}});
+    });
```
giladartzi: Creating driver for the inner component

packages/wix-ui-core/src/components/ToggleSwitch/ToggleSwitch.spec.tsx
```diff
@@ -7,6 +7,24 @@ import {toggleSwitchTestkitFactory as enzymeToggleSwitchTestkitFactory} from '..
 import {activeViewBox, activePathD, inactiveViewBox, inactivePathD} from './utils';
 import {mount} from 'enzyme';
 
+const getRefExpose = (Child, props) => {
+  return class RefExpose extends React.Component {
+    childRef: HTMLElement;
+
+    focus() {
```
giladartzi: Created focus and blur methods, instead of exposing the ref itself

PR #235, opened 2018-02-07
StylableText component

amiryonatan:


packages/wix-ui-core/src/components/StylableText/Text.spec.tsx
```diff
@@ -0,0 +1,57 @@
...
+      expect(driver.exists()).toBeTruthy();
+    });
+
+    it('should have ellipsis', () => {
+      const driver = createDriver(<Text ellipsis>Hello World</Text>);
+      expect(driver.isEllipsis()).toBeTruthy();
+    });
+
+    it('should not have title attribute by default', () => {
+      const driver = createDriver(<Text ellipsis>Hello World</Text>);
```
shlomitc: it looks exactly like the next text. what is the difference?

packages/wix-ui-core/src/components/StylableText/Text.st.css
```diff
@@ -0,0 +1,12 @@
...
+}
+
+/* ellipsis state */
+.root:ellipsis {
+    text-overflow: ellipsis;
+    display: inline-block;
+    width: 100%;
+    overflow: hidden;
+    white-space: nowrap;
+}
```
shlomitc: missing EOL

packages/wix-ui-core/stories/Text/index.ts
```diff
@@ -0,0 +1,20 @@
...
+  category: 'Components',
+  name: 'Text',
+  storyName: 'StylableText',
+  component: Text,
+  componentProps: (setState) => ({
+    'data-hook': 'storybook-text',
+    ellipsis: true,
+    children: 'hello'
+  }),
+  exampleProps: {
```
shlomitc: remove

packages/wix-ui-core/stories/index.tsx
```diff
@@ -24,6 +24,7 @@ const dropdownOptions =
 require('./Badge').story();
 require('./Button').story();
 require('./Input').story();
+require('./Text').story();
```
shlomitc: `StylableText`

packages/wix-ui-core/src/components/StylableText/Text.spec.tsx
shirlev-wix: I guess nothing. Was copied from the old text tests.. Removing..

shirlev-wix: Actually I think this test is needed if i'm reading the description. The test just doesn't match the title.

packages/wix-ui-core/src/components/StylableText/Text.st.css
shirlev-wix: EOF?

shlomitc: yes

PR #241, opened 2018-02-13
<Tooltip /> Core tooltip animations adjustments

liorzisman:


shlomitc:
The storybook is broken, can you please fix it?

PR #243, opened 2018-02-14
Add Video Component

kupriyanenko:


packages/wix-ui-core/src/components/Video/Video.driver.ts
```diff
@@ -0,0 +1,12 @@
+export const videoDriverFactory = (component) => {
+  const {element, componentInstance} = component;
```
shlomitc: I have some concerns regarding using `componentInstance`.
When users use this driver, they get a `vanilla` version of the driver, meaning they don't get the `componentInstace`, thus the driver won't work for them.

We had similar discussions when testing function exposed on a component..
In such cases, we said that users will render their own component that will call internally the `ref.function`.
```
class DummyComp extends React.Component {
	doSomethingOfTestedComponent() {
		this.r.doSomething()
	}

	render() {
		<TestedComponent ref={this.r = r;}/>
	}
}
```

And in the spec, render `DummyComp` and call the exposed function.

How do you expect users to test this component in their app? maybe using the `playableRef`?

packages/wix-ui-core/src/components/Video/Video.driver.ts
```diff
@@ -0,0 +1,12 @@
+export const videoDriverFactory = (component) => {
```
shlomitc: You also need to export the driver to the users:
https://github.com/wix/wix-ui/tree/master/packages/wix-ui-core/src/testkit

Then, make sure that you can use it, like here:
https://github.com/wix/wix-ui/blob/master/packages/wix-ui-core/src/components/Input/Input.spec.tsx#L214-L224

packages/wix-ui-core/src/components/Video/Video.driver.ts
kupriyanenko: can you give me some example of testing with this DummyComp?

shlomitc: something like this:

packages/wix-ui-core/stories/Video/index.tsx
```diff
@@ -0,0 +1,28 @@
...
+    height: 225,
+    title: 'Awesome title',
+    onPlay: () => setState({playing: true}),
+    onPause: () => setState({playing: false}),
+  }),
+  exampleProps: {
+    onPlay: () => 'Triggered onPlay',
+    onPause: () => 'Triggered onPause',
+    onEnd: () => 'Triggered onEnd',
+    playableRef: playable => {/* this.playable = playable */},
```
shlomitc: maybe exposing the ref somehow to the document will help debugging? Instead of just a comment.

packages/wix-ui-core/src/components/Video/Video.spec.tsx
```diff
@@ -0,0 +1,56 @@
+import * as React from 'react';
+import {createDriver} from './Video.driver';
+import {Video} from './';
+
+// dummy polyfil for tests
+global['requestAnimationFrame'] = (cb) => cb();
```
shlomitc: Does it make sense to add it to some test utility that should run for all the tests?

packages/wix-ui-core/src/components/Video/Video.driver.ts
```diff
@@ -0,0 +1,19 @@
+import * as React from 'react';
```
shlomitc: I suggest mentioning that this driver is a private one in order not to confuse the auto-documentation in the future. maybe something like `Video.driver.private.ts`

PR #247, opened 2018-02-15
Stylable badge

lbelinsk:
Changed the api a little bit.
The BO component will not have the "minWidth" as a mixin prop since imo it is too specific to the counterBadge.

The core shouldn't have the specific D20 color in it's palette, so i changed it to be grey.

### DIFF
--- a/packages/wix-ui-core/src/components/StylableBadge/Badge.driver.ts
+++ b/packages/wix-ui-core/src/components/StylableBadge/Badge.driver.ts
@@ -1,0 +1,13 @@
+export const badgeDriverFactory = ({element}) => {
+    return {
+      /** checks if element exists */
+      exists: () => !!element,
+      /** returns elements type attribute */
+      getType: () => element.getAttribute('type'),
+      /** returns elements innerHtml */
+      getContent: () => element.innerHTML,
+      /** returns elements innerText */
+      getContentText: () => element.textContent,
+    };
+  };
+  
--- a/packages/wix-ui-core/src/components/StylableBadge/Badge.driver.ts
+++ b/packages/wix-ui-core/src/components/StylableBadge/Badge.driver.ts
@@ -10,4 +10,3 @@ export const badgeDriverFactory = ({element}) => {
       getContentText: () => element.textContent,
     };
   };
-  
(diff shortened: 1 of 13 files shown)


packages/wix-ui-core/src/components/StylableBadge/Badge.driver.ts
```diff
@@ -0,0 +1,13 @@
...
+      exists: () => !!element,
+      /** returns elements type attribute */
+      getType: () => element.getAttribute('type'),
+      /** returns elements innerHtml */
+      getContent: () => element.innerHTML,
+      /** returns elements innerText */
+      getContentText: () => element.textContent,
+    };
+  };
+
```
shlomitc: need a new line

packages/wix-ui-core/src/components/StylableBadge/Badge.protractor.driver.ts
```diff
@@ -0,0 +1,7 @@
+export const badgeDriverFactory = component => ({
+    /** returns the component element */
+    element: () => component,
+    /** returns the component text */
+    getTextContent: () => component.getText(),
+  });
+
```
shlomitc: new line

packages/wix-ui-core/stories/StylableBadge/index.tsx
```diff
@@ -0,0 +1,16 @@
+import * as React from 'react';
+import {Badge as StylableBadge} from '../../src/components/StylableBadge';
+import commonStyle from '../../src/components/StylableBadge/BadgeStyle.st.css';
+
+export class StylableBadgeStory extends React.Component<{}, {}> {
```
shlomitc: I think that the AutoDocs tool works for it

lbelinsk: I see that it is possible only for the core "reset" look, not for the default look if I am not wrong

shlomitc: I think it’s good to have the core look.
In general, the suggested theme should only be an example and not the main thing.
You can add it as an additional example at the bottom of the automated story

shlomitc: I think it’s good to have the core look.
In general, the suggested theme should only be an example and not the main thing.
You can add it as an additional example at the bottom of the automated story

PR #252, opened 2018-02-19
wix-ui-core: adding TimePicker

guybenron:
Work in progress!

### DIFF
--- a/packages/wix-ui-core/src/components/Input/Input.tsx
+++ b/packages/wix-ui-core/src/components/Input/Input.tsx
@@ -18,6 +18,10 @@ export interface InputProps {
   onBlur?: React.FocusEventHandler<HTMLElement>;
   onChange?: React.ChangeEventHandler<HTMLInputElement>;
   onFocus?: React.FocusEventHandler<HTMLElement>;
+  onMouseDown?: React.EventHandler<React.MouseEvent<HTMLInputElement>>;
+  onMouseUp?: React.EventHandler<React.MouseEvent<HTMLInputElement>>;
+  onMouseMove?: React.EventHandler<React.MouseEvent<HTMLInputElement>>;
+  onDragStart?: React.EventHandler<React.DragEvent<HTMLInputElement>>;
   onKeyDown?: React.KeyboardEventHandler<HTMLInputElement>;
   onKeyPress?: React.KeyboardEventHandler<HTMLInputElement>;
   onKeyUp?: React.KeyboardEventHandler<HTMLInputElement>;
@@ -50,6 +54,10 @@ export class Input extends React.Component<InputProps, InputState> {
     onBlur: PropTypes.func,
     onChange: PropTypes.func,
     onFocus: PropTypes.func,
+    onMouseDown: PropTypes.func,
+    onMouseUp: PropTypes.func,
+    onMouseMove: PropTypes.func,
+    onDragStart: PropTypes.func,
     onKeyDown: PropTypes.func,
     onKeyPress: PropTypes.func,
     onKeyUp: PropTypes.func,
@@ -83,6 +91,10 @@ export class Input extends React.Component<InputProps, InputState> {
       autoComplete,
       autoFocus,
       onChange,
+      onMouseDown,
+      onMouseUp,
+      onMouseMove,
+      onDragStart,
       onKeyDown,
       onKeyPress,
       onKeyUp,
@@ -113,6 +125,10 @@ export class Input extends React.Component<InputProps, InputState> {
           onBlur={this.handleBlur}
           onChange={onChange}
           onFocus={this.handleFocus}
+          onMouseDown={onMouseDown}
+          onMouseUp={onMouseUp}
+          onMouseMove={onMouseMove}
+          onDragStart={onDragStart}
           onKeyDown={onKeyDown}
           onKeyPress={onKeyPress}
           onKeyUp={onKeyUp}
--- a/packages/wix-ui-core/src/components/Input/Input.tsx
+++ b/packages/wix-ui-core/src/components/Input/Input.tsx
@@ -18,6 +18,7 @@ export interface InputProps {
   onBlur?: React.FocusEventHandler<HTMLElement>;
   onChange?: React.ChangeEventHandler<HTMLInputElement>;
   onFocus?: React.FocusEventHandler<HTMLElement>;
+  onClick?: React.EventHandler<React.MouseEvent<HTMLInputElement>>;
   onMouseDown?: React.EventHandler<React.MouseEvent<HTMLInputElement>>;
   onMouseUp?: React.EventHandler<React.MouseEvent<HTMLInputElement>>;
   onMouseMove?: React.EventHandler<React.MouseEvent<HTMLInputElement>>;
@@ -54,6 +55,7 @@ export class Input extends React.Component<InputProps, InputState> {
     onBlur: PropTypes.func,
     onChange: PropTypes.func,
     onFocus: PropTypes.func,
+    onClick: PropTypes.func,
     onMouseDown: PropTypes.func,
     onMouseUp: PropTypes.func,
     onMouseMove: PropTypes.func,
@@ -91,6 +93,7 @@ export class Input extends React.Component<InputProps, InputState> {
       autoComplete,
       autoFocus,
       onChange,
+      onClick,
       onMouseDown,
       onMouseUp,
       onMouseMove,
@@ -125,6 +128,7 @@ export class Input extends React.Component<InputProps, InputState> {
           onBlur={this.handleBlur}
           onChange={onChange}
           onFocus={this.handleFocus}
+          onClick={onClick}
           onMouseDown={onMouseDown}
           onMouseUp={onMouseUp}
           onMouseMove={onMouseMove}
--- a/packages/wix-ui-core/src/components/TimePicker/TimePicker.driver.ts
+++ b/packages/wix-ui-core/src/components/TimePicker/TimePicker.driver.ts
@@ -1,0 +1,43 @@
+import {ComponentFactory} from 'wix-ui-test-utils';
+import {FIELD} from './';
+import {func} from 'prop-types';
+
+// Typescript was yelling at using ComponentFactory<TimePicker> for some reason and this makes it go away
+export interface TimePickerShell {
+  increment: func;
+  decrement: func;
+}
+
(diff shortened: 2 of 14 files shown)


packages/wix-ui-core/src/components/Input/Input.tsx
```diff
@@ -85,7 +85,7 @@ export class Input extends React.Component<InputProps, InputState> {
     /** Standard component tabIndex */
     tabIndex: number,
     /** The type of the input - number / text */
-    type: oneOf(['number', 'text']),
+    type: oneOf(['number', 'text', 'time']),
```
shlomitc: can you add a test for that?

packages/wix-ui-core/src/components/TimePicker/TimePicker.driver.ts
```diff
@@ -0,0 +1,42 @@
...
+    exists: () => !!element,
+    /** returns a boolean indicating if the timePicker is disabled */
+    isDisabled: () => timePicker.disabled,
+    /** returns the input type attribute */
+    getInputType: () => timePicker.type,
+    /** returns the value (corresponds to state.value) */
+    getValue: () => timePicker.value,
+
+    /** calls the increment function on the timePicker component */
+    increment: (field?: FIELD) => componentInstance.increment(field),
```
shlomitc: Using the `componentInstance` is forbidden here.
It will not be available for the end user, meaning this function is useless.

I suggest testing such functions as an interaction (key press) and not calling to it directly

packages/wix-ui-core/src/components/TimePicker/TimePicker.spec.tsx
```diff
@@ -0,0 +1,288 @@
...
+
+    it('should display time in 12-hour format when true', () => {
+      const driver = createDriver(<TimePicker value = {SOME_VALUE} useAmPm = {true} />);
+      expect(driver.getValue()).toEqual(convertToAmPm(SOME_VALUE));
+    });
+  });
+
+  describe('step prop', () => {
+    it('should default to 1', () => {
+      const driver = createDriver(<TimePicker value = "10:30" />);
```
shlomitc: isn't it a duplicated test?

packages/wix-ui-core/src/components/TimePicker/TimePicker.spec.tsx
```diff
@@ -0,0 +1,288 @@
...
+    it('should increment value by 5 minutes when step is set to 5', () => {
+      const driver = createDriver(<TimePicker value = "10:30" step = {5}/>);
+      driver.increment();
+      expect(driver.getValue()).toEqual('10:35');
+    });
+  });
+
+  describe('separateSteps prop', () => {
+    it('should default to false', () => {
+      const driver = createDriver(<TimePicker value = "10:59" />);
```
shlomitc: duplicated test

packages/wix-ui-core/src/components/TimePicker/TimePicker.spec.tsx
```diff
@@ -0,0 +1,288 @@
...
+  });
+
+  describe('value prop', () => {
+    it('should use blank "--:--" as default', () => {
+      const driver = createDriver(<TimePicker />);
+      expect(driver.getValue()).toEqual('--:--');
+    });
+
+    it('should set the value according to the value prop', () => {
+      const driver = createDriver(<TimePicker value = {SOME_VALUE}/>);
```
shlomitc: redundant test

packages/wix-ui-core/src/components/TimePicker/TimePicker.tsx
```diff
@@ -0,0 +1,381 @@
...
+} from './utils';
+
+export interface TimePickerProps extends InputProps {
+  onChange?: (value: any) => any;
+  useNativeInteraction?: boolean;
+  useAmPm?: boolean;
+  step?: number;
+  separateSteps?: boolean;
+  value?: string;
+  valueWhenNull?: string;
```
shlomitc: why not just `placeholder`?

packages/wix-ui-core/src/components/Input/Input.tsx
guybenron: Added

packages/wix-ui-core/src/components/TimePicker/TimePicker.driver.ts
guybenron: The increment/decrement functions are an actual API on the component and cannot be simulated by keypress. I don't understand how componentInstance works :(

packages/wix-ui-core/src/components/TimePicker/TimePicker.spec.tsx
guybenron: I'm not sure what the standard you use is so I went a little overboard. There are two things I want to test: make sure default is 1, and make sure that expected behavior for 1 is correct.

I don't mind writing fewer tests :) Just not sure how to test those two different things. This is the same for the rest of the redundant/duplicated tests you mentioned

packages/wix-ui-core/src/components/TimePicker/TimePicker.tsx
guybenron: Because I am an idiot. Fixed

packages/wix-ui-core/src/components/TimePicker/TimePicker.spec.tsx
```diff
@@ -0,0 +1,298 @@
...
+      });
+    });
+
+    describe('focus', () => {
+      it('should focus the input element', () => {
+        const container = renderIntoDocument(<TimePicker/>);
+        const element = findRenderedDOMComponentWithTag(container, 'input');
+        expect(document.activeElement === element).toBeFalsy();
+        container.focus();
+        expect(document.activeElement === element).toBeTruthy();
```
lbelinsk: We should have `isFocused` method on the Input. If we don't please add it there, not inline in the test

packages/wix-ui-core/src/components/TimePicker/TimePicker.spec.tsx
```diff
@@ -0,0 +1,298 @@
...
+      });
+    });
+  });
+
+  describe('onChange prop', () => {
+    it('should be called with a new time when a new valid time is set', () => {
```
lbelinsk: check not only when set with arrowDown, but also when typing a value

packages/wix-ui-core/src/components/TimePicker/TimePicker.tsx
```diff
@@ -0,0 +1,392 @@
...
+    ...Input.propTypes,
+
+    /**
+     *  Callback function when user changes the value of the component.
+     *  Will be called only with valid values (this component is semi-controlled)
+     */
+    onChange: func,
+
+    /** Use native (input type = 'time') interaction */
+    useNativeInteraction: bool,
```
lbelinsk: I am not sure If this is not to hybrid...
Why not to always have type="time" and to wrap with the desired style. or alternatively have no type="time" and custom it..

packages/wix-ui-core/src/components/TimePicker/TimePicker.tsx
```diff
@@ -0,0 +1,392 @@
...
+  leftpad,
+  getFieldFromPos,
+  parseTime,
+  isValidTime
+} from './utils';
+
+export interface TimePickerProps extends InputProps {
+  onChange?: (value: any) => any;
+  useNativeInteraction?: boolean;
+  useAmPm?: boolean;
```
lbelinsk: showAmPM ? your call

packages/wix-ui-core/src/components/TimePicker/TimePicker.tsx
```diff
@@ -0,0 +1,392 @@
...
+      const step = props[propName];
+      if (typeof step === 'undefined' || step === null) { return null; }
+      if (typeof step !== 'number')            { return new Error(`Invalid prop '${propName}' supplied to '${componentName}': [${step}] is not a number.`); }
+      const integerStep = Math.trunc(step);
+      if (integerStep !== step)                { return new Error(`Invalid prop '${propName}' supplied to '${componentName}': [${step}] is not an integer.`); }
+      if (integerStep < 1 || integerStep > 60) { return new Error(`Invalid prop '${propName}' supplied to '${componentName}': [${step}] is not in range 1-60.`); }
+    },
+
+    /** When true, does not tie hour to minute when incrementing / decrementing (i.e, inc on 00:59 will go to 00:00 instead of 01:00) */
+    separateSteps: bool,
```
lbelinsk: sounds like a complication of the API. is it really needed?

packages/wix-ui-core/src/components/TimePicker/TimePicker.spec.tsx
```diff
@@ -0,0 +1,298 @@
...
+    });
+
+    it('should ignore this prop when value is set', () => {
+      const driver = createDriver(<TimePicker value = {SOME_VALUE} placeholder = {SOME_VALUE_WHEN_NULL} />);
+      expect(driver.getValue()).toEqual(SOME_VALUE);
+    });
+  });
+
+  // TODO
+  describe('styles', () => {
```
lbelinsk: nah, let's leave it to eyes in the e2e tests

packages/wix-ui-core/src/components/TimePicker/index.ts
```diff
@@ -0,0 +1,2 @@
+export {FIELD} from './utils';
```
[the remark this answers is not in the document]
guybenron: Because typescript.

packages/wix-ui-core/src/components/TimePicker/TimePicker.tsx
guybenron: Can't have always type = time because web browsers have poor support and no consistency, plus is sometimes impossible to style (e.g., chrome).
This is mostly for mobile web use

packages/wix-ui-core/src/components/TimePicker/TimePicker.tsx
guybenron: I can't say for sure. It's not a complicated idea, and I feel like sometimes you might want to customize it.

packages/wix-ui-core/src/components/TimePicker/TimePicker.tsx
guybenron: Urgh. Naming convention is hard. Not sure what to do here

PR #253, opened 2018-02-19
improve text tests

lbelinsk:
- Organized the tests
- Added missing tests
- enriched the driver for the upcoming back office tests

### DIFF
--- a/packages/wix-ui-core/src/components/StylableText/Text.driver.ts
+++ b/packages/wix-ui-core/src/components/StylableText/Text.driver.ts
@@ -10,6 +10,10 @@ export const textDriverFactory = ({element}) => {
     /** check if component has ellipsis */
     isEllipsis: () => stylableDOMUtil.hasStyleState(element, 'ellipsis'),
     /** check if element has title attribute */
-    hasTitleAttribute: () => element.getAttribute('title') !== null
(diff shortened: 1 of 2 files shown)


packages/wix-ui-core/src/components/StylableText/Text.driver.ts
```diff
@@ -10,6 +10,10 @@ export const textDriverFactory = ({element}) => {
     /** check if component has ellipsis */
     isEllipsis: () => stylableDOMUtil.hasStyleState(element, 'ellipsis'),
     /** check if element has title attribute */
-    hasTitleAttribute: () => element.getAttribute('title') !== null
+    hasTitleAttribute: () => element.getAttribute('title') !== null,
+    /** get the rendered tag name */
+    getTagName: () => element.tagName,
+    /** get the rendered content */
+    getChildren: () => element.innerHTML
```
shlomitc: I don't really like the `getChildren` name. also, when do we want to use `innerHTML`? Why not just the text?

lbelinsk: because theoretically, it can be not only text. (although it is an abuse).
What would you prefer it to be?

PR #255, opened 2018-02-20
adding build step to create index.st.css

JoMarton:


amiryonatan:
Can we not do a script that generates it automatically (like we do in import-path) ? 
and not manage this file in the source code ?

shlomitc:
What about tree shaking? Does stylable takes care of it?

JoMarton:
`index.st.css` will be generated automatically. I add it now to `gitignore`.

PR #264, opened 2018-02-22
<Checkbox /> - fix driver

danielstro:
Checked / indeterminate in the driver only work if you use the default tick marks or if you pass a certain data-hook, which causes some problems in wix-ui-backoffice.
I changed them to check for stylable states instead, so they will be less development dependant.

Also, I replaced an overly complex test svg with a span for readability.

### DIFF
--- a/packages/wix-ui-core/src/components/Checkbox/Checkbox.driver.ts
+++ b/packages/wix-ui-core/src/components/Checkbox/Checkbox.driver.ts
@@ -24,6 +24,8 @@ export const checkboxDriverFactory = ({element, eventTrigger}) => {
     getType: () => element.getAttribute('type'),
     /** checks if the tickmark exists, i.e. the checkbox is checked */
     isChecked: () => hasStyleState(element, 'checked'),
+    /** returns true if the element has indeterminate state */
+    isIndeterminate: () => hasStyleState(element, 'indeterminate'),
     /** returns elements textContent */
     getTextContent: () => element.textContent,
     /** returns the checkbox children */
@@ -36,25 +38,11 @@ export const checkboxDriverFactory = ({element, eventTrigger}) => {
     isDisabled: () => element.getAttribute('disabled') === '',
     /** returns true if the element has error state */
     hasErrorState: () => hasStyleState(element, 'error'),
-    /** returns true if the element has indeterminate state */
-    hasIndeterminateState: () => hasStyleState(element, 'indeterminate'),
     /** returns true if the element has focus state */
     hasFocusState: () => hasStyleState(element, 'focus'),
     /** returns true if the element has disabled state */
     hasDisabledState: () => hasStyleState(element, 'disabled'),
     /** returns true if the element has error state */
     hasReadOnlyState: () => hasStyleState(element, 'readonly'),
-    styles: {
-      /** returns elements min-width css property */
-      getMinWidth: () => getCheckboxStyle().minWidth,
-      /** returns elements width css property */
-      getWidth: () => getCheckboxStyle().width,
-      /** returns elements height css property */
-      getHeight: () => getCheckboxStyle().height,
-      /** returns elements padding css property */
-      getPadding: () => getCheckboxStyle().padding,
-      /** returns elements border-radius css property */
-      getBorderRadius: () => getCheckboxStyle().borderRadius,
-    }
   };
 };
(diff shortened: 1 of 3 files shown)


packages/wix-ui-core/src/components/Checkbox/Checkbox.spec.tsx
```diff
@@ -245,7 +227,7 @@ describe('Checkbox', () => {
         />
       );
 
-      expect(checkbox.isIndeterminate()).toBe(true);
+      expect(checkbox.hasIndeterminateState()).toBe(true);
```
JoMarton: I see a mismatch between what the test says it does (display an icon) and what is says it expects (to have an indeterminate state).  I think names like `hasXstate` are confusing to driver users. If this method is used to make sure that checkbox is indeterminate, let's call it `isIndereminate`. what do you think?

packages/wix-ui-core/src/components/Checkbox/Checkbox.driver.ts
```diff
@@ -23,17 +23,13 @@ export const checkboxDriverFactory = ({element, eventTrigger}) => {
     /** returns elements type attribute */
```
JoMarton: about `getCheckboxStyle` - does it make sense to use `window.getComputedStyle(...)` in a driver for node tests?
I don't even see this method used anywhere

packages/wix-ui-core/src/components/Checkbox/Checkbox.tsx
```diff
@@ -146,7 +146,7 @@ export class Checkbox extends React.PureComponent<CheckboxProps, CheckboxState>
     const focus = this.state.isFocused;
 
     return (
-      <div {...style('root', {checked, disabled, focus, readonly, error, indeterminate}, this.props) }
+      <div {...style('root', {checked: indeterminate ? false : checked, disabled, focus, readonly, error, indeterminate}, this.props) }
```
JoMarton: regarding style states - I understand you want the component to be styled according to indeterminate state when both checked and indeterminate, but that can be easily achieved using the stylesheet: `root:checked:not(:indereminate)`. I prefer that and keep the style state mapping simple and not co-dependent

packages/wix-ui-core/src/components/Checkbox/Checkbox.driver.ts
```diff
@@ -23,17 +23,13 @@ export const checkboxDriverFactory = ({element, eventTrigger}) => {
...
     getTextContent: () => element.textContent,
     /** returns the checkbox children */
     children: () => element.querySelectorAll('[data-hook="CHECKBOX_CHILD_CONTAINER"]'),
     /** returns the checkbox tickmark */
-    tickmark: () => element.querySelector('[data-hook="CHECKBOX_TICKMARK"]'),
-    /** returns the indeterminate icon */
-    indeterminateMark: () => element.querySelector('[data-hook="CHECKBOX_INDETERMINATE"]'),
-    /** returns a boolean indicating whether the checkbox is in indeterminate state */
-    isIndeterminate: () => !!element.querySelector('[data-hook="CHECKBOX_INDETERMINATE"]'),
+    tickmark: () => element.querySelector('[data-hook="CHECKBOX_BOX"]').firstChild,
```
shlomitc: I really don't like returning DOM Elements. it doesn't make much sense for the user.

Why not just create a specific data-hook, one for the regular tick icon and one for the indeterminate icon and very that the right one exists?

packages/wix-ui-core/src/components/Checkbox/Checkbox.tsx
danielstro: This is a bit of a bizarre case, that may require more thinking but right now, when indeterminate is passed as a prop, we ignore checked. So I added this to make sure that the styling wouldn't collide.
I know I can use that selector however everyone else will need to know to use it as well.
The problem is that indeterminate and checked don't go together. The native checkbox gives priority to checked so maybe this is what we should do

packages/wix-ui-core/src/components/Checkbox/Checkbox.driver.ts
danielstro: I agree but since we allow custom icons, this will not work unless the custom icons have the data-hook as well.

packages/wix-ui-core/src/components/Checkbox/Checkbox.spec.tsx
danielstro: I agree, the reason I kept it with this name and didn't change it is because technically this is what the driver is checking, but it doesn't matter

packages/wix-ui-core/src/components/Checkbox/Checkbox.driver.ts
danielstro: no idea why this is here

PR #273, opened 2018-02-27
<RadioButton /> - Focusing input ref on click to enable keyboard nav instantly

liorzisman:


packages/wix-ui-core/src/components/RadioButton/RadioButton.tsx
```diff
@@ -46,6 +46,8 @@ export interface RadioButtonState {
 
 export class RadioButton extends React.Component<RadioButtonProps, RadioButtonState> {
 
+  radioRef = undefined;
```
shlomitc: `private readioRef`?

liorzisman: roger

liorzisman:
@shlomitc verifying the radio input is getting focused as expected in the onChange test which is the most relevant to it

PR #274, opened 2018-02-27
<Checkbox /> - fix driver

danielstro:
In the protractor drive, the isChecked method didn't work.

### DIFF
--- a/packages/wix-ui-core/src/components/Checkbox/Checkbox.protractor.driver.ts
+++ b/packages/wix-ui-core/src/components/Checkbox/Checkbox.protractor.driver.ts
@@ -6,5 +6,5 @@ export const checkboxDriverFactory = component => ({
     /** Indicates whether the component is disabled or not */
     isDisabled: () => component.getAttribute('disabled') === '',
     /** Indicates whether the component is checked */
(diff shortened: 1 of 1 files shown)


packages/wix-ui-core/src/components/Checkbox/Checkbox.protractor.driver.ts
```diff
@@ -6,5 +6,5 @@ export const checkboxDriverFactory = component => ({
     /** Indicates whether the component is disabled or not */
     isDisabled: () => component.getAttribute('disabled') === '',
     /** Indicates whether the component is checked */
-    isChecked: () => component.getAttribute('aria-checked') !== 'false'
+    isChecked: () => component.getAttribute('aria-checked').then(val => val === 'true')
```
shlomitc: can `aria-checked` be applied without a `true` value but still apply it's behavior

shlomitc:
@danielstro - do you remember if this PR is still relevant?

danielstro:
@shlomitc it is not. If we need it after the checkbox refactor I will create a new PR for it but I don't think we will need it.

PR #275, opened 2018-02-27
Stylable Toggle Switch

alisey:
The initial goal was to accept icons through props, but I noticed some issues, started changing stuff, and basically rewrote the component and added some adjustments to CSS and tests.

Changes:
* Fix type signatures
* Add props for icons and tabIndex
* Remove props for inline styles of internal elements
* Use the native checkbox to handle events
* Fix CSS for the focus outline (top: 50%; translateY(-50%) was giving the component incorrect bounding box)
* Move non-essential stying from the base style to theme
* Rename internal parts: outerLabel -> track, innerLabel -> knob, toggleIcon -> knobIcon
* Don't add autogenerated ID to the component (there's no point in it if it's private)
* Don't pass event to onChange handler (there's no need in it at the moment)
* Add focus and focusVisible states (the latter can be used to show the outline only when the component is focused via keyboard, like native checkboxes and material ui switches)


packages/wix-ui-core/src/components/StylableToggleSwitch/ToggleSwitch.tsx
```diff
@@ -1,90 +1,122 @@
...
+          disabled,
+          focus: this.state.focus,
+          focusVisible: this.state.focusVisible
+        }, this.props)}
+        style={this.props.style}
+      >
+        <div className={tsStyle.track} />
+        <div className={tsStyle.knob}>
+          <div className={tsStyle.knobIcon}>
+            {checked ? this.props.checkedIcon : this.props.uncheckedIcon}
```
lbelinsk: Did you check that it is ok also when there is no icon provided?

packages/wix-ui-core/src/components/StylableToggleSwitch/ToggleSwitch.tsx
```diff
@@ -1,90 +1,122 @@
...
+          </div>
+        </div>
         <input
+          id={this.props.id}
```
lbelinsk: In this refactored component you don't need the id prop, so it is ok to remove it. There is no convention to have id prop for components

packages/wix-ui-core/src/components/StylableToggleSwitch/ToggleSwitch.tsx
```diff
@@ -1,90 +1,122 @@
...
+
+  private handleChange: React.ChangeEventHandler<HTMLInputElement> = e => {
+    if (this.props.onChange) {
+      this.props.onChange();
+    }
+  }
+
+  // Doesn't get invoked if the input is disabled.
+  private handleMouseDown: React.MouseEventHandler<HTMLElement> = e => {
+    if (e.button === 0 && !this.state.focus) {
```
lbelinsk: I am not sure I understand this condition.

packages/wix-ui-core/src/components/StylableToggleSwitch/ToggleSwitch.tsx
```diff
@@ -1,90 +1,122 @@
...
+      this.focusedByMouse = true;
+    }
+  }
+
+  private handleFocus: React.FocusEventHandler<HTMLElement> = e => {
+    this.setState({focus: true, focusVisible: !this.focusedByMouse});
```
lbelinsk: I think now that we have the new feature of state with params in stylable, we can have focus state which gets params "mouse" or "keyboard". It feels more semantically correct, instead of having focus & focusVisible   ----> focus(mouse) or focus(keyboard).
WDYT?

packages/wix-ui-core/src/components/StylableToggleSwitch/ToggleSwitch.spec.tsx
```diff
@@ -6,31 +6,29 @@ import {isTestkitExists} from 'wix-ui-test-utils/vanilla';
...
       driver.click();
 
       expect(driver.isChecked()).toBe(false);
     });
 
     it('should pass down to input when checked', () => {
-      const driver = createDriver(<ToggleSwitch checked onChange={noop}/>);
-      expect(driver.isChecked()).toBeTruthy();
+      const driver = createDriver(<ToggleSwitch checked />);
+      expect(driver.isChecked()).toBe(true);
```
lbelinsk: The difference between toBeTruthy() and toBe(true) is that toBeTruthy() returns true not only for "true" value, but also for any truthy value like 1, 'a', ... so semantically it is more correct, but It is not that important in this case

packages/wix-ui-core/src/components/StylableToggleSwitch/ToggleSwitch.spec.tsx
```diff
@@ -67,44 +65,28 @@ describe('ToggleSwitch', () => {
...
-
+  describe('attributes', () => {
     it('should apply user specified id', () => {
       const testId = 'testId';
-      const driver = createDriver(<ToggleSwitch onChange={noop} id={testId}/>);
+      const driver = createDriver(<ToggleSwitch id={testId}/>);
       expect(driver.getId()).toBe(testId);
     });
+
+    it('should apply user specified tabIndex', () => {
```
lbelinsk: I would add a test "should have tabIndex 0  by default" as well

packages/wix-ui-core/src/components/StylableToggleSwitch/ToggleSwitch.spec.tsx
```diff
@@ -67,44 +65,28 @@ describe('ToggleSwitch', () => {
...
-    });
-  });
-
-  describe('id prop', () => {
-    it('should apply arbitrary unique id be default', () => {
-      const driver = createDriver(<ToggleSwitch onChange={noop}/>);
-      expect(driver.getId()).toBeDefined();
-    });
-
+  describe('attributes', () => {
```
lbelinsk: Missing tests for the new configurable Icons feature. It is completely not tested. when checked. unchecked, w/o icons...

packages/wix-ui-core/src/components/StylableToggleSwitch/ToggleSwitch.spec.tsx
```diff
@@ -67,44 +65,28 @@ describe('ToggleSwitch', () => {
...
     });
   });
 
   describe('enzyme testkit', () => {
     it('should exist', () => {
-      expect(isEnzymeTestkitExists(<ToggleSwitch onChange={noop}/>, enzymeToggleSwitchTestkitFactory, mount)).toBe(true);
+      expect(isEnzymeTestkitExists(<ToggleSwitch />, enzymeToggleSwitchTestkitFactory, mount)).toBe(true);
     });
   });
 });
```
lbelinsk: There are still some uncovered parts. For example, the focus new feature. You can test that the component is in the expected state when focusing with mouse or keyboard...

packages/wix-ui-core/src/components/StylableToggleSwitch/ToggleSwitch.st.css
```diff
@@ -1,34 +1,29 @@
...
+  left: 0;
+  margin: 0;
+  padding: 0;
+  cursor: inherit;
+}
 
-  cursor: pointer;
+.track {}
 
-  top: 50%;
```
lbelinsk: This was crucial for santa, just pointing it out that we must be sure all desired features are still working for the product request

alisey: Right, Gilad showed me why it's needed. But it shouldn't be part of the core, it should be in the stylable theme for santa toggle switch, which doesn't exist yet.

We also need to address this problem in ToggleSwitchStyle, and I plan to rewrite it, just don't want to expand the scope of this PR. You know what, I'm going to put this vertical alignment thing into ToggleSwitchStyle, which is going to break focus outline styling, but that's ok since it was already broken. And then I'm going to remove it and replace with something else later.
image

packages/wix-ui-core/src/components/StylableToggleSwitch/ToggleSwitch.tsx
alisey: I remember we needed this for Santa when I was working on pagination. Santa needs to be able to put ID on the root. Also useful for things like `<label for="myswitch">`

packages/wix-ui-core/src/components/StylableToggleSwitch/ToggleSwitch.tsx
```diff
@@ -1,90 +1,122 @@
...
   }
+
+  private handleKeyDown: React.KeyboardEventHandler<HTMLElement> = e => {
+    // Pressing any key should make the focus visible, even if the checkbox
+    // was initially focused by mouse.
+    this.setState({focusVisible: true});
+  }
+
+  private handleChange: React.ChangeEventHandler<HTMLInputElement> = e => {
+    if (this.props.onChange) {
```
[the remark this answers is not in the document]
alisey: I do want it optional, mainly because all event handlers for standard controls are optional. Will add it to default props.

packages/wix-ui-core/src/components/StylableToggleSwitch/ToggleSwitch.tsx
alisey: I'm going to remove the second part of it, it's no longer necessary. Would it be clear like this?
```
if (e.button === 0) {
    this.focusedByMouse = true;
}
```

packages/wix-ui-core/src/components/StylableToggleSwitch/ToggleSwitch.tsx
alisey: Yeah, I like your suggestion, much more readable that way. On the other hand, `:focus-visible` is a part of CSS4 draft, if some day it becomes standard then I think it's preferable to follow the standard API. So I don't know really.

I'm going to go with your suggestion, because it's nicer, and we can reconsider in the future.

packages/wix-ui-core/src/components/StylableToggleSwitch/ToggleSwitch.tsx
alisey: Ok in what sense? If the icon is not provided it's just an empty div.

alisey:
* Added inine styles back (incompatible with what Barak is doing, he removed them)
* onChange is now noop by default
* Added tests for the icon
* Added test for default tab index
* Added back `top: 50%` style

- Didn't change focus-visible to focus(keyboard/mouse), because then plain `:focus` stops working.
- Didn't add tests for focus behavior, because in node environment I would need to simulate the sequence of events I expect on mouse/keyboard press, which doesn't necessarily correspond to the sequence a browser would generate, and the tests would be passing even if the component is totally broken. We really need browser envoriment with non-simulated events for such tests.

packages/wix-ui-core/src/components/StylableToggleSwitch/ToggleSwitch.tsx
```diff
@@ -1,90 +1,128 @@
...
+  private focusedByMouse = false;
 
   render() {
     const {checked, disabled, styles} = this.props;
-    const {id} = this;
 
     return (
-      <label {...style('root', {checked, disabled}, this.props)} style={styles.root} tabIndex={0} ref={ref => this.toggle = ref}>
+      <div
+        {...tsStyle('root', {
```
shlomitc: why not just `style`? this is the convention around all other components

alisey: There was `style`, which is inline style, then `styles`, which is a collection of styles for the parts of the component, and then also `style` which is the stylesheet. It was kinda confusing to me, and I wanted to make them more distinct.

shlomitc: ok, so maybe we should give it a better name in general, but now it looks weird and it won't work for components with a single character like input. will it be `iStyle`? I even prefer calling it `s` instead.

alisey: How about `css`? What we're importing is basically a stylesheet.

shlomitc: cool

alisey: Got a different idea. The imported style is now `style`, and the inline styles are `inlineStyles`.

packages/wix-ui-core/src/components/StylableToggleSwitch/ToggleSwitch.st.css
```diff
@@ -1,34 +1,25 @@
 .root {
-  -st-states: checked, disabled;
+  /* Info on focus-visible: https://drafts.csswg.org/selectors-4/#the-focus-visible-pseudo */
+  -st-states: checked, disabled, focus, focus-visible;
```
lbelinsk: why in the end you decided not to use the stylable states with params feature?
Is it because of the CSS4?

alisey: @lbelinsk That's one reason, another is that with `:focus(mouse/keyboard)` we lose the ability to use plain `:focus`. We had the same issue in checkbox, where we wanted to have `:checked(true/false/indeterminate)` and then went back to `:checked` and `:indeterminate`.

PR #276, opened 2018-02-27
change badge style

lbelinsk:
Placing text beside icons showed me that this needs to be changed.

Before:
image

After:
image


shlomitc:
Is this screenshot taking from wix-ui-backoffice?
If so - I think this style change you did should be applied from the wix-ui-backoffice, and not in the wix-ui-core

lbelinsk:
@shlomitc I think that each theme would like to be able to align the text beside some icon, so I do think we want it to be in the core

Gilad changed this in the toggleSwitch as well at the time to inline-flex

Issue #259, opened 2018-02-21
Fix Checkbox


danielstro:
Implementation:

- [x] this.id is not in use (by the implementation), so why is it initialized? this is also the only cause of importing `lodash.unique`.
- [x] event has `checked` === true,  when checkbox is in indeterminate mode. why?
- [x] comp extends PureComponent, but has an array in props. will render every time if not the same array (probable case)
- [ ] different ref function is created on every render. can be caught by linter?
- [x] tabIndex is hardcoded to 0 for the root that causes the nested tabIndex to be contextually seperator - bug!
- [x] ~~handleClick should call this.checkbox.click(). not focus. focus should have separate handling.~~
  Changed the component to use label so this happens natively.
- [x] the event instance being passed to this.props.onChange is not really a Syntetic event. spreading does not give us the prototype methods. event instances are recycled in react, so we need to find a proper pattern.
- [ ] the event that is passed to onChange is wrong, we need to find a better pattern.
- [ ] If the checkbox is placed in another element that has an "onClick" listener, it will be called twice, one because of the label and again because of the input. A possible solution is to place stopPropagation on the input's onClick but this still returns an event with the wrong target

TypeScript:
- [x] this.checkbox should be "HTMLInputElement | null", instead of using ! in the ref function and forcing a wrong type. this will ensure every place using it will check it.
- [x] onChange has a combined type of all even handlers- bad. specific events have specific event flags
- [x] Props should extend the relevant native HTMLInput attributes interface. e.g. React.InputHTMLAttributes<HTMLInputElement>

Styles:
- [ ] ~~checked state should be Enum with true/false/interminate~~
- [x] icon should be a single part and not ::tickIcon / ::indeterminateIcon
- [x] vars should never have width / height for root - this can be simple CSS declarations
Documentation (inline):
- [x] the documentation for the component's input (props) should be next to the definitions in the props interface (and not on propTypes). this way, IDEs pick them automatically when showing auto-completion.
- [x] comments should have helpful content, or not there at all
- [x] focus-visible isn't working properly. Right now we are not allowing focus state when clicking on the component but this isn't true for the native checkbox.

Generally:
- [x] when using TypeScript with React, one gets full static type validation for props, meaning using the prop-types package is redundent (assuming all pieces use typescript).    
- [x] I would like to find a way to remove the propsTypes from source, our projects should move to typescript, and we should get static validation. we could optionally try some debug build with something like: https://github.com/grncdr/ts-react-loader#what-it-does    
[...]

PR #281, opened 2018-03-01
<Checkbox /> - Clean & Improve comp

danielstro:
This is the first of several changes regarding <Checkbox />. 
Mostly cleaning code and some cosmetic changes, but also changing the checked state to use enum and removing this.id.
For more information see this issue #259

### DIFF
--- a/packages/wix-ui-core/src/components/Checkbox/Checkbox.spec.tsx
+++ b/packages/wix-ui-core/src/components/Checkbox/Checkbox.spec.tsx
@@ -71,20 +71,10 @@ describe('Checkbox', () => {
     expect(checkbox.input().required).toBe(true);
   });
 
-  it('Accepts "autofocus" prop', async () => {
-    if (document.hasFocus()) {
+  it('Accepts "autofocus" prop', () => {
+    const checkbox = createDriver(<Checkbox autoFocus />);
 
-      const checkbox = createDriver(
-        <Checkbox autoFocus />
-      );
-
-      expect(document.activeElement).toBe(checkbox.input());
-
-    } else {
-      console.error(// tslint:disable-line no-console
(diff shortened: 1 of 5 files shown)


packages/wix-ui-core/src/components/Checkbox/CheckboxStyle.st.css
```diff
@@ -39,8 +39,8 @@
     border: value(borderDefault);
 }
 
-.root:checked::box,
-.root:indeterminate::box {
+.root:checked(true)::box,
```
alisey: We're losing the ability to simply write `.myCheckbox:checked`. I was dealing with a similar issue in the toggle switch yesterday, I wanted to have `:focus` and `:focus(mouse)`, but it seems like Stylable doesn't support this at the moment.

idoros: I agree, I know this is going back from the previous review, but its better as 2 states: checked & indeterminate

packages/wix-ui-core/src/components/Checkbox/Checkbox.spec.tsx
```diff
@@ -95,7 +81,7 @@ describe('Checkbox', () => {
       expect(document.activeElement).toBe(checkbox.input());
 
     } else {
-      console.warn(// tslint:disable-line no-console
+      console.error(// tslint:disable-line no-console
```
idoros: In which scenario are the test run without focus support?

danielstro: I think this is a leftover from stylable components, I think the tests were flakey without this. I will remove it and try to see if this affects the tests.

PR #286, opened 2018-03-05
ToggleSwitch RTL

alisey:


packages/wix-ui-core/src/components/StylableToggleSwitch/ToggleSwitch.spec.tsx
```diff
@@ -81,6 +81,11 @@ describe('ToggleSwitch', () => {
       const driver = createDriver(<ToggleSwitch tabIndex={7} />);
       expect(driver.getTabIndex()).toBe(7);
     });
+
```
lbelinsk: I would add a test for ltr by default

packages/wix-ui-core/src/components/StylableToggleSwitch/ToggleSwitch.tsx
```diff
@@ -50,7 +51,9 @@ export class ToggleSwitch extends React.PureComponent<ToggleSwitchProps, ToggleS
     /** Icon inside of the knob when checked */
     checkedIcon: propTypes.node,
     /** Icon inside of the knob when unchecked */
-    uncheckedIcon: propTypes.node
+    uncheckedIcon: propTypes.node,
+    /** Direction */
+    dir: propTypes.oneOf(['ltr', 'rtl'])
```
lbelinsk: what about adding dafaultProps.dir = 'ltr' ?

alisey:
LTR by default, added test.

shlomitc:
are we sure the using `ltr` is needed? shouldn't it just be the default behavior and only when you need `rtl` pass it as a prop (or use it from the `body`)?

PR #289, opened 2018-03-06
chore(Input): remove `id` from state and pass from props instead

argshook:


packages/wix-ui-core/src/components/Input/Input.tsx
```diff
@@ -137,7 +132,7 @@ export class Input extends React.Component<InputProps, InputState> {
         disabled={disabled}
         autoComplete={autoComplete}
         autoFocus={autoFocus}
-        id={this.state.id}
+        id={id}
```
shlomitc: how should it be handled if you don't have an id? undefined?

argshook: i'm missing `id: ''` in default props. However, whole approach of handling props is what i'd like to talk about. I would `{...rest}` instead of defining all props.

PR #290, opened 2018-03-06
Docs/input autodocs

argshook:


packages/wix-ui-core/src/components/Input/Input.tsx
```diff
@@ -16,25 +15,58 @@ const createAriaAttributes = props => {
...
 
 export interface InputProps {
-  disabled?: boolean;
+  /** Turns on or off autocomplete property, which is responsible for default browser autocomplete suggestion */
   autoComplete?: 'on' | 'off';
+
+  /** Standard React Input autoFocus (focus the element on mount) */
   autoFocus?: boolean;
+
+  className?: string;
```
shlomitc: I think it's redundant in the component

argshook: copied from `propTypes`, i'll remove it

argshook: we'll i can't remove it because https://github.com/wix/wix-ui/blob/master/packages/wix-ui-core/src/baseComponents/InputWithOptions/InputWithOptions.tsx#L177

PR #293, opened 2018-03-07
chore(Input): define only what's used

argshook:
this is for discussion

NOT TO BE MERGED YET

### DIFF
--- a/packages/wix-ui-core/src/components/Input/Input.tsx
+++ b/packages/wix-ui-core/src/components/Input/Input.tsx
@@ -1,152 +1,56 @@
 import * as React from 'react';
-const uniqueId = require('lodash.uniqueid');
 import style from './Input.st.css';
-
-const createAriaAttributes = props => {
-  const aria = 'aria';
-  const ariaAttribute = {};
-  Object
-    .keys(props)
-    .filter(key => key.startsWith(aria))
-    .map(key => ariaAttribute[`${aria}-${key.substr(aria.length).toLowerCase()}`] = props[key]);
-  return ariaAttribute;
-};
 
 const NUMBER_REGEX = /^[\d.,\-+]*$/;
 
 export interface InputProps {
-  /** Turns on or off autocomplete property, which is responsible for default browser autocomplete suggestion */
-  autoComplete?: 'on' | 'off';
-
-  /** Standard React Input autoFocus (focus the element on mount) */
-  autoFocus?: boolean;
-
   className?: string;
 
   /** Makes the component disabled */
   disabled?: boolean;
 
-  /** Name for the input */
-  name?: string;
-
-  /** Standard input onBlur callback */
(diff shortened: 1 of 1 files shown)


packages/wix-ui-core/src/components/Input/Input.tsx
```diff
@@ -1,152 +1,56 @@
...
-  Object
-    .keys(props)
-    .filter(key => key.startsWith(aria))
-    .map(key => ariaAttribute[`${aria}-${key.substr(aria.length).toLowerCase()}`] = props[key]);
-  return ariaAttribute;
-};
-
 const NUMBER_REGEX = /^[\d.,\-+]*$/;
 
 export interface InputProps {
```
lbelinsk: I tend to agree that in Input, since it is so low level component, we would need to support many props that are being directly mapped as attributes for the html input.
That `...props` approach would not be good in most of the components.

In the future people would ask for more props, and then they would have to do pr's for the core if we won't add it.

Actually, since everything is optional with question marks, we can use the `Partial` key word.
```js
InputProps = Partial<...  
```

packages/wix-ui-core/src/components/Input/Input.tsx
```diff
@@ -1,152 +1,56 @@
...
-        onKeyDown={onKeyDown}
-        onKeyUp={onKeyUp}
-        placeholder={placeholder}
-        readOnly={readOnly}
-        required={required}
-        tabIndex={tabIndex}
-        type={type}
-        value={value}
-        {...ariaAttributes}
+        {...restProps}
```
lbelinsk: You do want to pass the `disabled` prop as well to the html input

packages/wix-ui-core/src/components/Input/Input.tsx
amiryonatan: If u won't declare the properties consumers will not get Autocomplete for the available props.
besides, we will get an exception that Input does not have a prop type that u gave it if its not defined

amiryonatan: We could use the following type as props:
`React.InputHTMLAttributes<HTMLInputElement>` 
which is the native `input` attributes

amiryonatan: @argshook @lbelinsk 
Btw, Erez brought an interesting matter.
should we not keep the propTypes for JS usages of the library ? 
If we remove them JS developers will get warnings when passing wrong values ..

packages/wix-ui-core/src/components/Input/Input.tsx
```diff
@@ -1,152 +1,56 @@
...
   };
 
-  constructor(props) {
-    super(props);
-    this.state = {
-      id: uniqueId('Input')
-    };
-    this._onChange = this._onChange.bind(this);
-  }
+  _onChange = event => {
```
alisey: This is unused, right?

PR #294, opened 2018-03-07
wix-ui-core exposing focus, blur, and select on Input

guybenron:


packages/wix-ui-core/src/components/Input/Input.tsx
```diff
@@ -79,16 +79,22 @@ export interface InputState {
...
 
+  private inputRef: HTMLInputElement;
+
   static defaultProps = {
     type: 'text'
   };
 
   constructor(props) {
     super(props);
+    this.inputRef = null;
```
lbelinsk: needed?

PR #303, opened 2018-03-09
chore(AutoExampleDriver): clear focus by clickin `body` when calling `reset`

### LINKED ISSUE #297: AutoExampleDriver.reset() - need to focus on body
When doing `AutoExampleDriver.reset()` I assume that the page is as it was when it was loaded.
which means that the focus should be on the `body`.

If my test sets a focus on the tested component, I would expect it to be blured after `AutoExampleDriver.reset()`.

So can we add 
`browser.$('body').click();` tot he `reset` function?

argshook:
closes #297

### DIFF
--- a/packages/wix-storybook-utils/src/AutoExample/protractor.driver.js
+++ b/packages/wix-storybook-utils/src/AutoExample/protractor.driver.js
@@ -46,6 +46,8 @@ module.exports = {
 
     return browser.executeScript(script, props);
   },
-  reset: () =>
-    browser.executeScript('window.autoexample.resetState()')
+  reset: () => {
+    browser.$('body').click();
+    return browser.executeScript('window.autoexample.resetState()');
+  }
 };


argshook:
as far as i remember all protractor api is async and controlled internally. i guess i can write `async` but protractor does that for me already

hepiyellow:
According to this, we should use async/await:
http://www.protractortest.org/#/async-await

> The Web Driver Control Flow is used to synchronize your commands so they reach the browser in the correct order (see /docs/control-flow.md for details). **In the future, the control flow is being removed** (see SeleniumHQ's github issue for details). Instead of the control flow, you can synchronize your commands with promise chaining or the upcoming ES7 feature async/await.

shlomitc:
`protractor` does have his magic when coming to promises in the control flow, but this can be disabled (and we actually encourage it) in specific projects.

So I agree with Erez that we should use `async/await` there

PR #304, opened 2018-03-11
[WIP !] -Add withExamples story url param

hepiyellow:
It is helpful for e2e tests, that the e2e story will not include examples, so that we can test an isolated <AutoExample> component.
For example when testing focus behavior, the examples may interfeer if they have autoFocus, and make the test implementation more complicated.

- Added url param `withExamples` to control the rendering of the exmaples.
- in `wix-ui-test-utils`
  -  make `withExamples` default to `true.
  - add `createStoryUrl()` which accepts a params object (deprecate the usage of the multiple arguments `getStoryUrl()`)

### DIFF
--- a/packages/wix-storybook-utils/package.json
+++ b/packages/wix-storybook-utils/package.json
@@ -17,6 +17,7 @@
   },
   "homepage": "https://github.com/wix/wix-ui#readme",
   "main": "./dist/src/index.js",
+  "types": "./dist/src/index.d.js",
   "scripts": {
     "pretest": "npm run build && build-storybook",
     "test": "haste test --jest",
--- a/packages/wix-storybook-utils/package.json
+++ b/packages/wix-storybook-utils/package.json
@@ -55,6 +55,7 @@
     "import-path": "latest",
     "jsx-to-string": "git+https://github.com/amiryonatan/jsx-to-string.git#release_branch",
     "loader-utils": "^1.1.0",
+    "query-string": "^5.1.0",
     "react-autodocs-utils": "^2.0.1",
     "react-collapse": "^4.0.3",
     "react-docgen": "^2.20.0",
@@ -66,6 +67,7 @@
   "devDependencies": {
     "@storybook/addon-options": "^3.2.16",
     "@storybook/react": "^3.2.16",
+    "@types/query-string": "^5.1.0",
     "enzyme": "^3.0.0",
     "enzyme-adapter-react-15": "^1.0.5",
     "eslint-config-wix": "latest",
--- a/packages/wix-storybook-utils/src/StoryNew/index.js
+++ b/packages/wix-storybook-utils/src/StoryNew/index.js
@@ -6,12 +6,9 @@ import Markdown from '../Markdown';
 import CodeBlock from '../CodeBlock';
 import AutoExample from '../AutoExample';
 import AutoDocs from '../AutoDocs';
-import * as queryString from 'query-string';
 import styles from '../Story/styles.scss';
-
+import {isStoryUrlWithExamples} from '../storyUtils';
 const isE2E = global.self === global.top;
-
-export const IS_EXAMPLES_VISIBLE_PARAM_NAME = 'withExamples';
 
 export default ({
   category,
@@ -35,7 +32,7 @@ export default ({
             parsedSource={_metadata}
             />
 
-          {queryString.parse(window.location.search)[IS_EXAMPLES_VISIBLE_PARAM_NAME] !== undefined && examples}
+          {isStoryUrlWithExamples(window.location.search) && examples}
         </div>
       );
     }
--- a/packages/wix-storybook-utils/src/StoryNew/index.js
+++ b/packages/wix-storybook-utils/src/StoryNew/index.js
@@ -6,10 +6,12 @@ import Markdown from '../Markdown';
 import CodeBlock from '../CodeBlock';
 import AutoExample from '../AutoExample';
 import AutoDocs from '../AutoDocs';
-
+import * as queryString from 'query-string';
 import styles from '../Story/styles.scss';
 
 const isE2E = global.self === global.top;
+
+export const IS_EXAMPLES_VISIBLE_PARAM_NAME = 'withExamples';
 
 export default ({
   category,
@@ -33,7 +35,7 @@ export default ({
             parsedSource={_metadata}
             />
 
-          {examples}
+          {queryString.parse(window.location.search)[IS_EXAMPLES_VISIBLE_PARAM_NAME] !== undefined && examples}
         </div>
       );
     }
--- a/packages/wix-ui-test-utils/src/protractor/protractor-helpers.ts
+++ b/packages/wix-ui-test-utils/src/protractor/protractor-helpers.ts
@@ -8,11 +8,31 @@ import {
 
 const encode = global.encodeURIComponent;
 
-export const getStoryUrl = (
-  kind: string,
-  story: string
-): string =>
-  `iframe.html?selectedKind=${encode(kind)}&selectedStory=${encode(story)}`;
+// TODO: import from `wix-storybook-utils` or move getStoryUrl() to
+// `wix-storybook-utils` and import/export it here.
+const  WITH_EXAMPLES_PARAM_NAME = 'withExamples';
+
+export interface StoryUrlParams {
+  kind: string;
+  story: string;
+  withExamples?: boolean;
+}
+
+/**
+ * @deprecated
+ * @see createStoryUrl
+ */
+export const getStoryUrl = (kind: string, story: string): string =>
+  createStoryUrl({kind, story});
+
+/**
+ *
+ * @param {StoryUrlParams} params withExamples defaults to true
+ */
+export const createStoryUrl = (params: StoryUrlParams): string => {
+  const {kind, story, withExamples = true} = params;
+  return `iframe.html?selectedKind=${encode(kind)}&selectedStory=${encode(story)}${withExamples ? `&${WITH_EXAMPLES_PARAM_NAME}=` : ''}`;
+};
 
 export const scrollToElement = (element: ElementArrayFinder) =>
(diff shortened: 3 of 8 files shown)


packages/wix-storybook-utils/package.json
```diff
@@ -55,6 +55,7 @@
     "import-path": "latest",
     "jsx-to-string": "git+https://github.com/amiryonatan/jsx-to-string.git#release_branch",
     "loader-utils": "^1.1.0",
+    "query-string": "^5.1.0",
```
argshook: i would prefer not to add two new dependencies just for `window.location.search.includes('specificEnoughString')`

packages/wix-ui-test-utils/src/protractor/protractor-helpers.ts
```diff
@@ -8,11 +8,31 @@ import {
...
+ */
+export const getStoryUrl = (kind: string, story: string): string =>
+  createStoryUrl({kind, story});
+
+/**
+ *
+ * @param {StoryUrlParams} params withExamples defaults to true
+ */
+export const createStoryUrl = (params: StoryUrlParams): string => {
+  const {kind, story, withExamples = true} = params;
```
argshook: this can be in line `32`?

packages/wix-ui-test-utils/test/protractor-helpers.spec.ts
```diff
@@ -0,0 +1,31 @@
...
+    });
+
+    it('Should create story url with options', () => {
+      expect(createStoryUrl({kind, story}))
+      .toEqual(`${EXPECTED_URL_BASE}&withExamples=`);
+    });
+
+    it('Should create story url without examples', () => {
+      expect(createStoryUrl({kind, story, withExamples: false}))
+      .toEqual(`${EXPECTED_URL_BASE}`);
```
argshook: unnecessary template string

packages/wix-ui-test-utils/src/protractor/protractor-helpers.ts
```diff
@@ -8,11 +8,31 @@ import {
...
+  return `iframe.html?selectedKind=${encode(kind)}&selectedStory=${encode(story)}${withExamples ? `&${WITH_EXAMPLES_PARAM_NAME}=` : ''}`;
```
argshook: why `=` after `withExamples`?
```
export const createStoryUrl = ({kind, story, withExamples = true}: StoryUrlParams): string => {
  const examplesQuery = withExamples ? `&${WITH_EXAMPLES_PARAM_NAME}` : '';
  return `iframe.html?selectedKind=${encode(kind)}&selectedStory=${encode(story)}${examplesQuery}`;
}

File: packages/wix-storybook-utils/src/StoryNew/index.js [hunk 5]
```diff
@@ -6,11 +6,13 @@ import Markdown from '../Markdown';
 import CodeBlock from '../CodeBlock';
 import AutoExample from '../AutoExample';
 import AutoDocs from '../AutoDocs';
-
+import * as queryString from 'query-string';
```
argshook: why not handle this in wix-ui-test-utils (and later move to wix-storybook-utils)?

```
import { isURLWithExamples } from 'wix-ui-test-utils'
// then later
{ isURLWithExamples(window.location.search) && examples }
```

then no need for `IS_EXAMPLES_VISIBLE_PARAM_NAME`

packages/wix-ui-test-utils/src/protractor/protractor-helpers.ts
```diff
@@ -8,11 +8,31 @@ import {
...
+}
+
+/**
+ * @deprecated
+ * @see createStoryUrl
+ */
+export const getStoryUrl = (kind: string, story: string): string =>
+  createStoryUrl({kind, story});
+
+/**
```
argshook: `export const isURLWithExamples = url => url.includes(WITH_EXAMPLES_QUERY)`

packages/wix-storybook-utils/package.json
hepiyellow: It is more safe to parse the queryString first.
If you use `includes()`, then you might catch a param value instead of a param name.
If I am looking for `withExamples` param name to exist, it would also catch:
e.g `?storyName=My%20story%20withExmaples`

packages/wix-ui-test-utils/src/protractor/protractor-helpers.ts
hepiyellow: What do you mean?

packages/wix-ui-test-utils/src/protractor/protractor-helpers.ts
hepiyellow: I just saw that the browser adds `=` when loading a page with such URL (a URL with a param name and no value and no `=`). So I thought it would be more consistent.

packages/wix-storybook-utils/src/StoryNew/index.js
hepiyellow: I see it such that `wix-storybook-utils` is giving a service (render utils),
and `wix-ui-test-utils` is just a convenience wrapper for testing, so people would not have to explicitly import stuff from `wix-storybook-utils` in their tests.

So it is as if `wix-storybook-utils` is a library, and `wix-ui-test-utils` consumes it. So it sounds weird that a library would import from it's consumer.

Also, if we are going to move it later to `wix-storybook-utils`, then what do we gain by differing it?

packages/wix-ui-test-utils/src/protractor/protractor-helpers.ts
argshook: you should be able to destructure `params` in line 32, i think?

packages/wix-storybook-utils/package.json
argshook: that's why `'specificEnoughString'`. make it `'EXAMPLES_VISIBLE'` and you don't need two more dependencies.

packages/wix-ui-test-utils/src/protractor/protractor-helpers.ts
hepiyellow: Yeah, if you still think that `wix-storybook-utils` should import from `wix-ui-test-utils`,... then thats the way to do it.

packages/wix-storybook-utils/package.json
hepiyellow: What is the "pain" in 2 more dependencies (which is actually 1, the other is just @types)? which hurts more than having a clean robust code.

packages/wix-ui-test-utils/test/protractor-helpers.spec.ts
hepiyellow: fixed

packages/wix-ui-test-utils/src/protractor/protractor-helpers.ts
hepiyellow: fixed

packages/wix-storybook-utils/package.json
argshook: wdym? more dependencies === more headache. if this were used for critical production endpoint or something, sure, but here it's merely for specific e2e cases. let's add it if you insist, but i can't really see how it helps

packages/wix-storybook-utils/src/StoryNew/index.js
argshook: my idea is to have everything about implementation of `getStoryUrl` in one place, either `wix-storybook-utils` or `wix-ui-test-utils`. which is better i dunno, probably `wix-storybook-utils`, but we can't move there without updating many tests, so i'd say `wix-ui-test-utils` for now.

packages/wix-ui-test-utils/src/protractor/protractor-helpers.ts
argshook: i'd just prefer it all being in one place, cause now there's `WITH_EXAMPLES_PARAM_NAME` and also `IS_EXAMPLES_VISIBLE_PARAM_NAME`, string created in one package, but queried in another

packages/wix-storybook-utils/package.json
hepiyellow: thanks. I'm keeping it (-:

packages/wix-storybook-utils/src/StoryNew/index.js
hepiyellow: Are you still against having `wix-ui-test-utils` import the `getStoryUrl` from `wix-storybook-utils`?
It a circular dependency:
```
 wix-storybook-utils:devDependencies:
    {..wix-ui-test-utils...}:dependencies:
         {...wix-storybook-utils}
```
Which I think is not a real issue.
That will solve the thing.

packages/wix-ui-test-utils/src/protractor/protractor-helpers.ts
hepiyellow: you are right. it should all be in one place.

packages/wix-storybook-utils/src/StoryNew/index.js
argshook: i guess that's fine, we already have few cases like this. for example `wix-style-react` uses `wix-storybook-utils` which uses `wix-style-react`, no problems yet

hepiyellow:
Hi, @argshook , can you help me fix this TS error.
I am getting a weird TS error when building `wix-ui-test-utils`:
```
test/protractor-helpers.spec.ts(1,22): error TS2305: Module '"/Users/erezm/Projects/wix-ui/packages/wix-ui-test-utils/src/protractor/protractor-helpers"' has no exported member 'createStoryUrl'.
```

I did locally, `npm link` in `wix-storubook-utils`.
and `npm link wix-storybook-utils` in `wix-ui-test-utils`.

Why is the error on `protractor-helpers`? The module I am importing from is in `wix-storybook-utils`.

hepiyellow:
replaced with #325

PR #307, opened 2018-03-11
wix-storybook-utils: AutoExampleDriver.reset() - Remove click on body

hepiyellow:
For some reason, in more than 1 case, I encountered, that clicking on the body, actually make the tested component focused.

Remove this, until further investigation.

### DIFF
--- a/packages/wix-storybook-utils/src/AutoExample/protractor.driver.js
+++ b/packages/wix-storybook-utils/src/AutoExample/protractor.driver.js
@@ -51,7 +51,6 @@ module.exports = {
     return browser.executeScript(script, props);
   },
   reset: () => {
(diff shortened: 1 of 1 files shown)


hepiyellow:
After some research, it seems that the only solution for reseting the tab order (without reloading the page),
is to focus on a wrapper element.

So we need to a focusable wrapper element in the AutoExample iframe.html, which is focused onLoad. And the autoExampleDriver.reset() focuses on it.

A more general solution would be one that the story renders <AutoExample>, so we can wrap it with anything we want. 
Maybe we can add a prop to the story, like `renderExample`, which will be a function which receives an <AutoExample> ReactElement.

PR #312, opened 2018-03-13
ToggleSwitch - Fixing focus to work cross browser

giladartzi:


packages/wix-ui-core/src/components/ToggleSwitch/styles.ts
```diff
@@ -33,8 +33,8 @@ export const styles = (theme: ToggleSwitchTheme) => {
       '-webkit-tap-highlight-color': 'rgba(0, 0, 0, 0)',
 
       '& > input[type=checkbox]': {
-        width: 0,
-        height: 0,
+        width: '1px',
```
lbelinsk: @alex5il nite that gilad changed that because a bug in safari: when the size is 0 the focus isn't working

PR #316, opened 2018-03-16
Add stylable for Video

kupriyanenko:


packages/wix-ui-core/src/components/Video/Video.tsx
```diff
@@ -44,12 +54,24 @@ const mapPropsToMethods = {
   }
 };
 
+const playButton = (
```
shlomitc: Why is a custom playButton needed? and why is it part of the core?
If this button is something worth adding to the core, maybe it can be part of `playable`

packages/wix-ui-core/src/components/Video/Video.tsx
```diff
@@ -146,9 +178,39 @@ export class Video extends React.PureComponent<VideoProps, VideoState> {
...
+    };
+
     return (
-      <div ref={el => this.containerRef = el}></div>
+      <div
+        ref={el => this.containerRef = el}
+        id={id}
+        style={{width, height, ...style}}
+        {...styles('root', {}, this.props)}>
+        {!this.state.hasBeenPlayed && (
```
shlomitc: same idea here, why is this logic in `wix-ui-core` and not in `playable`?

packages/wix-ui-core/src/components/Video/Video.tsx
kupriyanenko: 1. In playable we don't have a title for the cover, but in the santa, we should have. 
2. Because of this code part of `wix-ui`, I can use stylable and I have fully forking theming for the santa.

packages/wix-ui-core/src/components/Video/Video.st.css
```diff
@@ -0,0 +1,43 @@
+.root {
+    position: relative;
+    z-index: 0;
```
JoMarton: `z-index` isn't advisable.
I understand you use it for overlaying the cover on top of the player? maybe switch visibility of the player with style states?
I'm concerned that in the dynamic editor / viewer environment that might cause unexpected behavior with z-index wars...

packages/wix-ui-core/src/components/Video/Video.tsx
```diff
@@ -1,12 +1,18 @@
 import * as React from 'react';
-import {string, number, func, bool, array, oneOfType, Requireable} from 'prop-types';
+import {string, number, func, object, bool, array, oneOfType, Requireable} from 'prop-types';
 import {create, VIDEO_EVENTS, ENGINE_STATES} from 'playable';
+import styles from './Video.st.css';
 
 export interface VideoProps {
+  id?: string;
+  style?: React.CSSProperties;
```
JoMarton: seems like having `style` and also `width, height` in props is redundant, unless there is a specific use case

JoMarton:
Just to clarify and make sure - `wix-ui-core` is a "generic" components library that should be used by many consumers. If you have some specific features that relate to only one type of consumer it should be implemented in the specific themed library.

packages/wix-ui-core/src/components/Video/Video.tsx
kupriyanenko: santa pass `style`, `width` and `height` to my component, if I ignore it - it doesn't work. Pagination component doing the same.

kupriyanenko: you can check how I use it: https://github.com/wix-private/santa/blob/70dc649c2b730278d9fb4bf38c041656ec139ee5/packages/videoPlayer/src/main/videoPlayer.js#L25, similar to `Pagination`

shlomitc:
1. It sounds to me that adding a cover is a good feature anyway, so why not adding it to `playable`? It also means you need to duplicate the icon which doesn't sound to me like a good duplication.

2. We will soon add a guide on how to apply icons. Having them in the code like that is not the best approach. in the meantime, moving the icon to a separate file will be better

3. might want to move the entire cover behavior to a separate component, to separate the concerns

kupriyanenko:
@shlomitc icon moved. The main issue with cover - if the cover will be part of playable - it's impossible to use stylable to style it in the santa.

packages/wix-ui-core/src/components/Video/Video.st.css
kupriyanenko: ok, we will change it on playable level and then here.

packages/wix-ui-core/src/components/Video/Video.tsx
JoMarton: I could be wrong but as far as I saw, santa implicitly passes on to your component a style object (no need to explicitly pass width and height). 
I know pagination does the same thing and that also can be refactored out... 
anyway as a core component it should ignore specific requirements from consumers like santa. any integration need should be addressed in the themed library

kupriyanenko:
@shlomitc @JoMarton I can move all cover logic to `wix-ui-santa`, do you think it will be better? because cover should be stylable for the santa.

JoMarton:
I am on the fence regarding the cover. if it is useful to everyone, and must be stylable - lets keep it in the core. if not - just as easy to implement in `ui-santa`

shlomitc:
@kupriyanenko - the reason for not being able to customize it internally is because playable doesn't support custom `classes`?

packages/wix-ui-core/src/components/Video/Video.tsx
kupriyanenko: removed style (for that in `santa-editor` I need to set `styleId` like class). But `width` and `height` need to leave, it's part of basic component API.

kupriyanenko:
@shlomitc yes, and I don't want to add it now. If we will add it - we will need to freeze our internal structure for HTML/CSS.

shlomitc:
ok. so my only last concern is that there is relatively a lot of css for it.

Usually, the css for core components is very slim and acts as resets, but this is the first example of a more complicated core component.

@lbelinsk - do you have a better suggestion regarding the CSS?

kupriyanenko:
@shlomitc I think at least I can remove CSS related to color and font, the rest related to layouting.

PR #317, opened 2018-03-18
Remove InputWithAffixes, add affixes to the normal Input

alisey:


packages/wix-ui-core/src/components/Input/Input.tsx
```diff
@@ -1,171 +1,110 @@
...
+          'root',
+          {disabled: this.props.disabled, error: this.props.error, focus: this.state.focus},
+          this.props
+        )}
+      >
+        {this.props.prefix}
+        <input
+          ref={input => this.input = input}
+          autoComplete={this.props.autoComplete}
+          autoFocus={true}
```
amiryonatan: Why is it always true ? 
Should we spread the props ?

packages/wix-ui-core/src/components/Input/Input.tsx
```diff
@@ -1,171 +1,110 @@
...
-};
+export interface InputProps {
+  className?: string;
+  error?: boolean;
+  prefix?: JSX.Element;
+  style?: React.CSSProperties;
+  suffix?: JSX.Element;
 
-const NUMBER_REGEX = /^[\d.,\-+]*$/;
+  // Props passed down to the native input, add more as needed.
```
amiryonatan: I don't understand why don't we just use React Input Props ?

packages/wix-ui-core/src/components/Input/Input.tsx
```diff
@@ -1,171 +1,110 @@
...
+    onChange: () => null
+  };
 
-    if (disabled ||
-        readOnly ||
-        (type === 'number' && !(NUMBER_REGEX.test(e.target.value)))) {
-      return;
-    }
+  state: InputState = {
+    focus: false,
```
amiryonatan: I really don't like it, but i guess there's no way around it

packages/wix-ui-core/src/components/Input/Input.tsx
```diff
@@ -1,171 +1,110 @@
...
 
-    this.props.onChange && this.props.onChange(e);
-  }
+  private input: HTMLInputElement;
 
-  focus() {
-    this.inputRef.focus();
+  render() {
+    return (
+      <div
```
amiryonatan: Should we still render the div even if there is no prefix / suffix ?

packages/wix-ui-core/src/components/Input/Input.tsx
```diff
@@ -1,171 +1,110 @@
...
+        />
+        {this.props.suffix}
+      </div>
+    );
   }
 
-  blur() {
-    this.inputRef.blur();
-  }
+  focus() { this.input.focus(); }
```
amiryonatan: Why are those functions necessary ?

packages/wix-ui-core/src/components/Input/Input.tsx
```diff
@@ -1,171 +1,110 @@
 import * as React from 'react';
+import * as PropTypes from 'prop-types';
```
amiryonatan: Import only the types u need

packages/wix-ui-core/src/components/Input/Input.tsx
```diff
@@ -1,171 +1,110 @@
...
-    .keys(props)
-    .filter(key => key.startsWith(aria))
-    .map(key => ariaAttribute[`${aria}-${key.substr(aria.length).toLowerCase()}`] = props[key]);
-  return ariaAttribute;
-};
+export interface InputProps {
+  className?: string;
+  error?: boolean;
+  prefix?: JSX.Element;
+  style?: React.CSSProperties;
```
amiryonatan: why do we allow style ?

packages/wix-ui-core/src/components/Input/Input.tsx
```diff
@@ -1,171 +1,110 @@
...
-  const aria = 'aria';
-  const ariaAttribute = {};
-  Object
-    .keys(props)
-    .filter(key => key.startsWith(aria))
-    .map(key => ariaAttribute[`${aria}-${key.substr(aria.length).toLowerCase()}`] = props[key]);
-  return ariaAttribute;
-};
+export interface InputProps {
+  className?: string;
```
amiryonatan: is necessary ?

packages/wix-ui-core/src/components/Input/Input.tsx
```diff
@@ -1,171 +1,110 @@
...
 export class Input extends React.Component<InputProps, InputState> {
-  static displayName = 'Input';
-
-  private inputRef: HTMLInputElement;
-
-  static defaultProps = {
-    type: 'text'
+  static propTypes = {
+    /** Wrapper class name */
+    className: PropTypes.string,
```
amiryonatan: Should match the interface

packages/wix-ui-core/src/components/Input/Input.tsx
amiryonatan: export type InputProps = React.InputHTMLAttributes<HTMLInputElement> & {
  prefix?: React.ReactNode;
  suffix?: React.ReactNode;
};

packages/wix-ui-core/src/components/Input/Input.tsx
alisey: Yes, otherwise styling becomes inconsistent, you'd need different CSS rules for when `<input>` is the root, and for when it's not.

packages/wix-ui-core/src/components/Input/Input.tsx
alisey: It's just my personal preference, it's also how they have it in official React docs.

packages/wix-ui-core/src/components/Input/Input.tsx
alisey: yes

packages/wix-ui-core/src/components/Input/Input.tsx
alisey: I'm not sure if we should list standard input props in prop types. We could simply say in the docs "supports standard input props", but that wouldn't be totally true since we only support subset of those. But also I don't want to add a bunch of boilerplate.

PR #323, opened 2018-03-19
Slider

InterAl:


packages/wix-ui-core/src/components/Slider/Slider.st.css
```diff
@@ -1,5 +1,5 @@
 .root {
-    -st-states: layout(string), tickMarksPosition(string), dir(string), showTicks, disabled;
+    -st-states: layout(enum(vertical, horizontal)), tickMarksPosition(enum(default, middle, across)), dir(enum(ltr, rtl)), showTicks, disabled;
```
lbelinsk: why not orientation instead of layout?

InterAl: Changed to orientation.

PR #324, opened 2018-03-20
Label component

Zombiefruit:
Added label component to wix-ui-core.


packages/wix-ui-core/src/components/Label/Label.driver.ts
```diff
@@ -0,0 +1,12 @@
+export const labelDriverFactory = ({element}) => {
+  return {
+    /** check if element exists */
+    exists: () => !!element,
+    /** get the rendered content */
+    getChildren: () => element.innerHTML,
+    /** get the id of the component */
+    getId: () => element.getAttribute('id'),
+    /** get the "for" attribute of the component */
+    getFor: () => element.getAttribute('for')
```
lbelinsk: a bit confusing :)
Maybe getForAttribute?

packages/wix-ui-core/src/components/Label/Label.protractor.driver.ts
```diff
@@ -0,0 +1,6 @@
+export const labelDriverFactory = component => ({
+    /** returns the component element */
+    element: () => component,
+    /** returns the component label */
+    getLabelContent: () => component.getText()
```
lbelinsk: Since in wix-style-react we don't want to deprecate things for no good reason, let's stick to the method  getLabelText  https://github.com/wix/wix-style-react/blob/master/src/Label/Label.driver.js#L8

packages/wix-ui-core/src/components/Label/Label.spec.tsx
```diff
@@ -0,0 +1,32 @@
...
+  const createDriver = createDriverFactory(labelDriverFactory);
+
+  it('Renders with default values', async () => {
+    const label = createDriver(<Label />);
+
+    expect(label.exists()).toBe(true);
+  });
+
+  it('Renders children', async () => {
+    const label = createDriver(<Label><div>HELLO</div></Label>);
```
lbelinsk: The label children should only be string imo. don't you think?

packages/wix-ui-core/src/components/Label/Label.spec.tsx
```diff
@@ -0,0 +1,32 @@
+import * as React from 'react';
+import {labelDriverFactory} from './Label.driver';
+import {createDriverFactory} from 'wix-ui-test-utils/driver-factory';
+import {Label} from './Label';
+
+describe('Label', () => {
+  const createDriver = createDriverFactory(labelDriverFactory);
+
+  it('Renders with default values', async () => {
```
lbelinsk: Instead of having this test, do it while testing the testkit with `isEnzymeTestkitExist`...

packages/wix-ui-core/src/components/Label/LabelStyle.st.css
```diff
@@ -0,0 +1,25 @@
+:import {
+  -st-from: "./Label.st.css";
+  -st-default: Label;
+}
+
+:vars {
+  fontFamily: Arial;
```
lbelinsk: How did you decide that this is the common one?

packages/wix-ui-core/src/components/Label/LabelStyle.st.css
```diff
@@ -0,0 +1,25 @@
...
+  fontSize: 12px;
+  lineHeight: normal;
+  color: #000;
+  textTransform: none;
```
lbelinsk: aren't these are the default css valuse?
We are starting to rollback on the idea of mixins, so maybe we can drop some stuff here

packages/wix-ui-core/src/components/Label/Label.spec.tsx
```diff
@@ -0,0 +1,32 @@
...
+  it('takes an id prop', async () => {
+    const label = createDriver(<Label id="hey" />);
+
+    expect(label.getId()).toBe('hey');
+  });
+
+  it('takes an htmlFor prop', async () => {
+    const label = createDriver(<Label for="hey" />);
+
+    expect(label.getFor()).toBe('hey');
```
lbelinsk: missing testkit tests

packages/wix-ui-core/src/components/Label/Label.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+/**
+ * Label
+ */
+export const Label: React.SFC<LabelProps> = props => {
+  return <label {...style('root', {}, props)} htmlFor={props.for} id={props.id}>{props.children}</label>;
+};
+
+Label.propTypes = {
+  /** Label children */
```
lbelinsk: no need to say "Label" as it is a documentation for label which is obvious

packages/wix-ui-core/src/components/Label/Label.e2e.ts
```diff
@@ -12,6 +12,6 @@ describe('Label', () => {
     const driver = labelTestkitFactory({dataHook});
 
     return waitForVisibilityOf(driver.element(), 'Cannot find Label')
-      .then(() => expect(driver.getLabelContent()).toBe('HELLO'));
+      .then(() => expect(driver.getLabelText()).toBe('hello'));
```
lbelinsk: you don't have this method. only getChildren in the driver.
You changed it only in the protractor driver

lbelinsk: check the wsr drivers here: https://github.com/wix/wix-style-react/tree/master/src/Label

packages/wix-ui-core/src/components/Label/Label.driver.ts
```diff
@@ -7,6 +7,6 @@ export const labelDriverFactory = ({element}) => {
     /** get the id of the component */
```
lbelinsk: should be also renamed to getLabelText.
see here:
https://github.com/wix/wix-style-react/tree/master/src/Label

PR #329, opened 2018-03-21
add native calc mixin

lbelinsk:


packages/wix-ui-core/src/mixins/calc/calc.spec.ts
```diff
@@ -0,0 +1,24 @@
+import * as calc from './index';
+
+describe('calc mixin', () => {
+  it('should calculate an expression with no units', () => {
+    expect(calc('12 / 2')).toBe('6');
+  });
+
+  it('should inferr the unit typed value when have values with some unit and some without units', () => {
```
alisey: typo, infer

packages/wix-ui-core/src/mixins/calc/calc.spec.ts
```diff
@@ -0,0 +1,24 @@
...
+    expect(calc('12% / 2')).toBe('6%');
+    expect(calc('12 / 2%')).toBe('6%');
+  });
+
+  it('should calculate an expression with px units', () => {
+    expect(calc('12px / 2px')).toBe('6px');
+  });
+
+  it('should calculate an expression with px units', () => {
```
alisey: "with percent units"

packages/wix-ui-core/src/mixins/calc/calc.spec.ts
```diff
@@ -0,0 +1,24 @@
...
+    expect(calc('12% / 2%')).toBe('6%');
```
alisey: Interesting use case, ideally this should throw, because division by % is not valid in native calc https://jsfiddle.net/hpmp5e8L/
But for now we can ignore it.

packages/wix-ui-core/package.json
```diff
@@ -20,7 +20,7 @@
     "url": "git+https://github.com/wix/wix-ui.git"
   },
   "scripts": {
-    "build": "haste build && import-path --path src/components --dts && import-path --path src/utils --dts && stc --srcDir=\"./dist/src\" --diagnostics",
+    "build": "haste build && import-path --path src/components --dts && import-path --path src/utils --dts && stc --srcDir=\"./dist/src\" --diagnostics && babel src/mixins -d dist/src/mixins",
```
shlomitc: It became enormous, can you please split it into subtasks or a separate script?

PR #331, opened 2018-03-22
Checkbox: move icons from CSS to props

alisey:


packages/wix-ui-core/src/components/Checkbox/Checkbox.tsx
```diff
@@ -20,8 +21,9 @@ export interface CheckboxProps {
   onChange?: React.EventHandler<OnChangeEvent | OnClickEvent | OnKeydownEvent>;
   id?: string;
   tabIndex?: number;
-  tickIcon?: React.ReactNode;
-  indeterminateIcon?: React.ReactNode;
+  checkedIcon?: JSX.Element;
+  uncheckedIcon?: JSX.Element;
```
danielstro: I can't see the default value for this. If there is no prop it will not render an icon. Is this on purpose?

alisey: Yes

PR #337, opened 2018-03-25
Pagination - adding minimum width measurement

JoMarton:


packages/wix-ui-core/src/components/Pagination/Pagination.tsx
```diff
@@ -146,11 +159,64 @@ export class Pagination extends React.Component<PaginationProps, PaginationState
...
+      const pageStripNode = compNode.querySelector(`#${this.getId('pageStrip')}`);
+      const slashNode = compNode.querySelector(`#${this.getId('slash')}`);
+      const totalPagesNode = compNode.querySelector(`#${this.getId('totalPages')}`);
+      const pageInputNode = compNode.querySelector(`#${this.getId('pageInput')}`);
+
+      let selectionMinWidth = 0;
+
+      if (pageStripNode) {
+          // means we're in "pages" pagination mode
+          selectionMinWidth = getMinWidth(pageStripNode);
```
alisey: Not sure how this works, unless we specify min-width for the page strip in CSS, this should always return zero. Does it really return anything but zero?

JoMarton: I'm relying on the minWidth CSS...

packages/wix-ui-core/src/components/Pagination/root-min-width.ts
```diff
@@ -1,49 +1,36 @@
...
-    getWidthWithMargins(lastBtnNode);
+    const navButtonsMinWidth = getWidthWithMargins(getById('navButtonNext')) +
+    getWidthWithMargins(getById('navButtonPrevious')) +
+    getWidthWithMargins(getById('navButtonFirst')) +
+    getWidthWithMargins(getById('navButtonLast'));
 
     let selectionMinWidth = 0;
 
     if (paginationMode === 'pages') {
         // means we're in "pages" pagination mode
```
alisey: the comment seems redundant now when the condition is literally `if (paginationMode === 'pages')`

PR #358, opened 2018-03-29
<Checkbox /> - Fix types and implementation

danielstro:
Part of issue #259 
- Completely changed the checkbox types to extend native input types.
- Removed proptypes, it is not needed now because checkbox doesn't have a storybook demo.
- Changed the inner implementation to use a label instead of a div
- Had to change the driver to fit the new implementation
- Added more tests to the e2e


packages/wix-ui-core/src/components/Checkbox/Checkbox.tsx
```diff
@@ -1,111 +1,58 @@
 import * as React from 'react';
 import style from './Checkbox.st.css';
-import {bool, func, string, number, array, node} from 'prop-types';
+import {noop} from 'lodash';
```
JoMarton: I would prefer using our own noop instead of importing from lodash just for that

packages/wix-ui-core/src/components/Checkbox/Checkbox.protractor.driver.ts
```diff
@@ -1,10 +1,13 @@
...
     /** Returns the component instance */
     element: () => component,
     /** Simulates a click on the component */
     click: () => component.click(),
     /** Indicates whether the component is disabled or not */
     isDisabled: () => component.getAttribute('disabled') === '',
-    /** Indicates whether the component is checked */
-    isChecked: () => component.getAttribute('aria-checked') !== 'false'
-  });
+    /** returns a boolean indicating if the toggleSwitch is checked */
```
alisey: toggleSwitch => checkbox

packages/wix-ui-core/src/components/Checkbox/Checkbox.spec.tsx
```diff
@@ -151,15 +141,11 @@ describe('Checkbox', () => {
     it('focuses the native input if autofocus is passed', () => {
       const checkbox = createDriver(<Checkbox autoFocus />);
 
+      expect(document.activeElement).toEqual(checkbox.input());
```
alisey: Isn't this the same as two lines below?

packages/wix-ui-core/src/components/Checkbox/Checkbox.tsx
```diff
@@ -131,42 +78,31 @@ export class Checkbox extends React.Component<CheckboxProps, CheckboxState> {
...
-      this.handleChange(e);
-    }
-  }
-
-  private handleClick: React.MouseEventHandler<HTMLDivElement> = e => {
-    this.handleChange(e);
-    this.checkbox && this.checkbox.focus();
-    this.setState({isFocused: false});
+  private handleMouseDown: React.MouseEventHandler<HTMLElement> = (e) => {
+    e.preventDefault();
```
alisey: Are you sure you need to prevent default? If yes, then add a comment why.

packages/wix-ui-core/src/components/Checkbox/Checkbox.tsx
```diff
@@ -131,42 +78,31 @@ export class Checkbox extends React.Component<CheckboxProps, CheckboxState> {
...
+    this.focusedByMouse = !this.state.isFocused;
```
alisey: Need to check the mouse button.

packages/wix-ui-core/src/components/Checkbox/Checkbox.tsx
```diff
@@ -131,42 +78,31 @@ export class Checkbox extends React.Component<CheckboxProps, CheckboxState> {
...
+    this.setState({isFocused: true});
```
alisey: Not sure if this is needed. In the toggle switch the mousedown handler simply sets a flag without changing state: https://github.com/wix/wix-ui/blob/master/packages/wix-ui-core/src/components/ToggleSwitch/ToggleSwitch.tsx#L114

packages/wix-ui-core/src/components/Checkbox/Checkbox.tsx
```diff
@@ -131,42 +78,31 @@ export class Checkbox extends React.Component<CheckboxProps, CheckboxState> {
...
   private handleInputBlur: React.FocusEventHandler<HTMLInputElement> = () => {
-    this.state.isFocused && this.setState({isFocused: false});
+    this.state.isFocused && this.setState({isFocused: false, focusVisible: false});
+    this.focusedByMouse = false;
   }
 
   private handleInputFocus: React.FocusEventHandler<HTMLInputElement> = () => {
-    !this.state.isFocused && this.setState({isFocused: true});
+    this.setState({isFocused: true, focusVisible: !this.focusedByMouse});
+    this.focusedByMouse = false;
```
alisey: shouldn't be needed

packages/wix-ui-core/src/components/Checkbox/Checkbox.tsx
```diff
@@ -1,111 +1,58 @@
...
-}
-
-export interface CheckboxProps {
-  checked?: boolean;
-  disabled?: boolean;
-  onChange?: React.EventHandler<OnChangeEvent | OnClickEvent | OnKeydownEvent>;
-  id?: string;
-  tabIndex?: number;
-  checkedIcon?: JSX.Element;
+export interface CheckboxProps extends React.InputHTMLAttributes<HTMLElement> {
```
alisey: The interface advertises that all input props are accepted, but the props are actually not being spread onto the input.

packages/wix-ui-core/src/components/Checkbox/Checkbox.tsx
```diff
@@ -1,111 +1,58 @@
...
     const focus = this.state.isFocused;
 
     return (
-      <div {...style('root', {checked, disabled, focus, readonly, error, indeterminate}, this.props) }
-        onClick={this.handleClick}
-        onKeyDown={this.handleKeydown}
-        role="checkbox"
-        aria-checked={this.props.indeterminate ? 'mixed' : this.props.checked}
-      >
+      <label {...style('root', {checked, disabled, focus, readonly, error, indeterminate, 'focus-visible': this.state.focusVisible}, this.props) }
```
alisey: would prefer if focus and readonly were assigned directly from props/state, without intermediate variables that are not used anywhere else

packages/wix-ui-core/src/components/Checkbox/Checkbox.e2e.ts
```diff
@@ -2,16 +2,50 @@ import * as eyes from 'eyes.it';
...
     const driver = checkboxTestkitFactory({dataHook});
 
     return waitForVisibilityOf(driver.element(), 'Cannot find Checkbox');
   });
+
+  eyes.it('should toggle', () => {
+    const driver = checkboxTestkitFactory({dataHook});
+
+    return waitForVisibilityOf(driver.element(), 'Cannot find Checkbox')
+    .then(() => {
```
JoMarton: you can use `async await` for better readability if you like that better

packages/wix-ui-core/src/components/Checkbox/Checkbox.e2e.ts
```diff
@@ -2,16 +2,50 @@ import * as eyes from 'eyes.it';
...
+
+      driver.click();
+      expect(driver.checked()).toBeTruthy();
+
+      driver.click();
+      expect(driver.checked()).toBeFalsy();
+    });
+  });
+
+  eyes.it('should support accessiblility features', () => {
```
JoMarton: It seems to me like visually this test is equal to the one before it so you can make it only `it` instead of `eyes.it`

packages/wix-ui-core/src/components/Checkbox/Checkbox.e2e.ts
alisey: you don't even need async/await, take a look at these tests in pagination:
https://github.com/wix/wix-ui/blob/master/packages/wix-ui-core/src/components/Pagination/Pagination.e2e.ts#L45

there's some magic that makes it work without having to write async and await explicitly.

packages/wix-ui-core/src/components/Checkbox/Checkbox.tsx
```diff
@@ -1,111 +1,58 @@
...
-export interface CheckboxProps {
-  checked?: boolean;
-  disabled?: boolean;
-  onChange?: React.EventHandler<OnChangeEvent | OnClickEvent | OnKeydownEvent>;
-  id?: string;
-  tabIndex?: number;
-  checkedIcon?: JSX.Element;
+export interface CheckboxProps extends React.InputHTMLAttributes<HTMLElement> {
+  /** The onChange function will be called with a new checked value */
+  onChange?: React.EventHandler<OnChangeEvent>;
```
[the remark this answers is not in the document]
danielstro: Yes, but this is a different onChange, we add the "checked" property to it.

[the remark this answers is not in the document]
JoMarton: ah ok sorry didn't notice that

packages/wix-ui-core/src/components/Checkbox/Checkbox.tsx
danielstro: done

packages/wix-ui-core/src/components/Checkbox/Checkbox.tsx
danielstro: Checkbox is focused regardless of mouse button

packages/wix-ui-core/src/components/Checkbox/Checkbox.tsx
danielstro: If someone is holding the mouse button (Or clicking rapidly) focus is lost. This is here to fix that.

packages/wix-ui-core/src/components/Checkbox/Checkbox.tsx
```diff
@@ -1,172 +1,102 @@
...
-  state = {isFocused: false};
+  private focusedByMouse = false;
+  state = {isFocused: false, focusVisible: false};
 
   public render()  {
-    const {checked, disabled, readOnly: readonly, error, indeterminate, required} = this.props;
-    const focus = this.state.isFocused;
+    const {checked, disabled, error, indeterminate, indeterminateIcon, checkedIcon, uncheckedIcon,
+      //These variables were not meant to be used
+      onChange, children, className, ...rest} = this.props;
```
alisey: would rename rest to inputProps, and mention in the comment above the reason why we are creating variables that are not meant to be used.

packages/wix-ui-core/src/components/Checkbox/Checkbox.tsx
alisey: Add a comment explaining why we prevent default.

packages/wix-ui-core/src/components/Checkbox/Checkbox.tsx
alisey: Can it be simply `this.focusedByMouse = true`?

packages/wix-ui-core/src/components/Checkbox/Checkbox.tsx
alisey: Add a comment.

PR #365, opened 2018-04-02
Checkbox - removed spread

danielstro:
Temporary fix for wix-ui-backoffice.


packages/wix-ui-core/src/components/Checkbox/Checkbox.tsx
```diff
@@ -54,7 +52,16 @@ export class Checkbox extends React.Component<CheckboxProps, CheckboxState> {
             onFocus={this.handleInputFocus}
             onBlur={this.handleInputBlur}
             ref={ref => this.checkbox = ref}
-            {...inputProps}
+            //temp fix
```
shlomitc: do you need to change also the  `CheckboxProps` interface?

danielstro: I don't think so, the interface didn't have these props originally so I see no reason to add them.

PR #374, opened 2018-04-04
VBox Component

Zombiefruit:
Implemented a stylable version of the VBox in wix-ui-core.


packages/wix-ui-core/yarn-error.log
```diff
@@ -0,0 +1,160 @@
+Arguments:
```
lbelinsk: Don't add this file. you can add it to  .gitigmore file

Zombiefruit: Already merged a PR today for the gitignore, forgot to remove the originally committed file in VBox

PR #376, opened 2018-04-04
Pagi- adding minHeight along to minWidth

JoMarton:


packages/wix-ui-core/src/components/Pagination/root-min-width.ts
```diff
@@ -15,22 +16,35 @@ export const measureAndSetRootMinWidth = (compNode: HTMLElement, paginationMode:
         parseInt(style.marginLeft, 10);
     };
 
+    const getHeightWithMargins = (elmnt: HTMLElement): number => {
```
alisey: `elmnt` looks kinda cryptic, how about `element`, `elem`, `node`?

PR #378, opened 2018-04-04
Slider

InterAl:


packages/wix-ui-core/src/components/Slider/Slider.st.css
```diff
@@ -1,117 +1,125 @@
...
 }
 
 .root:dir(rtl) .trackFill {
   right: 0;
 }
 
 .root:showTicks .inner {
-    height: 88.888%;
-    width: 100%;
+  height: calc(100% - value(tickMarksGap));
```
lbelinsk: IE11 doesn't support calc very well. you should import the calc mixin
https://github.com/wix/wix-ui/blob/60f224d575d53d918d95f3afdd096a933a296de9/packages/wix-ui-core/src/components/ToggleSwitch/ToggleSwitchLayout.st.css#L8

packages/wix-ui-core/src/components/Slider/Slider.tsx
```diff
@@ -15,15 +15,18 @@ export interface SliderProps {
...
+  tickMarksPosition?: 'normal' | 'middle' | 'across';
+  tickMarksShape?: 'none' | 'line' | 'dot';
   tooltipPrefix?: string;
   tooltipSuffix?: string;
   trackSize?: number;
   thumbShape?: 'circle' | 'square' | 'rectangle' | 'bar';
   disabled?: boolean;
+  readOnly?: boolean;
   dir?: string;
+  previewState?: string;
```
lbelinsk: previewState feature should sit in wix-ui-santa, not in the core.
Check Pagination component.

packages/wix-ui-core/src/components/Slider/Thumb.tsx
```diff
@@ -47,7 +47,7 @@ export function getThumbSize(shape: string, ...rest) {
 class CircleThumb extends React.PureComponent<any> {
   render() {
     return (
-      <div {...pStyle('thumbShape', {shapeType: 'circle'})}/>
+      <div {...pStyle('thumbShape', {shapeType: 'circle'})} style={{borderRadius: '50%'}}/>
```
lbelinsk: maybe not inline?

packages/wix-ui-core/src/components/Slider/Slider.tsx
InterAl: Not in this case, as it determines whether the tooltip should remain opened or not. wix-ui-santa can't control that.

packages/wix-ui-core/src/components/Slider/Slider.st.css
InterAl: Done.

packages/wix-ui-core/src/components/Slider/Thumb.tsx
InterAl: In this case I prefer it inlined, as I want it to override any CSS value defined in santa.

PR #381, opened 2018-04-05
TextLink: add basic functionality for TextLink component

xmityaz:


argshook:
gonna be `TextLink` and not just `Link`? @shlomitc

lbelinsk:
I prefer `Link`

shlomitc:
so `Link` it is - please change the name of the component

packages/wix-ui-core/src/components/TextLink/TextLink.tsx
```diff
@@ -0,0 +1,68 @@
...
+        )}
+        onFocus={this.handleFocus}
+        onBlur={this.handleBlur}
+      >
+        {children}
+      </a>
+    );
+  }
+
+  private handleFocus: React.FocusEventHandler<HTMLElement> = event => {
```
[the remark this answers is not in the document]
hepiyellow: This basic focus handling is ok. We'll make it focusable later.

packages/wix-ui-core/src/components/TextLink/TextLink.tsx
```diff
@@ -0,0 +1,68 @@
+import * as React from 'react';
+import * as PropTypes from 'prop-types';
+import style from './TextLink.st.css';
+import {Text} from '../Text/index';
```
hepiyellow: remove unused import

packages/wix-ui-core/src/components/TextLink/TextLink.tsx
```diff
@@ -0,0 +1,68 @@
...
+import {createHOC} from '../../createHOC';
```
hepiyellow: You should use this when exporting the component

packages/wix-ui-core/src/components/TextLink/TextLink.tsx
```diff
@@ -0,0 +1,68 @@
...
+    focus: false,
+  };
+
+  render() {
+    const {focus} = this.state;
+    const {href, disabled, target, children} = this.props;
+
+    return (
+      <a
+        href={disabled ? 'javascript:void(0);' : href}
```
hepiyellow: Did you try using `null` instead of `javascript:void(0)`?
I think it will render the `<a>` tag without the `href` property. Which will remove the default underline from the text. I guess it is a desired feature. (try to retain/imitate the native behaviour when possible)

packages/wix-ui-core/src/components/TextLink/TextLink.tsx
```diff
@@ -0,0 +1,68 @@
+import * as React from 'react';
+import * as PropTypes from 'prop-types';
+import style from './TextLink.st.css';
+import {Text} from '../Text/index';
+import {createHOC} from '../../createHOC';
+
+export interface TextLinkProps {
+  className?: string;
+  disabled?: boolean;
+  href?: string;
```
hepiyellow: should not be optional

packages/wix-ui-core/src/components/TextLink/TextLink.tsx
```diff
@@ -0,0 +1,68 @@
...
+  target?: '_blank' | '_self' | '_parent' | '_top';
+  children?: React.ReactNode;
```
hepiyellow: should not be optional

packages/wix-ui-core/src/components/TextLink/TextLink.tsx
```diff
@@ -0,0 +1,68 @@
...
+export interface TextLinkState {
+  focus: boolean;
+}
+
+export class TextLink extends React.Component<TextLinkProps, TextLinkState> {
+  static propTypes = {
+    /** Makes the link disabled */
+    disabled: PropTypes.bool,
+    /** Standard anchor href property */
+    href: PropTypes.string,
```
hepiyellow: href should be required

packages/wix-ui-core/src/components/TextLink/TextLink.tsx
```diff
@@ -0,0 +1,68 @@
...
+    /** Standard anchor target property */
+    target: PropTypes.oneOf(['_blank', '_self', '_parent', '_top']),
+    /** Any node to be rendered (usually text node) */
+    children: PropTypes.any,
```
hepiyellow: children should be required, and of type `React.node`

packages/wix-ui-core/src/components/TextLink/TextLink.tsx
```diff
@@ -0,0 +1,68 @@
+import * as React from 'react';
+import * as PropTypes from 'prop-types';
+import style from './TextLink.st.css';
+import {Text} from '../Text/index';
+import {createHOC} from '../../createHOC';
+
+export interface TextLinkProps {
```
hepiyellow: Please add the documentation comments here as well

packages/wix-ui-core/src/components/TextLink/TextLink.tsx
```diff
@@ -0,0 +1,68 @@
...
+  href?: string;
+  target?: '_blank' | '_self' | '_parent' | '_top';
+  children?: React.ReactNode;
+}
+
+export interface TextLinkState {
+  focus: boolean;
+}
+
+export class TextLink extends React.Component<TextLinkProps, TextLinkState> {
```
hepiyellow: I think you can extend `React.PureComponent` instead of `React.Component`

packages/wix-ui-core/src/components/TextLink/TextLink.tsx
hepiyellow: Also, you should check if `this.props.href` is not undefined.

packages/wix-ui-core/src/components/TextLink/TextLink.tsx
```diff
@@ -0,0 +1,68 @@
...
+  state: TextLinkState = {
+    focus: false,
+  };
+
+  render() {
+    const {focus} = this.state;
+    const {href, disabled, target, children} = this.props;
+
+    return (
+      <a
```
hepiyellow: This might require a little discussion, but I think we should allow any props, and spread them onto the `<a>`  (to allow event handlers: onFocus, onBlur, onClick, onKeyDown, onMouseEnter,...)
I think the way to do it is to omit only the `href` property from the given props, and for the onFocus and onBlur handlers, simply run our logic and call the `this.props.onFocus`,...
This means that you don't need to do anything with `target` (remove it from the prop types).
Use `AnchorHTMLAttributes` as the props type.

packages/wix-ui-core/src/components/TextLink/TextLink.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+export interface TextLinkProps {
+  className?: string;
+  disabled?: boolean;
+  href?: string;
+  target?: '_blank' | '_self' | '_parent' | '_top';
+  children?: React.ReactNode;
+}
+
+export interface TextLinkState {
```
hepiyellow: you don't need to export the state ( I think)

packages/wix-ui-core/src/components/TextLink/index.tsx
```diff
@@ -0,0 +1 @@
+export {TextLink, TextLinkProps} from './TextLink';
```
hepiyellow: you can use `export {*} from ...` if you like it.

packages/wix-ui-core/stories/TextLink/TextLink.story.tsx
```diff
@@ -0,0 +1,15 @@
...
+  category: 'Components',
+  storyName: 'TextLink',
+
+  component: TextLink,
+  componentPath: '../../src/components/TextLink',
+
+  componentProps: () => ({
+    'data-hook': 'storybook-text-link',
+    children: 'I want you to click',
+    href: 'https://wix.com'
```
hepiyellow: better not use `https`, `http` is enough

packages/wix-ui-core/src/components/TextLink/TextLink.spec.tsx
```diff
@@ -0,0 +1,71 @@
...
+describe('TextLink', () => {
+  const createDriver = createDriverFactory(textLinkDriverFactory);
+
+  it('should exist', () => {
+    const textLink = createDriver(<TextLink />);
+    expect(textLink.exists()).toBeTruthy();
+  });
+
+  describe('style states', () => {
+    it('should support disabled state', () => {
```
hepiyellow: You should also test that it is NOT disabled when you don't pass the `disabled` property.

packages/wix-ui-core/src/components/TextLink/TextLink.spec.tsx
```diff
@@ -0,0 +1,71 @@
...
+      const textLink = createDriver(<TextLink disabled />);
+      expect(textLink.hasStyleState('disabled')).toBeTruthy();
+    });
+
+    it('should support focus state', () => {
```
hepiyellow: you should also test that it is NOT focused when initialy rendered

packages/wix-ui-core/src/components/TextLink/TextLink.spec.tsx
```diff
@@ -0,0 +1,71 @@
...
+      const textLink = createDriver(<TextLink disabled />);
+      textLink.focus();
+      expect(textLink.hasStyleState('focus')).toBeTruthy();
```
hepiyellow: Disabled elements should not be focusable. Are you sure this is what you wanted to test?

packages/wix-ui-core/src/components/TextLink/TextLink.tsx
```diff
@@ -0,0 +1,68 @@
...
+  focus: boolean;
+}
+
```
hepiyellow: There is nothing "Text"-ish in this component, it is simple Anchor abstraction with a disabled feature. I agree that we can call it <Link> or even <Anchor> (-:

packages/wix-ui-core/src/components/TextLink/TextLink.tsx
argshook: no, why? anchor can easily be used within page without href and with name. moreover, it could be used with onClick without href

packages/wix-ui-core/src/components/TextLink/TextLink.tsx
argshook: why omit href? just spread all props, don't omit anything. set default props for default behaviour

packages/wix-ui-core/src/components/TextLink/TextLink.tsx
argshook: no `void()` please. if there is no `props.href`, then no `href` attribute should be created. same with disabled case

xmityaz:
Thanks guys for review.
Will do the changes soon

packages/wix-ui-core/src/components/TextLink/TextLink.spec.tsx
```diff
@@ -0,0 +1,71 @@
...
+import {mount} from 'enzyme';
+import {isTestkitExists} from 'wix-ui-test-utils/vanilla';
+import {textLinkTestkitFactory} from '../../testkit';
+import {textLinkTestkitFactory as enzymeTextLinkTestkitFactory} from '../../testkit/enzyme';
+import {isEnzymeTestkitExists} from 'wix-ui-test-utils/enzyme';
+
+describe('TextLink', () => {
+  const createDriver = createDriverFactory(textLinkDriverFactory);
+
+  it('should exist', () => {
```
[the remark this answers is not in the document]
shlomitc: This is what we do everywhere, so I think it's ok to keep it like this for now

packages/wix-ui-core/src/components/TextLink/TextLink.tsx
shlomitc: it shouldn't be used anyway, maybe in `wix-ui-backoffice`, but not very useful here anyway

packages/wix-ui-core/src/components/TextLink/TextLink.spec.tsx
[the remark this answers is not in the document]
hepiyellow: I am hust saying that the implementation of the exists() could be to return always true (a common bug), and the test would still pass. 
So we need a case that calks exists(), and expects false. 

BTW: this exists() can be part of a BaseDriver, which all drivers would extend, so we don’t have to re-implement and re-test for every driver.

packages/wix-ui-core/src/components/TextLink/TextLink.tsx
hepiyellow: Dont the e2e tests in the core need to use a datahook?
We can add the hook only in the story if you dont want it in the exported component.

packages/wix-ui-core/src/components/TextLink/TextLink.tsx
hepiyellow: Ok.

argshook:
fyi for all, as agreed with @xmityaz i'll take this over

shlomitc:
I'm closing this PR. it looks like we don't want to have a wrapper for linking. we can just use any link/anchor and attach styles or create a `Text` component that applies styles for `a` elements

PR #388, opened 2018-04-08
Timepicker refactor

guybenron:
Using new Input component, and moving tickers into the core TimePicker


packages/wix-ui-core/src/components/TimePicker/Tickers.st.css
```diff
@@ -0,0 +1,38 @@
...
+  flex-direction: column;
+  z-index: 1;
+}
+
+:global([dir="rtl"]) .root {
+  right: auto;
+  left: value(tickerInset);
+}
+
+.root:rtl {
```
lbelinsk: why the duplication? do you need rtl state? I see that you don't even add rtl state on the component in the tsx file

packages/wix-ui-core/src/components/TimePicker/Tickers.tsx
```diff
@@ -0,0 +1,45 @@
+import * as React from 'react';
+import {func} from 'prop-types';
+import style from './Tickers.st.css';
+
+const ArrowUp = props => (
```
lbelinsk: you should not use a ad-hoc svg. import the icon from wix-ui-icons-common or add an icon if needed.

lbelinsk:
@alisey

packages/wix-ui-icons-common/src/general/raw/TickerDown.svg
```diff
@@ -0,0 +1,5 @@
+<svg width="10" height="5" viewBox="0 0 10 4" xmlns="http://www.w3.org/2000/svg">
```
alisey: All new icons need to approved with Ben Benhorin ([wuwa](https://github.com/wuwa)) or Domas Markevičius ([milkyfruit](https://github.com/milkyfruit))

wix-ui-icons-common already contains ticker icons:

```js
import FormFieldSpinnerUp from 'wix-ui-icons-common/system/FormFieldSpinnerUp';
import FormFieldSpinnerDown from 'wix-ui-icons-common/system/FormFieldSpinnerDown';
```

packages/wix-ui-core/src/components/TimePicker/TimePicker.tsx
```diff
@@ -363,20 +369,29 @@ export class TimePicker extends React.PureComponent<TimePickerProps, TimePickerS
...
-    const {useNativeInteraction, useAmPm, ...rest} = this.props;
+    const {useNativeInteraction, useAmPm, showTickers, ...rest} = this.props;
     const passThroughProps = omit(rest, Object.keys(TimePicker.propTypes));
 
     let {value} = this.state;
     if (useAmPm) {
       value = convertToAmPm(value);
     }
 
+    const tickers = !showTickers ? null : (
```
alisey: `const tickers = showTickers && (...);`

packages/wix-ui-core/src/components/TimePicker/TimePicker.tsx
```diff
@@ -363,20 +369,29 @@ export class TimePicker extends React.PureComponent<TimePickerProps, TimePickerS
   }
 
   render() {
-    const {useNativeInteraction, useAmPm, ...rest} = this.props;
+    const {useNativeInteraction, useAmPm, showTickers, ...rest} = this.props;
     const passThroughProps = omit(rest, Object.keys(TimePicker.propTypes));
```
alisey: Don't rely on propTypes because we're going to remove them once they're not needed for the storybook. Also seems brittle, e.g. if I add `className` to the Tickers `propTypes` I'm going to break the component.

packages/wix-ui-core/src/components/TimePicker/Tickers.tsx
```diff
@@ -0,0 +1,32 @@
...
+  /** increment handler */
+  onIncrement?: React.MouseEventHandler<HTMLButtonElement>;
+  /** decrement handler */
+  onDecrement?: React.MouseEventHandler<HTMLButtonElement>;
+}
+
+export const Tickers: React.SFC<TickersProps> = (props) => (
+  <div {...style('root', {}, props)}>
+    <button onClick={props.onIncrement} className={style.ticker}>
+      <TickerUp width={10} height={5} fill="none"/>
```
alisey: Core components should not contain any icons: https://github.com/wix/wix-ui/blob/master/docs/icons.md#core-component-guidelines

Icons should be provided by consumers. I know this might sound like a strange decision.

packages/wix-ui-core/src/components/TimePicker/Tickers.st.css
```diff
@@ -0,0 +1,37 @@
...
+:vars {
+  tickerInset: 6px;
+  borderWidth: 0px;
+}
+
+.root {
+  position: absolute;
+  right: value(tickerInset);
+  top: value(borderWidth);
+  height: calc(100% - 2 * value(borderWidth));
```
alisey: why not simply `box-sizing: border-box`?

packages/wix-ui-core/src/components/TimePicker/Tickers.st.css
```diff
@@ -0,0 +1,37 @@
...
+  -st-from: "../../mixins/calc";
+  -st-default: calc
+}
+
+:vars {
+  tickerInset: 6px;
+  borderWidth: 0px;
+}
+
+.root {
```
alisey: Consumers should be able to import a core component and add their own styles without having to override positions, colors, fonts, borders, etc. Which means all core component should come with basically an empty stylesheet that only defines selectors and styling that is absolutely essential for the component's functioning.

Here's an example: https://github.com/wix/wix-ui/blob/master/packages/wix-ui-core/src/components/ToggleSwitch/ToggleSwitch.st.css

There are styles defined for the input because the component cannot work at all if the input doesn't fully cover it.

You can create an extra stylesheet that provides some basic styling, e.g.:

https://github.com/wix/wix-ui/blob/master/packages/wix-ui-core/src/components/ToggleSwitch/ToggleSwitchLayout.st.css

https://github.com/wix/wix-ui/blob/master/packages/wix-ui-core/src/components/ToggleSwitch/ToggleSwitchColor.st.css

But consumers shouldn't be forced to use it.

packages/wix-ui-core/src/components/TimePicker/Tickers.tsx
```diff
@@ -0,0 +1,32 @@
...
+  className?: string;
+  /** increment handler */
+  onIncrement?: React.MouseEventHandler<HTMLButtonElement>;
+  /** decrement handler */
+  onDecrement?: React.MouseEventHandler<HTMLButtonElement>;
+}
+
+export const Tickers: React.SFC<TickersProps> = (props) => (
+  <div {...style('root', {}, props)}>
+    <button onClick={props.onIncrement} className={style.ticker}>
```
argshook: button without type may accidentally submit forms, so it should have `type="button"`

packages/wix-ui-core/src/components/TimePicker/TimePicker.st.css
```diff
@@ -0,0 +1,11 @@
+:import {
+    -st-from: "../Input/Input.st.css";
+    -st-default: Input;
+}
+
+.root {
+    -st-extends: Input;
+    display: inline-block;
```
argshook: input is by default inline-block, guess you can remove this line

packages/wix-ui-core/src/components/TimePicker/TimePicker.tsx
argshook: no, we are not going to remove them as long as we don't have a good approach to convert interfaces to proptypes.

proptypes are not only for documentation, they're also for non TS consumers

packages/wix-ui-core/src/components/TimePicker/TimePicker.tsx
argshook: it would be even better to not have `tickers` inside render and construct them elsewhere. the displayOrNot condition should be decided in render, not in `const tickers = ...`

packages/wix-ui-core/src/components/TimePicker/TimePicker.tsx
```diff
@@ -363,20 +369,29 @@ export class TimePicker extends React.PureComponent<TimePickerProps, TimePickerS
...
+    );
+
     return (
       <Input
-        ref         = {ref => this._inputRef = ref}
         {...style('root', {}, this.props)}
         {...passThroughProps}
+        ref         = {ref => this._inputRef = ref}
         value       = {value}
+        prefix      = {tickers}
```
argshook: not suffix?

packages/wix-ui-core/src/components/TimePicker/Tickers.tsx
```diff
@@ -0,0 +1,32 @@
...
+    <button onClick={props.onDecrement} className={style.ticker}>
+      <TickerDown width={10} height={5} fill="none"/>
+    </button>
+  </div>
+);
+
+Tickers.displayName = 'Tickers';
+
+Tickers.propTypes = {
+  /** increment handler */
```
argshook: missing `className`

packages/wix-ui-core/src/components/TimePicker/Tickers.st.css
```diff
@@ -0,0 +1,37 @@
...
+  position: absolute;
```
argshook: something's fishy, i think i saw tickers are set in `<Input prefix={}>` but then rely on absolute positioning to go to right?

absolute positioning should not be needed and tickers should go to `suffix`. what am i missing?

if it's because of RTL, let's solve it separately within input, that one should take care of how to render prefix and/or suffix in different directions

packages/wix-ui-core/src/components/TimePicker/Tickers.tsx
alisey: Another issue here is that ticker buttons are focusable. In the native time input clicking on the ticker keeps the input focused, but in our component you need to manually change the focus back.

Maybe we could solve this by preventDefault() in the ticker's mousedown handler.

packages/wix-ui-core/src/components/TimePicker/utils.spec.ts
```diff
@@ -0,0 +1,219 @@
...
+
+      it('should return "00:--" for input of "23:--", default step', () => {
+        expect(increment({value: `23:${BLANK}`, field: FIELD.HOUR})).toEqual(`00:${BLANK}`);
+      });
+
+      it('should return "01:00" for input of "00:00", default step', () => {
+        expect(increment({value: '00:00', field: FIELD.HOUR})).toEqual('01:00');
+      });
+
+      it('should return "01:00" for input of "00:00", step = 20', () => {
```
lbelinsk: not sure I understand. step=20, field=hour, and the hour grows +1?
You mean that this test is checking that the hours will always grow by one?
If so, I think it is better to say `should increment by 1 hour for any step value`

packages/wix-ui-core/src/components/TimePicker/utils.spec.ts
```diff
@@ -0,0 +1,219 @@
...
+  describe('isValidTime', () => {
+    describe('24-hour values', () => {
+      it('should return true for 00:00', () => expect(isValidTime('00:00')).toBeTruthy());
+      it('should return true for 23:59', () => expect(isValidTime('23:59')).toBeTruthy());
+      it('should return false for 0000', () => expect(isValidTime('0000')).toBeFalsy());
+      it('should return false for 00;00', () => expect(isValidTime('00;00')).toBeFalsy());
+      it('should return false for 00:00:00', () => expect(isValidTime('00:0:000')).toBeFalsy());
+      it('should return false for z0:00', () => expect(isValidTime('00z00')).toBeFalsy());
+      it('should return false for 24:00', () => expect(isValidTime('24:00')).toBeFalsy());
+    });
```
lbelinsk: Can add also  isValid("--:--")..

WixBuildServer:
View storybook at: https://wix-wix-ui-icons-common-pr-388.surge.sh

WixBuildServer:
View storybook at: https://wix-wix-ui-core-pr-388.surge.sh

packages/wix-ui-core/src/components/TimePicker/TimePicker.tsx
```diff
@@ -16,14 +15,38 @@ import {
...
+  /** Use native (input type = 'time') interaction */
   useNativeInteraction?: boolean;
+
+  /** Display and interact as AM/PM instead of 24 hour */
   useAmPm?: boolean;
+
+  /** The "AM" and "PM" strings to display */
+  amPmStrings?: AmPmStrings;
+
+  /** Interval in minutes to increase / decrease the time when on minutes or external */
```
lbelinsk: I would explain the logic here that on the hours it is only step=1, ...

packages/wix-ui-core/src/components/TimePicker/TimePicker.tsx
```diff
@@ -16,14 +15,38 @@ import {
...
+  onTimeChange?: (value: string) => void;
+
+  /** Use native (input type = 'time') interaction */
   useNativeInteraction?: boolean;
+
+  /** Display and interact as AM/PM instead of 24 hour */
   useAmPm?: boolean;
+
+  /** The "AM" and "PM" strings to display */
+  amPmStrings?: AmPmStrings;
```
lbelinsk: Is this really needed?
Sounds like an unnecessary prop

packages/wix-ui-core/src/components/TimePicker/TimePicker.driver.ts
```diff
@@ -15,19 +14,15 @@ export const timePickerDriverFactory = ({element, eventTrigger}) => {
...
-
-    /** returns elements innerHtml */
-    styles: {
-      /** returns elements display css property */
-      getRootDisplay: () => window.getComputedStyle(element).display,
-      /** returns elements border-radius css property */
-      getBorderRadius: () => window.getComputedStyle(element.querySelector('.outerLabel')).borderRadius
-    }
+    /** returns the ticker group element */
+    getTickers: () => inputDriver.getPrefix(),
```
lbelinsk: should be getSuffix, no?

packages/wix-ui-core/src/components/TimePicker/TimePicker.tsx
argshook: maybe for i18n

packages/wix-ui-core/src/components/TimePicker/TimePicker.tsx
```diff
@@ -56,17 +79,12 @@ export class TimePicker extends React.PureComponent<TimePickerProps, TimePickerS
   /** To keep track of where to increment / decrement externally (ticker) */
   _lastFocusedField: FIELD;
 
-  /** To be able to call focus */
-  _inputRef: Input;
-
   static defaultProps = {
-    onChange             : () => { return null; },
+    onTimeChange         : () => null,
```
argshook: why rename? `onChange` is better imo. is there anything else other than time that should have callback?

packages/wix-ui-core/src/components/TimePicker/TimePicker.tsx
```diff
@@ -100,13 +91,7 @@ export class TimePicker extends React.PureComponent<TimePickerProps, TimePickerS
...
 
     /** Display and interact as AM/PM instead of 24 hour */
-    useAmPm: bool,
-
-    /** The "AM" and "PM" strings to display */
-    amPmStrings: shape({
-      am: string,
-      pm: string
-    }),
+    useAmPm: oneOf([AmPmOptions.None, AmPmOptions.Lowercase, AmPmOptions.Uppercase, AmPmOptions.Capitalized]),
```
lbelinsk: oneOf(Object.keys(AmPmOptions))

lbelinsk:
@argshook renamed `onTimeChange` to `onChange`. will be called each time a valid time was entered

PR #405, opened 2018-04-10
Update build scripts

amiryonatan:


packages/wix-storybook-utils/package.json
```diff
@@ -18,14 +18,13 @@
   "homepage": "https://github.com/wix/wix-ui#readme",
   "main": "./dist/src/index.js",
   "scripts": {
-    "pretest": "npm run build && build-storybook",
-    "test": "yoshi test --jest",
-    "test:watch": "yoshi test --jest --watch",
+    "build": "yoshi build && import-path --path src",
+    "test": "build-storybook && yoshi test --jest",
```
shlomitc: Why not add build-storybook also to the build step?

packages/wix-ui-icons-common/package.json
```diff
@@ -20,7 +20,7 @@
   },
   "scripts": {
     "build": "npm run buildIcons && yoshi build && npm run create-aliases && npm run generate-stylable-index",
-    "test": "npm run build && build-storybook",
+    "test": "build-storybook",
```
shlomitc: Same here

.travis.yml
```diff
@@ -14,7 +14,7 @@ before_install:
  - sudo dpkg -i google-chrome*.deb
  
 script:
- - npm run bootstrap && npm run test
+ - npm run bootstrap && npm run build && npm run test
```
shlomitc: Could be interesting to run the surge script before test. Since it runs anyway even if tests fail. But this requires the storybook build to be done in the test command

PR #407, opened 2018-04-11
upgrade packages to react 16

amiryonatan:


amiryonatan:
@shlomitc Travis passed, TC failed because it only runs wix-ui-core and not the entire suite

PR #414, opened 2018-04-12
Added puppeteer driver to Label

Zombiefruit:


packages/wix-ui-core/src/components/Label/Label.puppeteer.driver.ts
```diff
@@ -0,0 +1,5 @@
+export const labelDriverFactory = (component, page) => ({
+  click: () => component.click(),
+  getLabelText: () => page.eval(_component => _component.innerText, component),
```
shlomitc: @yanivefraim  - can you please also review the puppeteer driver? is there a best-practice for evaluating text?

yanivefraim: @shlomitc @lbelinsk - you should use `page.evaluate`, there's no such thing as `page.eval`.
regarding best practice for getting text, there's no better way, as far as I can tell (see the comment for my PR [here](https://github.com/GoogleChrome/puppeteer/pull/2172#issuecomment-372048563)

WixBuildServer:
View storybook at: https://wix-wix-ui-icons-common-pr-414.surge.sh

PR #416, opened 2018-04-12
Add [TS] Types to all protractor drivers

hepiyellow:
While at it, had to workaround a [protractor bug](https://github.com/angular/protractor/issues/4760) by adding `async` before the problematic functions.

EDIT:
after reading this:
https://github.com/Microsoft/TypeScript/issues/5711

There is another solution (not ideal):
To add:
`import {promise} from 'protractor/node_modules/@types/selenium-webdriver';`

that is, instead of adding `async`.

### DIFF
--- a/packages/wix-ui-core/src/components/Input/Input.protractor.driver.ts
+++ b/packages/wix-ui-core/src/components/Input/Input.protractor.driver.ts
@@ -1,10 +1,15 @@
-export const inputDriverFactory = component => {
+import {ElementFinder} from 'protractor';
+
+export const inputDriverFactory = (component: ElementFinder) => {
   const input = component.$('input');
 
   return {
     element: () => component,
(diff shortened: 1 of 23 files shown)


packages/wix-ui-core/src/components/Input/Input.protractor.driver.ts
```diff
@@ -1,10 +1,15 @@
-export const inputDriverFactory = component => {
+import {ElementFinder} from 'protractor';
+
+export const inputDriverFactory = (component: ElementFinder) => {
   const input = component.$('input');
 
   return {
     element: () => component,
-    enterText: text => input.clear().sendKeys(text),
```
hepiyellow: I am not sure if this actually worked. But I think I've fixed it.

hepiyellow:
Rebasing, and adding explicit interfaces.

PR #417, opened 2018-04-13
feat(Pop): POC of how popover should look like DONT MERGE THIS IS POC

argshook:
this is an example of how i imagine composition done in popover.

quick & dirty, actual implementation may differ from this example. The idea is to use functions for trigger and menu elements. some say it's "render prop". it's just a function.

this approach makes [magic children calculations](https://github.com/wix/wix-ui/blob/wix-ui-core/popover-poc/packages/wix-ui-core/src/utils/index.ts) unnecessary and, what's important, it totally disconnects popover mechanism from rendered components

to expand further, `arrow` could also come as prop to `Popover`. `if prop.arrow then render it otherwise render default arrow `

component api in this example has `show`, `hide` and `toggle`. it could also have `move`, `update` and whatever else we wish

### DIFF
--- a/packages/wix-ui-core/src/baseComponents/Pop/index.tsx
+++ b/packages/wix-ui-core/src/baseComponents/Pop/index.tsx
@@ -1,0 +1,68 @@
+import * as React from 'react';
+import Popper from 'popper.js';
+
+export interface Controls {
+  toggle(): void;
+  show(): void;
+  hide(): void;
+}
+
+export type Placement = Popper.Placement;
+
+export interface Props {
+  trigger(Controls): React.ReactNode;
+  menu(Controls): React.ReactNode;
+  placement?: Placement;
+}
+
+export interface State {
+  open: Boolean;
+}
+
+export class Pop extends React.PureComponent<Props, State> {
+  toggleRef: Element;
+  menuRef: Element;
+  popper: any;
+
+  state = {
+    open: false
+  };
+
+  defaultProps = {
+    placement: 'bottom'
+  };
+
+  componentDidMount() {
+    this.popper = new Popper(
+      this.toggleRef,
+      this.menuRef,
+      {
+        placement: this.props.placement || 'bottom'
+      }
+    );
+  }
+
+  _update(state) {
+    this.setState(state, this.popper.scheduleUpdate);
+  }
+
+  _componentInterface = {
+    toggle: () => this._update({open: !this.state.open}),
+    show: () => this._update({open: true}),
+    hide: () => this._update({open: false})
+  };
+
+  render() {
+    return (
+      <div style={{display: 'inline-block'}}>
+        <div ref={ref => this.toggleRef = ref}>
+          {this.props.trigger(this._componentInterface)}
+        </div>
+
+        <div ref={ref => this.menuRef = ref}>
+          {this.state.open && this.props.menu(this._componentInterface) }
+        </div>
+      </div>
+    );
+  }
+}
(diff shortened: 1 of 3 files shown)


shlomitc:
So let me see if I understand the concept.
1. The toggle receives a trigger element and menu element. both are "render props", meaning the `Pop` component renders them and adds the functionality of showing/hiding the popover.
2. It does not contain any logic related to mouse events, as `Tooltip` can decide to implement it on his own. As this behavior might also be applied to a popover, isn't it better to let the `Pop` component do it on his own?

Why is this approach better than using `react-popper`?

argshook:
1. roughly yes. You use `Popover` component and tell it
* what is the anchor (the `toggle`: a function that returns a button, icon, image, input, video, whatever)
* what is the dropped and positioned thing (the `menu`: a function that can also return any component)

here's how to use it: https://github.com/wix/wix-ui/pull/417/files#diff-4827b16d0e6a374a1790590cfc8d5abaR50

2. no, that's the opposite of what i'm trying to convey. If `Popover` knows about events, or what components it renders, then all the "render props" looses its meaning.

> Why is this approach better than using react-popper?
this is not related at all. `react-popper` is just abstraction layer over `popperjs` (which im using here). I didn't use `react-popper` cause im lazy, it can still be used

hepiyellow:
Thank !
I was about to do something similar, and now it is much clearer.
I am working on a `ClosableHOC` which will wrap the Popover, using you concept.

packages/wix-ui-core/src/baseComponents/Pop/index.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+export interface Controls {
+  toggle(): void;
+  show(): void;
+  hide(): void;
+}
+
+export type Placement = Popper.Placement;
+
+export interface Props {
```
hepiyellow: need to add `initiallyopened: boolean`

argshook: i don't think it's needed, consumer can call `open` on their `componentDidMount` by themselves and achieve same behaviour

hepiyellow: In your solution there would be a  double `render`  which could be avoided.

argshook: i think the number of renders would be the same in both cases (i think, didn't test). but before delving into performance optimization i would first streamline the api and make it as little and as flexible as possible

packages/wix-ui-core/src/baseComponents/Pop/index.tsx
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  _update(state) {
+    this.setState(state, this.popper.scheduleUpdate);
+  }
+
+  _componentInterface = {
+    toggle: () => this._update({open: !this.state.open}),
+    show: () => this._update({open: true}),
+    hide: () => this._update({open: false})
```
hepiyellow: Do you think that there is a use-case for adding a callback to this function, `onHide()`(for `show` also) which will be called AFTER the state has been set.
Meaning (using React's "state has been set" callback):
```
setState({open=false}, onHide)
```

See this use-case in the core's Dropdown:
https://github.com/wix/wix-ui/blob/43759f9ad5ff0220373201b06e0fb8436adc51f5/packages/wix-ui-core/src/baseComponents/Dropdown/Dropdown.tsx#L98

argshook: perhaps instead of api method, `open` and friends should be promises

```
_componentInterface = {
  open: () => new Promise(resolve => this._update({open: !this.state.open}, resolve))
}
// then later
open()
// or
open().then(onOpen)
```

shlomitc:
I'm closing the PR but I will keep the branch alive so we can refer to it later

PR #418, opened 2018-04-15
LabelWithOptions - Add checkbox

danielstro:
- Added the option to show checkboxes next to the options.
- Fixed a problem with checkbox that causes multiple events to be fired


packages/wix-ui-core/src/components/Checkbox/Checkbox.tsx
```diff
@@ -47,6 +47,7 @@ export class Checkbox extends React.Component<CheckboxProps, CheckboxState> {
           <input
             type="checkbox"
             className={styles.nativeCheckbox}
+            onClick={e => e.stopPropagation()}
```
shlomitc: why is this needed?
probably need a test for it

packages/wix-ui-core/src/components/LabelWithOptions/LabelWithOptions.driver.ts
```diff
@@ -1,23 +1,28 @@
...
 
   return Object.assign({}, dropdownDriver, labelDriver, {
     getSuffix: () => suffixElement,
-    isRequired: () => domUtils.hasStyleState(args.element, 'required'),
-    isError: () => domUtils.hasStyleState(args.element, 'error'),
-    isDisabled: () => domUtils.hasStyleState(args.element, 'disabled')
+    isRequired: () => domUtils.hasStyleState(element, 'required'),
+    isError: () => domUtils.hasStyleState(element, 'error'),
+    isDisabled: () => domUtils.hasStyleState(element, 'disabled'),
+    checkboxAt: (index: number) => {
```
shlomitc: driver function name should make it clear that it's a driver that returned

packages/wix-ui-core/src/components/LabelWithOptions/LabelWithOptions.st.css
```diff
@@ -15,12 +25,18 @@
 
 .root {
   -st-extends: Dropdown;
-  -st-states: required, error, disabled;
+  -st-states: required, error, disabled, checkbox;
 }
 
+@custom-selector :--selection .root:checkbox DropdownOption:selected;
```
shlomitc: the name `selection` might be too general

packages/wix-ui-core/src/components/LabelWithOptions/LabelWithOptions.st.css
```diff
@@ -30,4 +46,25 @@
 }
 
 .placeholder {
+}
+
+.optionCotainer {
+  display: flex;
+}
+
+.checkbox {
```
shlomitc: why do much styling is needed?

packages/wix-ui-core/src/components/LabelWithOptions/LabelWithOptions.tsx
```diff
@@ -127,42 +97,84 @@ export class LabelWithOptions extends React.PureComponent<LabelWithOptionsProps,
...
+    if (!this.props.checkbox) {
+      return this.props.options;
+    }
+    return this.props.options.map(option => {
+      let newOption = {id: option.id, isDisabled: option.isDisabled, isSelectable: option.isSelectable, value: option.value, render: null};
+      const checked = this.state.selectedIds.includes(option.id);
+      newOption.render = option.isSelectable ?
+                         value => (
+                          <div className={style.optionCotainer}>
+                            <Checkbox disabled={option.isDisabled} checked={checked} className={style.checkbox}/>{option.render(value)}
```
shlomitc: a disabled checkbox still shows a "hand" mouse, probably should be fixed in the `Checkbox` component

danielstro: This is an issue with the checkbox

packages/wix-ui-core/src/components/Checkbox/Checkbox.tsx
danielstro: This is one of the things that we can't really test because of the onChange problem. Because we don't really use a browser, we can't recreate this.

packages/wix-ui-core/src/components/LabelWithOptions/LabelWithOptions.driver.ts
danielstro: I'm not sure what name to give it. checkboxDriverAt?

packages/wix-ui-core/src/components/LabelWithOptions/LabelWithOptions.st.css
danielstro: I removed it to the bare minimum needed to actually see something in the storybook

packages/wix-ui-core/src/components/Checkbox/Checkbox.tsx
shlomitc: but shouldn't this be the responsibility of the container that holds the checkbox?

danielstro: I'm not sure, in this case when someone is using a checkbox they don't expect to receive two calls for their onChange handler, so they will not know to disable this.

shlomitc: ok

packages/wix-ui-core/src/components/LabelWithOptions/LabelWithOptions.st.css
shlomitc: 👍

packages/wix-ui-core/src/components/LabelWithOptions/LabelWithOptions.tsx
shlomitc: still needed here

PR #420, opened 2018-04-17
Thumbnail

lbelinsk:


packages/wix-ui-core/src/components/Thumbnail/index.ts
```diff
@@ -0,0 +1 @@
+export {Thumbnail, ThumbnailProps} from './Thumbnail';
```
hepiyellow: You can use `export * from ` ... any reason why not to?

packages/wix-ui-core/src/components/Thumbnail/Thumbnail.e2e.ts
```diff
@@ -0,0 +1,17 @@
...
+describe('Thumbnail', () => {
+  const storyUrl = getStoryUrl('Components', 'Thumbnail');
+
+  beforeEach(() => browser.get(storyUrl));
+  eyes.it('should exist', () => {
+    const dataHook = 'storybook-thumbnail';
+    const driver = thumbnailTestkitFactory({dataHook});
+
+    return waitForVisibilityOf(driver.element(), 'Cannot find Thumbnail')
+      .then(() => expect(driver.element().isPresent()).toBe(true));
```
hepiyellow: I don't mind if you add an `exists()` method to the driver. So it will look similar to the enzyme driver.

packages/wix-ui-core/src/components/Thumbnail/Thumbnail.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+export const Thumbnail: React.SFC<ThumbnailProps> = props => {
+  const children = React.Children.only(props.children);
+  const {selected, selectedIcon, onClick} = props;
+
+  return (
+    <div {...style('root', {selected}, props)} onClick={onClick}>
+      {children}
+
+      <div className={style.selectedIcon} data-hook="selectedIcon">
```
hepiyellow: I felt there is a convention of having data hook values as snake-case

packages/wix-ui-core/src/components/Thumbnail/index.ts
lbelinsk: Maybe just in case in the future someone will add something like:
```
export const SOME_CONST = "Erez";
```

Then we might not want to expose it, and it will happen by mistake. I prefer to define explicitly what i expose from the index of the component

packages/wix-ui-core/src/components/Thumbnail/Thumbnail.e2e.ts
lbelinsk: We don't have `exists` method on protractor drivers, because it has the native `isPresent`. Look for yourself on other protractor drivers

packages/wix-ui-core/src/components/Thumbnail/Thumbnail.tsx
lbelinsk: hm, ok, don't mind changing it.

packages/wix-ui-core/src/components/Thumbnail/Thumbnail.tsx
```diff
@@ -0,0 +1,46 @@
...
+  /** Callback when the element is clicked */
+  onClick?: React.EventHandler<React.MouseEvent<HTMLElement>>;
+  /** Icon to display in when thumbnail is selected */
+  selectedIcon?: React.ReactNode;
+  /** Item to sit inside the Thumbnail */
+  children?: React.ReactNode;
+}
+
+export const Thumbnail: React.SFC<ThumbnailProps> = props => {
+  const children = React.Children.only(props.children);
```
hepiyellow: Why do we enforce a single child?
Is it possible to add prop-type validation for single child?

packages/wix-ui-core/src/components/Thumbnail/Thumbnail.tsx
```diff
@@ -0,0 +1,46 @@
+import * as React from 'react';
+import {bool, node, func, any} from 'prop-types';
```
hepiyellow: Unused `any`

packages/wix-ui-core/src/components/Thumbnail/Thumbnail.tsx
```diff
@@ -0,0 +1,46 @@
...
+      <div className={style.selectedIcon} data-hook="selectedIcon">
+        {selectedIcon}
+      </div>
+    </div>
+  );
+};
+
+Thumbnail.displayName = 'Thumbnail';
+
+Thumbnail.propTypes = {
```
shlomitc: do we still need to use propTypes?

packages/wix-ui-core/src/components/Thumbnail/Thumbnail.spec.tsx
```diff
@@ -0,0 +1,70 @@
...
+      expect(driver.isSelected()).toEqual(false);
+    });
+
+    it('should be selected', () => {
+      const driver = createDriver(<Thumbnail selected><div/></Thumbnail>);
+      expect(driver.isSelected()).toEqual(true);
+    });
+  });
+
+  describe('selectedIcon prop', () => {
```
hepiyellow: Add test case that verifies that the passed Icon is rendered (like you did with the content)

PR #424, opened 2018-04-17
fix(wix-ui-core): popover content wrong width.

etaybarzilay89:
popover content's width was calculated according to first relative parent.
now, popover content is calculated according to element's prop: "appendTo".

### DIFF
--- a/packages/wix-ui-core/package.json
+++ b/packages/wix-ui-core/package.json
@@ -72,6 +72,7 @@
     "raf": "^3.4.0",
     "react": "^16.3.1",
     "react-dom": "^16.3.1",
+    "react-portal": "^4.1.5",
     "react-test-renderer": "^16.3.1",
     "stylable": "latest",
     "stylable-integration": "latest",
--- a/packages/wix-ui-core/src/baseComponents/Popover/Popover.driver.ts
+++ b/packages/wix-ui-core/src/baseComponents/Popover/Popover.driver.ts
@@ -1,5 +1,6 @@
+const queryDocumentOrElement = (element, query) => (document && document.querySelector(query)) || (element && element.querySelector(query));
 const getTargetElement = (element: Element | undefined) => element && element.querySelector('[data-hook="popover-element"]');
-const getContentElement = (element: Element | undefined) => element && element.querySelector('[data-hook="popover-content"]');
+const getContentElement = (element: Element | undefined) => queryDocumentOrElement(element, '[data-hook="popover-content"]');
 const getArrowElement = (element: Element | undefined) => element && element.querySelector('[data-hook="popover-arrow"]');
 
 export const popoverDriverFactory = ({element, eventTrigger}) => ({
--- a/packages/wix-ui-core/src/baseComponents/Popover/Popover.spec.tsx
+++ b/packages/wix-ui-core/src/baseComponents/Popover/Popover.spec.tsx
@@ -1,13 +1,13 @@
 import * as React from 'react';
 import {popoverDriverFactory} from './Popover.driver';
 import {createDriverFactory} from 'wix-ui-test-utils/driver-factory';
-import {Popover} from './';
+import {Popover, PopoverProps} from './';
 import {mount} from 'enzyme';
 import * as eventually from 'wix-eventually';
 
 describe('Popover', () => {
   const createDriver = createDriverFactory(popoverDriverFactory);
-  const createPopover = (props = {}) =>
+  const createPopover = (props: Partial<PopoverProps> = {}) =>
     <Popover placement="top" showArrow={true} shown={false} {...props}>
       <Popover.Element>
         <div>
@@ -54,12 +54,13 @@ describe('Popover', () => {
     expect(arrowLeft).toEqual('10px');
   });
 
-  it('should animate by default', () => {
+  it('should animate by default', async () => {
     const wrapper = mount(createPopover({shown: true}));
     const driver = popoverDriverFactory({element: wrapper.children().at(0).getDOMNode(), eventTrigger: null});
     wrapper.setProps({shown: false});
     expect(driver.isContentElementExists()).toBeTruthy();
-    return eventually(() => expect(driver.isContentElementExists()).toBeFalsy());
+    await eventually(() => expect(driver.isContentElementExists()).toBeFalsy());
+    wrapper.unmount();
   });
 
   it('should not animate in case timeout is set to 0', async () => {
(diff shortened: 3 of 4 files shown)


packages/wix-ui-core/src/baseComponents/Dropdown/Dropdown.driver.ts
```diff
@@ -1,5 +1,5 @@
 import {dropdownContentDriverFactory} from '../DropdownContent/DropdownContent.driver';
-import {popoverDriverFactory} from '../Popover/Popover.driver';
+import {popoverDriverFactory} from '../../components/Popover/Popover.driver';
```
hepiyellow: If you are moving Popover to `components` then you might as well move all the component in `baseComponents` to components. The division to components and baseComponents is not even more unclear.

packages/wix-ui-core/src/components/Popover/Popover.tsx
```diff
@@ -0,0 +1,241 @@
...
+          {popper}
+        </Portal> :
+        popper
+    );
+  }
+
+  defineMountingNode(props) {
+    const {appendTo} = props;
+    this.mountingNode = getMountingNode({appendTo, targetRef: this.targetRef});
+    this.stylesObj = style('root', {}, props);
```
hepiyellow: rename to `popperStylesObj`

packages/wix-ui-core/src/components/Popover/Popover.tsx
```diff
@@ -0,0 +1,241 @@
...
+};
+
+/**
+ * Popover
+ */
+export class Popover extends React.Component<PopoverType, PopoverState> {
+  static Element = createComponentThatRendersItsChildren('Popover.Element');
+  static Content = createComponentThatRendersItsChildren('Popover.Content');
+
+  targetRef = null;
```
hepiyellow: Add type

packages/wix-ui-core/src/components/Popover/Popover.tsx
```diff
@@ -0,0 +1,241 @@
...
+  mountingNode: HTMLElement = null;
+  stylesObj = null;
```
hepiyellow: add type

packages/wix-ui-core/src/components/Popover/Popover.tsx
```diff
@@ -0,0 +1,241 @@
...
+    this.stylesObj = style('root', {}, props);
+    this.attachOrDetachStyles(props);
+  }
+
+  componentDidMount() {
+    this.defineMountingNode(this.props);
+    this.setState({isMounted: true});
+  }
+
+  componentWillReceiveProps(nextProps) {
```
hepiyellow: This will be deprecated, use `UNSAFE_componentWillReceiveProps`
https://reactjs.org/docs/react-component.html#unsafe_componentwillreceiveprops

packages/wix-ui-core/src/components/Popover/Popover.tsx
```diff
@@ -0,0 +1,241 @@
...
+    this.defineMountingNode(nextProps);
```
hepiyellow: Need to check if `childrenObject.Element` was changed. If so, then need to `setState({isMounted:false})`

packages/wix-ui-core/src/components/Popover/Popover.tsx
```diff
@@ -0,0 +1,241 @@
...
+    const shouldRenderPopper = isMounted && (shouldAnimate || shown);
+
+    return (
+      <Manager
+        {...style('root manager', {}, this.props)}
+        onClick={onClick}
+        onMouseEnter={onMouseEnter}
+        onMouseLeave={onMouseLeave}>
+        <Target onKeyDown={onKeyDown} data-hook="popover-element" className={style.trigger}>
+          <span ref={r => this.targetRef = r}>
```
hepiyellow: I think this `span` is redundant, and also the `this.targetRef` might be redundant.
The `Manager` provides a context with the targetRef.
See : https://github.com/FezVrasta/react-popper/blob/65a8cbf71b07355510e975e7a27bcc5bf3b2d727/src/Manager.js#L20

packages/wix-ui-core/src/components/Popover/Popover.tsx
```diff
@@ -0,0 +1,241 @@
...
+            {childrenObject.Element}
+          </span>
+        </Target>
+        {shouldRenderPopper && this.renderPopperContent(childrenObject)}
```
hepiyellow: you can pass here `shouldAnimate` (which is calculated later on inside `renderPopperContent`), just to make it more readable.

packages/wix-ui-core/src/components/Popover/Popover.spec.tsx
```diff
@@ -0,0 +1,123 @@
...
+  it('should not animate in case timeout is set to 0', async () => {
+    const wrapper = mount(createPopover({shown: true, timeout: 0}));
+    const driver = popoverDriverFactory({element: wrapper.children().at(0).getDOMNode(), eventTrigger: null});
+    wrapper.setProps({shown: false});
+    expect(driver.isContentElementExists()).toBeFalsy();
+    expect(wrapper.text()).toBe('Element');
+    wrapper.unmount();
+  });
+
+  it('should append popover to body when appendTo is window', () => {
```
hepiyellow: Aggregate appendTo tests in a describe block

packages/wix-ui-core/src/components/Popover/Popover.tsx
```diff
@@ -0,0 +1,241 @@
...
+
+    return this.wrapWithAnimations(popper);
+  }
+
+  attachOrDetachStyles(props) {
+    const {shown, timeout} = props;
+    const shouldAnimate = !!timeout;
+
+    if (this.mountingNode) {
+      if (shouldAnimate) {
```
hepiyellow: change to `if (shouldAnimate | shown)`

packages/wix-ui-core/src/utils/index.ts
```diff
@@ -23,4 +23,22 @@ export const createComponentThatRendersItsChildren = (displayName: string) => {
   return Element;
 };
 
+export const attachStylesToNode = (node: Element, stylesObj) => {
```
hepiyellow: extract this to `stylableUtils.ts`

packages/wix-ui-core/src/utils/utils.spec.tsx
```diff
@@ -70,5 +83,30 @@ describe('Utils', () => {
       const props = {children: {prop: 'value'}};
       expect(component(props)).toEqual(props.children);
     });
-   });
+  });
+
+  describe('node styles attachment', () => {
+    const attributeName = 'data-att';
+    const stylesObj = {
+      className: 'CN',
```
hepiyellow: className may include 2 classes. Right now it fails. So need to add a test with 2 classes.

packages/wix-ui-core/src/utils/index.ts
```diff
@@ -23,4 +23,22 @@ export const createComponentThatRendersItsChildren = (displayName: string) => {
...
+  if (node) {
+    node.classList.add(stylesObj.className);
```
hepiyellow: `className` may include 2 classes (space separated), so this will fail !

packages/wix-ui-core/src/utils/index.ts
```diff
@@ -23,4 +23,22 @@ export const createComponentThatRendersItsChildren = (displayName: string) => {
...
+    Object.keys(stylesObj)
+      .filter(key => key.startsWith('data-'))
+      .forEach(key => node.setAttribute(key, stylesObj[key]));
+  }
+};
+
+export const detachStylesFromNode = (node: Element, stylesObj) => {
+  if (node) {
+    node.classList.remove(stylesObj.className);
```
hepiyellow: need to handle multiple classes seperated by space

packages/wix-ui-core/src/components/Popover/Popover.tsx
shlomitc: These lifecycle are probably not supported in React15.
We can decide to do the calculating in the render function instead.

PR #425, opened 2018-04-17
Added disabled style state to Label

Zombiefruit:


packages/wix-ui-core/src/components/Label/Label.spec.tsx
```diff
@@ -28,6 +28,12 @@ describe('Label', () => {
     expect(label.getForAttribute()).toBe('hey');
   });
 
+  it('takes a disabled prop', async () => {
```
lbelinsk: would add also `should not be disabled by default`

packages/wix-ui-core/src/components/Label/Label.tsx
```diff
@@ -7,13 +7,14 @@ export interface LabelProps {
   children?: string;
   for?: string;
   id?: string;
+  disabled?: boolean;
```
lbelinsk: missing in proptypes

packages/wix-ui-core/src/components/Label/Label.tsx
```diff
@@ -7,13 +7,14 @@ export interface LabelProps {
...
 }
 
 /**
  * Label
  */
 export const Label: React.SFC<LabelProps> = props => {
```
lbelinsk: I think it worth destructuring the props here, you are writing props.* a lot here

PR #426, opened 2018-04-17
add surge for storybook deployment

ronnyrin:


packages/wix-ui-core/package.json
```diff
@@ -30,11 +30,12 @@
     "test:e2e": "yoshi test --protractor",
     "lint": "yoshi lint",
     "start": "yoshi start --no-test & npm run storybook",
-    "release": "yoshi release && gh-pages-auto-release --dist=storybook-static",
+    "release": "yoshi release && gh-pages-auto-release --dist=storybook-static && npx teamcity-surge-autorelease",
```
lbelinsk: Instead of using npx, As far as I know it is a better practice to add `teamcity-surge-autorelease` as a dev-dep and then there is no need for npx. You can check wix-ui-adi

postPRBuild.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
```
lbelinsk: Is this ok that it is only in the root of the mono-repo?

packages/wix-ui-core/package.json
ronnyrin: one of the tasks this week was to change the readme to use `npx`

postPRBuild.sh
ronnyrin: it looks like the ci looks for the script at the root of mono-repo. anyway, it's still doesn't work in mono-repo, I will investigate tomorrow.

packages/wix-ui-core/package.json
lbelinsk: oh, no, it was the opposite :/ update the readme to not use npx...
Maybe it was already updated, sorry for the misunderstanding

ronnyrin: np, i will change

postPRBuild.sh
shlomitc: I think that we need to have a `postPRBuild.sh` file in each package and have a general one in the root.
The one in the root will call each one of the packages.

packages/wix-ui-icons-common/package.json
```diff
@@ -42,6 +42,7 @@
     "react": "^16.3.1",
     "react-dom": "^16.3.1",
     "svg2react-icon": "^2.0.0",
+    "teamcity-surge-autorelease": "latest",
```
shlomitc: prefer a semver version and not latest please

packages/wix-ui-core/package.json
```diff
@@ -73,6 +74,7 @@
     "react": "^16.3.1",
     "react-dom": "^16.3.1",
     "react-test-renderer": "^16.3.1",
+    "teamcity-surge-autorelease": "latest",
```
shlomitc: prefer a semver version and not latest please

packages/wix-ui-core/package.json
```diff
@@ -30,11 +30,12 @@
     "test:e2e": "yoshi test --protractor",
     "lint": "yoshi lint",
     "start": "yoshi start --no-test & npm run storybook",
-    "release": "yoshi release && gh-pages-auto-release --dist=storybook-static",
+    "release": "yoshi release && gh-pages-auto-release --dist=storybook-static && teamcity-surge-autorelease",
```
lbelinsk: if it would all work, also for master, we can remove `gh-pages-auto-release`. But we must be sure it is working, otherwise, we won't have storybook.

If we would change it and it would all work, we mustn't forget to update the documentation with this link

WixBuildServer:
View storybook at: https://wix-wix-ui-pr-426.surge.sh

WixBuildServer:
View storybook at: https://wix-wix-ui-core-pr-426.surge.sh

PR #428, opened 2018-04-18
LabelWithOptions - Add ellipsis

danielstro:
When there are too many selected option, the label overflows. I added a prop that allows ellipsis to be used instead.

### DIFF
--- a/packages/wix-ui-core/src/components/Label/Label.st.css
+++ b/packages/wix-ui-core/src/components/Label/Label.st.css
@@ -1,6 +1,6 @@
 :import {
     -st-from: '../../mixins/mixins.st.css';
-    -st-named: ellipsis;
+    -st-named: ellipsisMixin;
 }
 
 
@@ -9,5 +9,5 @@
 }
 
 .root:ellipsis {
-  -st-mixin: ellipsis;
+  -st-mixin: ellipsisMixin;
 }
--- a/packages/wix-ui-core/src/components/Label/Label.st.css
+++ b/packages/wix-ui-core/src/components/Label/Label.st.css
@@ -1,10 +1,13 @@
+:import {
+    -st-from: '../../mixins/mixins.st.css';
+    -st-named: ellipsis;
+}
+
+
 .root {
     -st-states: ellipsis;
 }
 
 .root:ellipsis {
-  display: block;
-  text-overflow: ellipsis;
-  overflow: hidden;
-  white-space: nowrap;
+  -st-mixin: ellipsis;
 }
--- a/packages/wix-ui-core/src/components/Label/Label.st.css
+++ b/packages/wix-ui-core/src/components/Label/Label.st.css
@@ -1,2 +1,10 @@
 .root {
+    -st-states: ellipsis;
+}
+
+.root:ellipsis {
+  display: block;
+  text-overflow: ellipsis;
+  overflow: hidden;
+  white-space: nowrap;
 }
--- a/packages/wix-ui-core/src/components/Label/Label.st.css
+++ b/packages/wix-ui-core/src/components/Label/Label.st.css
@@ -5,7 +5,7 @@
 
 
 .root {
-    -st-states: ellipsis;
+    -st-states: ellipsis, disabled;
 }
 
 .root:ellipsis {
--- a/packages/wix-ui-core/src/components/Label/Label.tsx
+++ b/packages/wix-ui-core/src/components/Label/Label.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import {string} from 'prop-types';
+import {string, bool} from 'prop-types';
 import style from './Label.st.css';
 
 export interface LabelProps {
@@ -10,11 +10,16 @@ export interface LabelProps {
   ellipsis?: boolean;
 }
 
+const defaultProps: LabelProps = {
+  ellipsis: false
+};
+
 /**
  * Label
  */
 export const Label: React.SFC<LabelProps> = props => {
-  return <label {...style('root', {ellipsis: props.ellipsis}, props)} htmlFor={props.for} id={props.id}>{props.children}</label>;
+  const {id, children, ellipsis} = props;
+  return <label {...style('root', {ellipsis}, props)} htmlFor={props.for} id={id}>{children}</label>;
 };
 
 Label.propTypes = {
@@ -25,5 +30,9 @@ Label.propTypes = {
   /** For property */
   for: string,
   /** ID of element */
-  id: string
+  id: string,
+  /** should the text be ellipsed or not */
+  ellipsis: bool
 };
+
+Label.defaultProps = defaultProps;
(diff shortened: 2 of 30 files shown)


packages/wix-ui-core/src/components/LabelWithOptions/LabelWithOptions.st.css
```diff
@@ -15,7 +15,7 @@
 
 .root {
   -st-extends: Dropdown;
-  -st-states: required, error, disabled;
+  -st-states: required, error, disabled, ellipsis;
```
danielstro: There is a problem with the story, there is no way to change the width of the root element and changing here seems like a mistake. This means that it is not possible to see the ellipsis in the story if the width isn't limited somehow.

danielstro: I removed the ellipsis from the story

packages/wix-ui-core/src/components/LabelWithOptions/LabelWithOptions.st.css
```diff
@@ -25,6 +25,13 @@
   -st-extends: Label;
 }
 
+.root:ellipsis .label {
```
lbelinsk: Maybe we can use some shared ellipsis mixin like we have here:
https://github.com/wix/wix-ui/blob/2aa8281b2b7e7c9a1012387ba428da6ae88f6fb5/packages/wix-ui-core/src/components/StylableText/Text.st.css#L7

It is duplicated

packages/wix-ui-core/src/components/LabelWithOptions/LabelWithOptions.st.css
lbelinsk: 1. Please open an issue for that. This is a feature request for autodocs.
2. You can pass a custom example to the autoExample. I forgot the name of it @argshook

lbelinsk: I think that the ellipsis is more of a `Label` state, not the `LabelWithOptions`, no?

It affects the Label style, not the entire composite.

packages/wix-ui-core/src/components/Label/Label.st.css
```diff
@@ -1,2 +1,10 @@
 .root {
+    -st-states: ellipsis;
+}
+
+.root:ellipsis {
```
lbelinsk: Still, this part is duplcated. we can do a common mixin for it

packages/wix-ui-core/src/components/Label/Label.tsx
```diff
@@ -7,13 +7,14 @@ export interface LabelProps {
...
 
 /**
  * Label
  */
 export const Label: React.SFC<LabelProps> = props => {
-  return <label {...style('root', {}, props)} htmlFor={props.for} id={props.id}>{props.children}</label>;
+  return <label {...style('root', {ellipsis: props.ellipsis}, props)} htmlFor={props.for} id={props.id}>{props.children}</label>;
 };
 
 Label.propTypes = {
```
lbelinsk: add ellipsis to proptypes as well

packages/wix-ui-core/src/components/Label/Label.tsx
```diff
@@ -7,13 +7,14 @@ export interface LabelProps {
   children?: string;
   for?: string;
   id?: string;
+  ellipsis?: boolean;
 }
 
 /**
  * Label
  */
 export const Label: React.SFC<LabelProps> = props => {
```
lbelinsk: Worth adding destructuring here for the props, we use a lot of `props.*`

packages/wix-ui-core/src/components/Label/Label.tsx
```diff
@@ -7,13 +7,14 @@ export interface LabelProps {
   children?: string;
   for?: string;
   id?: string;
+  ellipsis?: boolean;
 }
```
lbelinsk: worth adding defaultProps ellipsis false maybe

packages/wix-ui-core/src/components/Label/Label.tsx
```diff
@@ -7,13 +7,14 @@ export interface LabelProps {
...
 
 /**
  * Label
  */
 export const Label: React.SFC<LabelProps> = props => {
-  return <label {...style('root', {}, props)} htmlFor={props.for} id={props.id}>{props.children}</label>;
+  return <label {...style('root', {ellipsis: props.ellipsis}, props)} htmlFor={props.for} id={props.id}>{props.children}</label>;
```
lbelinsk: Add tests:

1. should not have ellipsis by default
2. should have ellipsis (and pass prop)

packages/wix-ui-core/src/components/LabelWithOptions/LabelWithOptions.tsx
```diff
@@ -157,6 +162,7 @@ export class LabelWithOptions extends React.PureComponent<LabelWithOptionsProps,
         <div className={style.selection}>
           <Label
             className={`${style.label} ${selectedOptions.length ? '' : style.placeholder}`.trim()}
+            ellipsis={ellipsis}
```
lbelinsk: missing test - make sure the Label gets this prop

packages/wix-ui-core/src/components/Label/Label.st.css
danielstro: Added a mixin for this and stylable text

packages/wix-ui-core/src/components/LabelWithOptions/LabelWithOptions.tsx
danielstro: done

packages/wix-ui-core/src/components/Label/Label.tsx
danielstro: done

packages/wix-ui-core/src/components/Label/Label.st.css
```diff
@@ -1,2 +1,13 @@
+:import {
+    -st-from: '../../mixins/mixins.st.css';
+    -st-named: ellipsis;
```
lbelinsk: Since you have state `ellipsis`, I would import it under a different name.

packages/wix-ui-core/src/components/Label/Label.tsx
```diff
@@ -1,19 +1,25 @@
...
+  ellipsis: false
+};
+
 /**
  * Label
  */
 export const Label: React.SFC<LabelProps> = props => {
-  return <label {...style('root', {}, props)} htmlFor={props.for} id={props.id}>{props.children}</label>;
+  const {id, children, ellipsis} = props;
+  return <label {...style('root', {ellipsis}, props)} htmlFor={props.for} id={id}>{children}</label>;
```
lbelinsk: Instead of `props.for`, use the destructureing for it as well

lbelinsk:
@danielstro there are conflicts here that needs to be resolved...

packages/wix-ui-core/src/components/Label/Label.tsx
danielstro: "for" is a reserved word so I can't destruct it

PR #430, opened 2018-04-22
Add optional dataHook prop name

hepiyellow:
Motivation is to allow using the teskit method `isTestkitExists()` without rendering the Element with a `dataHook` prop, which causes a Warning like:
```
React does not recognize the `dataHook` prop on a DOM element. If you intentionally want it
 to appear in the DOM as a custom attribute, spell it as lowercase `datahook` instead. If you accidentally
 passed it from a parent component, remove it from the DOM element.
          in button (created by Button)
          in Button
          in div
```

This camelCase dataHook is needed only in WSR.

### DIFF
--- a/packages/wix-ui-test-utils/src/enzyme/enzyme.ts
+++ b/packages/wix-ui-test-utils/src/enzyme/enzyme.ts
@@ -21,14 +21,26 @@ export function enzymeTestkitFactoryCreator<T extends BaseDriver> (driverFactory
   };
 }
 
+export interface Options {
+  withoutDataHook?: boolean;
+  /** The dataHookPropName exists in order to support legacy CamelCase `dataHook`
+   *  which is used in Wix-Style-React, while the current prop name used in
+   * `wix-ui-core` is snake-case `data-hook`.
+   * */
+  dataHookPropName?: 'data-hook' | 'dataHook';
+}
+
+/**
+ * Checks if the given Element accepts a data hook, and that the testkit factory finds the component's root element using that data hook.
+ *
+ * This method supports both new snake-case and legacy camelCase data hook prop name (e.g `data-hook` and `dataHook`).
+ * The default is to check by both prop name options.
+ */
 export function isEnzymeTestkitExists<T extends BaseDriver> (
   Element: React.ReactElement<any>,
   testkitFactory: (obj: WrapperData) => T,
   mount: MountFunctionType,
-  options?: {
-    withoutDataHook?: boolean,
-    dataHookPropName?: string
-  } ) {
+  options?: Options ) {
 
   const withoutDataHook = (options && options.withoutDataHook) || false;
   const dataHookPropName = (options && options.dataHookPropName);
--- a/packages/wix-ui-test-utils/src/enzyme/enzyme.ts
+++ b/packages/wix-ui-test-utils/src/enzyme/enzyme.ts
@@ -21,9 +21,20 @@ export function enzymeTestkitFactoryCreator<T extends BaseDriver> (driverFactory
   };
 }
 
-export function isEnzymeTestkitExists<T extends BaseDriver> (Element: React.ReactElement<any>, testkitFactory: (obj: WrapperData) => T, mount: MountFunctionType, options = {withoutDataHook: false}) {
-  const dataHook = options.withoutDataHook ? '' : 'myDataHook';
-  const elementToRender = React.cloneElement(Element, {dataHook, 'data-hook': dataHook});
+export function isEnzymeTestkitExists<T extends BaseDriver> (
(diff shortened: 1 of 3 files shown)


packages/wix-ui-test-utils/src/enzyme/enzyme.ts
```diff
@@ -21,9 +21,20 @@ export function enzymeTestkitFactoryCreator<T extends BaseDriver> (driverFactory
...
+export function isEnzymeTestkitExists<T extends BaseDriver> (
+  Element: React.ReactElement<any>,
+  testkitFactory: (obj: WrapperData) => T,
+  mount: MountFunctionType,
+  options?: {
+    withoutDataHook?: boolean,
+    dataHookPropName?: string
+  } ) {
+
+  const withoutDataHook = (options && options.withoutDataHook) || false;
```
lbelinsk: why don't you use default value syntax like we had?
Also for dataHookPropName?

packages/wix-ui-test-utils/test/testkit-helpers.spec.tsx
```diff
@@ -16,7 +16,19 @@ describe('isTestkitExists', () => {
     expect(isTestkitExists(<MyComp/>, testkitFactoryCreator(driver))).toEqual(true);
   });
 
+  it('vanilla should exist using data-hook only', () => {
+    expect(isTestkitExists(<MyComp/>, testkitFactoryCreator(driver), {dataHookPropName: 'data-hook'})).toEqual(true);
+  });
+
   it('enzyme should exist', () => {
```
lbelinsk: Missing `vanilla should exist without data-hook value`

packages/wix-ui-test-utils/src/enzyme/enzyme.ts
```diff
@@ -21,9 +21,20 @@ export function enzymeTestkitFactoryCreator<T extends BaseDriver> (driverFactory
...
-export function isEnzymeTestkitExists<T extends BaseDriver> (Element: React.ReactElement<any>, testkitFactory: (obj: WrapperData) => T, mount: MountFunctionType, options = {withoutDataHook: false}) {
-  const dataHook = options.withoutDataHook ? '' : 'myDataHook';
-  const elementToRender = React.cloneElement(Element, {dataHook, 'data-hook': dataHook});
+export function isEnzymeTestkitExists<T extends BaseDriver> (
+  Element: React.ReactElement<any>,
+  testkitFactory: (obj: WrapperData) => T,
+  mount: MountFunctionType,
+  options?: {
+    withoutDataHook?: boolean,
+    dataHookPropName?: string
```
lbelinsk: The type should be:

dataHookPropName?: 'dataHook' | 'data-hook';

packages/wix-ui-test-utils/src/enzyme/enzyme.ts
```diff
@@ -21,9 +21,20 @@ export function enzymeTestkitFactoryCreator<T extends BaseDriver> (driverFactory
...
 }
 
-export function isEnzymeTestkitExists<T extends BaseDriver> (Element: React.ReactElement<any>, testkitFactory: (obj: WrapperData) => T, mount: MountFunctionType, options = {withoutDataHook: false}) {
-  const dataHook = options.withoutDataHook ? '' : 'myDataHook';
-  const elementToRender = React.cloneElement(Element, {dataHook, 'data-hook': dataHook});
+export function isEnzymeTestkitExists<T extends BaseDriver> (
+  Element: React.ReactElement<any>,
+  testkitFactory: (obj: WrapperData) => T,
+  mount: MountFunctionType,
+  options?: {
```
lbelinsk: I think that here is a good example of something we need to document.
Explain about why we need this, why we have 2 dataHooks..
Prefered in the readme, not as a comment

hepiyellow: done

packages/wix-ui-test-utils/src/enzyme/enzyme.ts
hepiyellow: done

packages/wix-ui-test-utils/src/enzyme/enzyme.ts
hepiyellow: because the `options` object may be optional as well.
I changed this. I am not sure you would like it, but its internal (and tested) , so I am going to merge this.

packages/wix-ui-test-utils/test/testkit-helpers.spec.tsx
hepiyellow: vanilla doesn't have this option, so I guess it is not so important.

PR #434, opened 2018-04-22
wix-ui-core: `<Button>` - Add e2e test and protractor driver.

hepiyellow:


packages/wix-ui-core/src/components/Button/Button.protractor.driver.ts
```diff
@@ -0,0 +1,6 @@
+export const buttonDriverFactory = component => ({
+  /** returns the component element */
+  element: () => component,
+  /** returns the Button's text content */
+  getTextContent: () => component.getText(),
```
lbelinsk: I would rename it to `getButtonTextContent` https://github.com/wix/wix-style-react/blob/master/src/Backoffice/Button/Button.protractor.driver.js#L13 for backward compatability

lbelinsk: Also, on the way you can add click method. the rest of the methods we can deprecate in wsr

PR #435, opened 2018-04-23
wix-ui-core: Add typescript types for all protractor drivers

hepiyellow:


packages/wix-ui-core/src/components/Autocomplete/Autocomplete.e2e.ts
```diff
@@ -4,7 +4,7 @@ import {autocompleteTestkitFactory} from '../../testkit/protractor';
 import {browser} from 'protractor';
 import * as eventually from 'wix-eventually';
 
-describe('Autocomplete', () => {
+fdescribe('Autocomplete', () => {
```
argshook: fdescribe

hepiyellow: good catch!

PR #445, opened 2018-04-25
AddressInput - Adding address icon and dividers

giladartzi:


packages/wix-ui-core/src/components/AddressInput/AddressIcon.tsx
```diff
@@ -0,0 +1,7 @@
+import * as React from 'react';
```
lbelinsk: We have in this mono-repo `wix-ui-icons-common` so it should sit there with all of the icons. Please make sure Ben Benhorin or Domas are ok with this icon.

packages/wix-ui-core/src/components/AddressInput/AddressInput.tsx
```diff
@@ -2,6 +2,7 @@ import * as React from 'react';
 import style from './AddressInput.st.css';
 import {func, string, array, number, bool, oneOf, arrayOf, Requireable} from 'prop-types';
 import {InputWithOptions} from '../../baseComponents/InputWithOptions/InputWithOptions';
+import {AddressIcon} from './AddressIcon'
```
lbelinsk: The new approach we learned to be better is not to couple the core component with some icon. See how we refactored toggleswitch to receive `checkedIcon` and `unChecekdIcon` props.

packages/wix-ui-core/src/components/AddressInput/AddressInput.tsx
```diff
@@ -62,17 +63,25 @@ export interface AddressInputProps {
     forceOptions?: Array<Option>;
     /** Options to override default throttle value, 0 used to disable throttle */
     throttleInterval?: number;
+    /** If set to `true`, location icon will be displayed next to each suggested address */
+    showLocationIcon?: boolean;
```
lbelinsk: See my comment above. Better to replace this prop with `locationIcon` of type ReactElement

lbelinsk: We also need to add tests for this code

giladartzi:
@lbelinsk updated as requested

packages/wix-ui-core/src/components/AddressInput/AddressInput.tsx
```diff
@@ -190,10 +198,10 @@ export class AddressInput extends React.PureComponent<AddressInputProps, Address
         const requestId = ++this.addressRequestId;
         let resolveCurrentAddressRequest = () => null;
         this.currentAddressRequest = new Promise(resolve => resolveCurrentAddressRequest = resolve);
-        const {apiKey, lang, filterTypes} = this.props;
+        const {apiKey, lang, filterTypes, locationIcon} = this.props;
         const results = await this.client.autocomplete(apiKey, lang, createAutocompleteRequest(input, this.props));
         const filteredResults = filterAddressesByType(results, filterTypes);
-        const options = map(filteredResults, createOptionFromAddress);
+        const options = map(filteredResults, address => createOptionFromAddress(address, locationIcon));
```
lbelinsk: The function createOptionFromAddress is a member, so has access to prop.locationIcon, so no need to pass it to it.
then the line can ramain:
```const options = map(filteredResults, createOptionFromAddress);```
 (your call)

File: packages/wix-ui-core/src/components/AddressInput/AddressInput.tsx [hunk 5]
```diff
@@ -153,7 +159,9 @@ export class AddressInput extends React.PureComponent<AddressInputProps, Address
         /** Options to override default one, used for preview mode */
         forceOptions: array,
         /** Options to override default throttle value (ms), 0 used to disable throttle. Default value is 150 */
-        throttleInterval: number
+        throttleInterval: number,
+        /** If set to `true`, location icon will be displayed next to each suggested address */
```
lbelinsk: Update the doc here

packages/wix-ui-core/src/components/AddressInput/AddressInput.tsx
```diff
@@ -62,17 +62,23 @@ export interface AddressInputProps {
...
     inputValue: string;
 }
 
-function createOptionFromAddress(address) {
+function createOptionFromAddress(address, locationIcon) {
     return OptionFactory.create({
         id: address.place_id,
-        value: address.description
+        value: address.description,
+        render: val => <div className={style.addressOption}>
```
lbelinsk: Not important, but maybe it will be nicer to create a function `renderOPtionWithIcon` or something, and then use it here. will let you decide if you want.

packages/wix-ui-core/src/components/AddressInput/AddressInput.st.css
```diff
@@ -6,3 +6,17 @@
 .root {
     -st-extends: InputWithOptions;
 }
+
+.addressOption {
```
lbelinsk: since each class here is an API for the extending themes, I don't think having prefix `address` to each class name.

packages/wix-ui-core/src/components/AddressInput/AddressInput.spec.tsx
```diff
@@ -87,6 +87,16 @@ describe('AddressInput', () => {
         expect(helper.getOptionsText(driver)).toEqual([helper.ADDRESS_DESC_1, helper.ADDRESS_DESC_2]);
     });
 
+    it('Should render location icon if provided', async () => {
```
lbelinsk: I would also add test to check that there is no icon by default (when not provided)

giladartzi:
@lbelinsk fixed according to review

PR #447, opened 2018-04-26
fix documentation of Tooltip.appendTo/appendToParent

jakutis:


shlomitc:
removed appendToParent from the API, so we can close it. thanks!

PR #450, opened 2018-04-29
[WIP] - Upgrade to yoshi_2 (and storybook 4)

hepiyellow:
Still getting an error with `stylable-webpack-plugin`.

Looking into it, we discovered that in this line the `chunk.entryModule` sometimes is undefined.
In my example this line runs 2 times, once with a chunk called ‘preview’ in which entryModule is defined. And a second time with a chunk called `vendor~preview` in which the entryModule is NOT defined.

We tried Not to `addStylableModuleDependenty` when the entryModule is undefined, and our `build_storybook` passed (not sure yet if the results are proper).


hepiyellow:
Waiting for resolution of this issue:
https://github.com/wix-playground/stylable-webpack-plugin/issues/12

PR #451, opened 2018-04-29
Slider

InterAl:


packages/wix-ui-core/src/components/Slider/Slider.st.css
```diff
@@ -61,6 +61,7 @@
 }
 
 .inner {
+  z-index: 1;
```
lbelinsk: why?

InterAl: Should be on top of the tick marks.

PR #466, opened 2018-05-03
<Button> [TS-only] -  Add ButtonDriver type

hepiyellow:


packages/wix-ui-core/src/components/Button/Button.driver.ts
```diff
@@ -1,27 +1,42 @@
-export const buttonDriverFactory = ({element, eventTrigger}) => {
+import { BaseDriver, ComponentFactory, DriverFactory } from 'wix-ui-test-utils/driver-factory';
+
+export interface ButtonDriver extends BaseDriver {
+  /** click on the button root element */
```
lbelinsk: The autodoc isn't looking for the comments in the interface I guess.... only on the actual object

hepiyellow: You guess wrong (-: (I verified it)
We should start moving all docs to interfaces instead of actual objects.

PR #467, opened 2018-05-04
Slider

InterAl:


packages/wix-ui-core/src/components/Slider/Slider.tsx
```diff
@@ -1,3 +1,4 @@
+import {omit} from 'lodash';
```
lbelinsk: if you import like this you will end up with all of lodash in the bundle since tree shaking isn't really working without webpack 4. you need to import from 'lodash/omit';

hepiyellow:
Import omit = require(‘lodash/omit’)

PR #470, opened 2018-05-06
AddressInput - Fixing text alignment

giladartzi:


WixBuildServer:
View storybook at: https://wix-wix-ui-icons-common-pr-470.surge.sh

WixBuildServer:
View storybook at: https://wix-wix-ui-core-pr-470.surge.sh

packages/wix-ui-core/src/components/AddressInput/AddressInput.st.css
```diff
@@ -19,4 +19,8 @@
 
 .icon {
 
-}
\ No newline at end of file
+}
+
+.optionContent {
+    flex: 1;
```
lbelinsk: Not sure why the core needs to decide this. It doesn't feel like a mistake, but I am not fully convinced as well.

giladartzi: Basically, it causes the option to take the full width of its container. I think it's the expected behaviour for the core component

PR #472, opened 2018-05-06
Nadav/readmes

nadavwix:


nadavwix:
work on contribution guide

docs/tools/state-override-hoc.md
```diff
@@ -0,0 +1,21 @@
+# State override HOC
+
+this useful utility wraps a component allowing its state to be overriden, this is usefull for visual tests and editing interaction.
```
alisey: What do you mean by editing interaction?

docs/tools/auto-ssr.md
```diff
@@ -0,0 +1,5 @@
+# Auto SSR
+
+auto SSR is a test-suite for sanity testing component server side rendering
```
alisey: What do we compare the rendered string to? The previous valid value like in Eyes tests?

docs/contributing-components.md
```diff
@@ -0,0 +1,100 @@
...
+npm install
+npm run rebuild
+code .
+```
+
+Example UX Block:
+
+Files:
+
+ - component-name/component-name.tsx - the code
```
shlomitc: we use PascalCase for the component name

docs/contributing-components.md
```diff
@@ -0,0 +1,100 @@
...
+import {styles} from './component-name.st.css';
+
+export interface BlockProps {
+  label:?HTMLLabelProps;
+  dropDown:DropDownProps;
+  className:string;
+}
+
+export const ComponentName = (props:BlockProps)=><div styles('root',{},props)>
+  {props.label ? <label {...props.label} {...styles('label')}/> : null}
```
shlomitc: we use `styles.label` instead of calling the the `styles()` function

docs/contributing-components.md
```diff
@@ -0,0 +1,100 @@
...
+import {HTMLLabelProps} from 'react';
+import {styles} from './component-name.st.css';
+
+export interface BlockProps {
+  label:?HTMLLabelProps;
+  dropDown:DropDownProps;
+  className:string;
+}
+
+export const ComponentName = (props:BlockProps)=><div styles('root',{},props)>
```
shlomitc:
```ts
export const ComponentName = (props: BlockProps) => (
  <div {...styles('root', {}, props)}>
  ...
  </div>
);
```

docs/contributing-components.md
```diff
@@ -0,0 +1,100 @@
...
+
+## How do i decide if to write a UX block or component
+
+it is much easier to write UX blocks, so if you can, please do.
+
+in a UX block you can:
+- render a structure of html and components
+- hide and show nodes
+
+### Example UX Block
```
shlomitc: It is already written below, why do you need it here again?

WixBuildServer:
View storybook at: https://wix-wix-ui-icons-common-pr-472.surge.sh

README.md
```diff
@@ -1,20 +1,21 @@
...
-npm run rebuild
-code .
-```
+## Tools
+* AutoSsr tester - sanity tests component server side rendering using the different simulations provided in its meta file
+* AutoEyes tester - checks component visual using the different simulations (including state simulations)
+* StateOveride mutator - makes component state overridable (usefull for tests)
+
+## Themes
+Read what are themes [here](./docs/WHAT_ARE_THEMES.md)
```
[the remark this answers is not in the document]
nadavwix: cool thanks

docs/contributing-components.md
nadavwix: same same, both are ok

docs/contributing-components.md
nadavwix: thats problematic with git on windows

docs/tools/state-override-hoc.md
nadavwix: when styling a component in the editor you want to preview its different states.
for instance a cart with and without items. a drop down open etc..

docs/contributing-components.md
shlomitc: what is the problem?

docs/tools/state-override-hoc.md
shlomitc: Is this something useful not just for editing mode?
if not - it shouldn't sit in the core library, but in the editor's library

JoMarton: @alisey this probably refers to the case where you want to visually edit the component when in specific state. for example when dropdown is open or component is in loading state or stuff like that

docs/contributing-components.md
```diff
@@ -0,0 +1,100 @@
...
+
+```jsx
+import {DropDown, DropDownProps} from '../drop-down'
+import {HTMLLabelProps} from 'react';
+import {styles} from './component-name.st.css';
+
+export interface BlockProps {
+  label:?HTMLLabelProps;
+  dropDown:DropDownProps;
```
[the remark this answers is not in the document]
nadavwix: true

README.md
```diff
@@ -1,20 +1,21 @@
...
-## Getting started
-```
-git clone git@github.com:wix/wix-ui.git
-cd wix-ui
-npm install
-npm run rebuild
-code .
-```
+## Tools
+* AutoSsr tester - sanity tests component server side rendering using the different simulations provided in its meta file
```
hepiyellow: Do you mean 'variation'? (instead of 'simulations')

docs/contributing-components.md
```diff
@@ -0,0 +1,100 @@
...
+
+## Actual components:
+Components with logic. (and in many cases state)
+
+## How do i decide if to write a UX block or component
+
+it is much easier to write UX blocks, so if you can, please do.
+
+in a UX block you can:
+- render a structure of html and components
```
hepiyellow: Suggestion: "render a structure of native HTML nodes and Custom component nodes"

docs/tools/state-override-hoc.md
idoros: it should allow complex simulations in a tool like storybook (for dev or demo site) as well as the editing case

docs/contributing-components.md
idoros: Windows is case-sensitive, so everything to do with case sensitivity is problematic, renaming files in git or consuming files through a web server. 

In order to avoid confusion, writing all source files as `snake-case` solves those issues and simplify the need to know different conventions for different types of files. 

Note that almost any IDE today provides fuzzy search, so file lookup is not an issue.

README.md
JoMarton: I think both words mean the same thing - a set of props (and in some cases also state overrides) to give the component and run tests over

docs/tools/state-override-hoc.md
hepiyellow: A note about composite components:
I just wanted to say that a StateOverride HOC would not be useful in cases where there is a built-in stateful nested component. e.g:
a <Looader> component which has a built-in <Tooltip> (when in error mode). If Tooltip is rendered by the Loaderer (and not passed in as a prop. Meaning that you for example accept only an errorMessage string as a prop of the loader), then you can not wrap the Tooltip in the HOC. So you can not automatically simulate an opened tooltip.

JoMarton: @hepiyellow a very good point. I renamed the file to `.wip.md` and title includes `(WIP)` to signal this is theoretical at the moment and is used as a base for discussion

PR #478, opened 2018-05-08
Remove Highlighter component

alisey:
I've checked that `Highlighter` is not used directly in any of the projects using `wix-ui-core`


WixBuildServer:
View storybook at: https://wix-wix-ui-icons-common-pr-478.surge.sh

PR #487, opened 2018-05-09
Remove Text component

alisey:
The JSS Text component is used only in backoffice, and I have a PR to remove its usage from there.


WixBuildServer:
View storybook at: https://wix-wix-ui-icons-common-pr-487.surge.sh

WixBuildServer:
View storybook at: https://wix-wix-ui-core-pr-487.surge.sh

PR #489, opened 2018-05-09
chore(StoryPage): support explicit displayName

argshook:
story config now supports explicit `displayName`. if it is set, the one from parsed data will not be used.

this is to circumvent bug in react-docgen-typescript which ignores `displayName` and falls back on file name instead https://github.com/styleguidist/react-docgen-typescript/issues/82

not the best solution in the world, but will hold for a while

### DIFF
--- a/packages/wix-storybook-utils/src/AutoExample/index.js
+++ b/packages/wix-storybook-utils/src/AutoExample/index.js
@@ -328,6 +328,7 @@ export default class extends Component {
           />
 
         <Code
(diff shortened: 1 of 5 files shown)


WixBuildServer:
View storybook at: https://wix-wix-ui-icons-common-pr-489.surge.sh

WixBuildServer:
View storybook at: https://wix-wix-ui-core-pr-489.surge.sh

PR #504, opened 2018-05-21
Pagination: disabled state

alisey:


packages/wix-ui-core/src/components/Pagination/PageStrip.tsx
```diff
@@ -132,10 +133,10 @@ export class PageStrip extends React.Component<PageStripProps, PageStripState> {
           data-hook={`page-${pageNumber}`}
           aria-label={`Page ${pageNumber}`}
           className={style.pageButton}
-          tabIndex={pageUrl ? null : 0}
-          onClick={e => this.props.onPageClick(e, pageNumber)}
-          onKeyDown={e => this.props.onPageKeyDown(e, pageNumber)}
-          href={pageUrl ? pageUrl(pageNumber) : null}
+          tabIndex={disabled || pageUrl ? null : 0}
+          onClick={disabled ? null : (e => this.props.onPageClick(e, pageNumber))}
```
argshook: @alisey 
isn't this a bug?
```js
const a = null;
a(); // Uncaught TypeError: null is not a function
```

alisey: @argshook no, React has a special treatment for `null` and `undefined`  in event handlers: https://stackblitz.com/edit/react-rgnmtr?file=index.js

PR #519, opened 2018-05-27
wix-ui-core: [WIP] <Popover> - Multiple Popover CSS BUG

hepiyellow:
Found a bug in Popover.
If you have multiple Popovers in a page,
In this case both attach to the window (I didn't test other attach strategies),
and you close one Popover, the the other Popover's css is gone !

I think it has something to do with our hack to apply/remove css from a portal.

To see the bug in action:
Open Popover's story (in this PR's storybook),
and under the examples part, click the button saying "Click to close 1"

### DIFF
--- a/packages/wix-ui-core/stories/Popover.story.tsx
+++ b/packages/wix-ui-core/stories/Popover.story.tsx
@@ -2,6 +2,57 @@ import * as React from 'react';
 import {Popover} from '../src/components/Popover';
 import {Option} from '../src/baseComponents/DropdownOption';
 
+const exampleMargin = '50px';
+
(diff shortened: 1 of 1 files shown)


shlomitc:
The issue is related to this logic:
https://github.com/wix/wix-ui/blob/master/packages/wix-ui-core/src/components/Popover/Popover.tsx#L227

Whenever a popover is opened in the `body` context, its styles are applied to it manually using `attachStylesToNode`.
We can extend the logic by adding some sort of a counter / unique identifier for each opened popover, so even if more then one is open and close, it will remove only when all are removed

hepiyellow:
Will be fixed separatelt

PR #528, opened 2018-05-29
AddressInput - Exposing styles object

giladartzi:


lbelinsk:
@nadavov just thought it is important for you to see this pr.
It adds inline styles support only because Santa needs it.
Can it be an issue in the future for the style-pannel?

hepiyellow:
@giladartzi @lbelinsk @shlomitc 
1) First of all the type of the style prop should be `React.CSSProperties`.
2) It would be nice if you extract an interface `WithInlineStyleProp` and put is in `src/types/BaseProps.ts`. And use it in all the component's Props that you appended.
3) You could add the inlineStyle support in `wix-ui-santa`, by having a HOC that puts it on the root DOM node of the rendered component,  after componentDidMount(). Is that a good enough solution for you?

PR #531, opened 2018-05-30
wix-ui-core: <Popover> - Fix attachTo window

### LINKED ISSUE #523: wix-ui-core: <Popover> - Multiple Popover CSS BUG
Found a bug in Popover.
If you have multiple Popovers in a page,
In this case both attach to the window (I didn't test other attach strategies),
and you close one Popover, the the other Popover's css is gone !

I think it has something to do with our hack to apply/remove css from a portal.

To see the bug in action:
Open Popover's story (in this PR's storybook),
and under the examples part, click the button saying "Click to close 1"

See #519

hepiyellow:
Fixes #523 

Also fixes issue with React 15, where setting the attachTo to `document.body` is casing the popover to overwrite the `body` instead of appending to it.

### DIFF
--- a/packages/wix-ui-core/src/components/Popover/Popover.spec.tsx
+++ b/packages/wix-ui-core/src/components/Popover/Popover.spec.tsx
@@ -106,7 +106,6 @@ describe('Popover', () => {
         expect(document.body.classList).not.toContain(styles.root);
         wrapper.unmount();
       });
-
     });
 
     describe('viewport', () => {
@@ -173,8 +172,11 @@ describe('Popover', () => {
 
         wrapper.unmount();
       });
+
     });
   });
+
+
 
   describe('testkit', () => {
     it('should exist', () => {
@@ -187,4 +189,5 @@ describe('Popover', () => {
       expect(isEnzymeTestkitExists(createPopover(), enzymePopoverTestkitFactory, mount)).toBe(true);
     });
   });
+
 });
--- a/packages/wix-ui-core/src/components/Popover/Popover.spec.tsx
+++ b/packages/wix-ui-core/src/components/Popover/Popover.spec.tsx
@@ -203,8 +203,6 @@ describe('Popover', () => {
     });
   });
 
-
-
   describe('testkit', () => {
     it('should exist', () => {
       expect(isTestkitExists(createPopover(), popoverTestkitFactory)).toBe(true);
--- a/packages/wix-ui-core/src/components/Popover/Popover.spec.tsx
+++ b/packages/wix-ui-core/src/components/Popover/Popover.spec.tsx
@@ -1,8 +1,9 @@
 import * as React from 'react';
 import {popoverDriverFactory} from './Popover.driver';
 import {createDriverFactory} from 'wix-ui-test-utils/driver-factory';
-import {Popover, PopoverProps} from './';
-import {mount} from 'enzyme';
+import {Popover, PopoverProps, AppendTo} from './';
+import {Boundary} from 'popper.js';
+import {mount, ReactWrapper} from 'enzyme';
 import {ReactDOMTestContainer} from '../../../test/dom-test-container';
 import * as eventually from 'wix-eventually';
 import {popoverTestkitFactory as enzymePopoverTestkitFactory} from '../../testkit/enzyme';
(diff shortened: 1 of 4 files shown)


packages/wix-ui-core/src/components/Popover/Popover.tsx
```diff
@@ -260,18 +248,46 @@ export class Popover extends React.Component<PopoverType, PopoverState> {
...
+    this.appendToNode = this.getOrCreateAppendToNode({appendTo, targetRef: this.targetRef});
+    // TODO: I think we should throw an error is this.appendToNode is undefined, since 
+    // react-portal will attach it to a new default div under document.body,
+    // and we could not add the stylable styles to that div.
     if (this.appendToNode) {
-      this.stylesObj = style('root', {}, props);
-      this.applyStylesToAppendedNode(props);
+      // Why do we do this here ?(in componentDidMount and not ONLY in render? or when we actually attachStylesToNode)
+      this.stylesObj = style('root', {}, this.props);
+      // TODO: remove this, it is called in render
```
hepiyellow: @shlomitc do you agree?

alisey:
Can you please hold off merging, I have #539 which affects all the tests to make them compatible with the browser environment, want to merge them asap before people start changing more tests.

hepiyellow:
@alisey This is quite urgent, are you merging today?

packages/wix-ui-core/src/components/Popover/Popover.spec.tsx
```diff
@@ -80,97 +81,129 @@ describe('Popover', () => {
 
   describe('appendTo', () => {
     let wrapper, driver;
+    const createDriverFromWrapper = (wrapper2: ReactWrapper) => popoverDriverFactory({element: wrapper2.children().at(0).getDOMNode(), eventTrigger: null});
```
shlomitc: wrapper2 is an aweful name

packages/wix-ui-core/src/components/Popover/Popover.tsx
```diff
@@ -163,33 +164,19 @@ export class Popover extends React.Component<PopoverType, PopoverState> {
 
   static propTypes = {
     className: string,
-    /** The location to display the content */
```
shlomitc: let's not remove the comments, it is still in use by the `wix-ui-adi` Notification component (not in build time)

packages/wix-ui-core/src/components/Popover/Popover.tsx
```diff
@@ -223,14 +210,14 @@ export class Popover extends React.Component<PopoverType, PopoverState> {
     return this.wrapWithAnimations(popper);
   }
 
-  applyStylesToAppendedNode(props) {
-    const {shown} = props;
-    const shouldAnimate = shouldAnimatePopover(props);
+  applyStylesToAppendedNode() {
```
shlomitc: `applyStylesToPortaledNode`

packages/wix-ui-core/src/components/Popover/Popover.tsx
```diff
@@ -255,27 +242,38 @@ export class Popover extends React.Component<PopoverType, PopoverState> {
...
+    const {appendTo} = this.props;
+    this.appendToNode = getAppendToNode({appendTo, targetRef: this.targetRef});
     if (this.appendToNode) {
-      this.stylesObj = style('root', {}, props);
-      this.applyStylesToAppendedNode(props);
+      this.portalNode = document.createElement('div');
+      this.appendToNode.appendChild(this.portalNode);
+      // Why do we do this here ?(in componentDidMount and not ONLY in render? or when we actually attachStylesToNode)
+      this.stylesObj = style('root', {}, this.props);
+      // TODO: remove this, it is called in render
```
shlomitc: you can try removing it but I'm not sure what will be the effect of it

packages/wix-ui-core/src/components/Popover/Popover.spec.tsx
```diff
@@ -80,97 +81,129 @@ describe('Popover', () => {
 
   describe('appendTo', () => {
     let wrapper, driver;
+    const createDriverFromWrapper = (_wrapper: ReactWrapper) => popoverDriverFactory({element: _wrapper.children().at(0).getDOMNode(), eventTrigger: null});
```
alisey: In the master branch we have the following, which basically does what you need:
https://github.com/wix/wix-ui/blob/master/packages/wix-ui-core/src/components/Popover/Popover.spec.tsx#L19

alisey:
Sorry, but the tests are now completely unreadable. The control flow is all over the place, and from looking at a single test I cannot tell what it's doing, and whether it's doing it correctly. Verbosity is not a bad thing for tests.

And why did you have to change classList back to className? `expect(getExpectedContentParent().className).not.toBe(styles.root);` - this doesn't test anything, the class name could be `root groot`, and the test would not catch that.

hepiyellow:
OK:
- I removed the first `afterEach` which did the `unmount` and did is verbosely in each test. Maybe that simplifies the fact that there are 2 `afterEach`'s.
- I re-added the use of `classList` - I missed that in the conflict resolution. Thanks !

See PR: https://github.com/wix/wix-ui/pull/545

PR #535, opened 2018-05-31
Slider: remove most of the layout measurements

InterAl:


packages/wix-ui-core/src/components/Slider/Slider.st.css
```diff
@@ -12,10 +12,8 @@
 }
 
 .root {
-  -st-states: orientation(enum(vertical, horizontal)), tickMarksPosition(enum(normal, middle, across)), tickMarksShape(enum(line, dot)), dir(enum(ltr, rtl)), showTicks, disabled;
+  -st-states: orientation(enum(vertical, horizontal)), tickMarksPosition(enum(normal, middle, across)), tickMarksShape(enum(line, dot)), dir(enum(ltr, rtl)), disabled;
```
lbelinsk: Didn't you tell me that the vertical orientation is not needed?
If so, shouldn't this `orientation` state be removed

packages/wix-ui-core/src/components/Slider/Slider.tsx
```diff
@@ -35,30 +35,17 @@ export interface Style {
...
-  innerRect: Rect;
-  trackRect: Rect;
 }
 
 export class Slider extends React.PureComponent<SliderProps, SliderState> {
   root: HTMLDivElement;
-  inner: HTMLDivElement;
   track: HTMLDivElement;
-  ContinuousStep = 0.1;
+  ContinuousStep = 0.01;
```
lbelinsk: why capital C?
because it is a constant?
ain't it more common to have it with lower case or all capital cased?  `CONTINUOUS_STEP`?

packages/wix-ui-core/src/components/Slider/Slider.st.css
InterAl: It is needed, just not for santa.

packages/wix-ui-core/src/components/Slider/Slider.tsx
InterAl: Changed.

As for a context: I removed all the render-time measurements from the component. Reason: Santa.

PR #541, opened 2018-06-03
wix-ui-core: Fix TS issues: add import Validator

hepiyellow:


hepiyellow:
Instead of this, we should stay with Typescript 2.8

PR #544, opened 2018-06-03
Improve speed and stability of AddressInput tests

alisey:
They were using huge timeouts and failing randomly because of race conditions.

### DIFF
--- a/packages/wix-ui-core/src/components/AddressInput/AddressInput.spec.tsx
+++ b/packages/wix-ui-core/src/components/AddressInput/AddressInput.spec.tsx
@@ -32,6 +32,7 @@ describe('AddressInput', () => {
           Client={GoogleMapsClientStub}
           onSelect={onSelectSpy}
           handler={handler || Handler.geocode}
+          throttleInterval={0}
           {...rest}
         />);
     };
@@ -55,14 +56,14 @@ describe('AddressInput', () => {
     });
 
     it('Should throttle calls to MapsClient.autocomplete', async () => {
-        init();
+        init({throttleInterval: 10});
         driver.setValue('n');
         driver.setValue('ne');
         driver.setValue('new');
         await eventually(() => {
             expect(GoogleMapsClientStub.prototype.autocomplete).toHaveBeenCalledWith(helper.API_KEY, 'en', {input: 'n'});
             expect(GoogleMapsClientStub.prototype.autocomplete).toHaveBeenCalledWith(helper.API_KEY, 'en', {input: 'new'});
-        });
+        }, {interval: 5});
         expect(GoogleMapsClientStub.prototype.autocomplete).toHaveBeenCalledTimes(2);
     });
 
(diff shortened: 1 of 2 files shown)


packages/wix-ui-core/src/components/AddressInput/AddressInput.spec.tsx
```diff
@@ -32,6 +32,7 @@ describe('AddressInput', () => {
           Client={GoogleMapsClientStub}
           onSelect={onSelectSpy}
           handler={handler || Handler.geocode}
+          throttleInterval={0}
```
alisey: The only test that relies on throttling of HTTP requests to the geocoding service overrides this value, for all other tests it should not matter.

packages/wix-ui-core/src/components/AddressInput/AddressInput.spec.tsx
```diff
@@ -55,14 +56,14 @@ describe('AddressInput', () => {
...
-        init();
+        init({throttleInterval: 10});
         driver.setValue('n');
         driver.setValue('ne');
         driver.setValue('new');
         await eventually(() => {
             expect(GoogleMapsClientStub.prototype.autocomplete).toHaveBeenCalledWith(helper.API_KEY, 'en', {input: 'n'});
             expect(GoogleMapsClientStub.prototype.autocomplete).toHaveBeenCalledWith(helper.API_KEY, 'en', {input: 'new'});
-        });
+        }, {interval: 5});
```
alisey: By default `eventually()` tests the condition every 200ms. There's no reason to wait so long.

packages/wix-ui-core/src/components/AddressInput/AddressInput.spec.tsx
```diff
@@ -313,6 +318,7 @@ describe('AddressInput', () => {
             driver.setValue('n');
 
             const secondRequest = GoogleMapsClientStub.setAddressesPromise([helper.ADDRESS_2]);
+            driver.click();
```
alisey: We're trying to simulate the following:

1. The user clicks on the input.
2. Types 'n' into the input.
3. Clicks the first option in the dropdown.
4. Types 'e' into the input.
5. Clicks the first option in the dropdown.

The problem was before steps 3 and 4. Selecting an option blurs the input, and you cannot type into it without clicking on it first. Of course, the simulation can do that, but the dropdown only appears if the input has been clicked on.

The test only worked because when you select an option the dropdown doesn't disappear immediately, it waits for 150ms, so if the test is running fast enough, the dropdown is still there by the time we get to the step 4.

packages/wix-ui-core/src/components/AddressInput/AddressInput.spec.tsx
shlomitc: We can configure eventually to have a shorter interval once and reuse it across all tests

```
import * as wixEventually from 'wix-eventually';

const defaultTimeouts = { timeout: 500, interval: 10 };
export async function eventually(condition, timeouts = defaultTimeouts) {
  return await wixEventually.with(Object.assign({}, defaultTimeouts, timeouts))(() => condition());
}
```

alisey: Were would you suggest adding that?

shlomitc: it can either sit in `test/test-common` or if you think this is essential to all libraries, it can sit in `wix-ui-test-utils`

packages/wix-ui-core/src/components/AddressInput/AddressInput.spec.tsx
shlomitc: what do you think about using fake timers?
https://facebook.github.io/jest/docs/en/timer-mocks.html

I wonder if it will work when running tests in browser

alisey: I think it could be a good idea, but we need to consider a few things:
1. `jest.useFakeTimers()` depends on `process` and `fs` and will not work in browser.
2. Faking the timers affects all functions, whether you want it or not, e.g. Mocha's timeout handler, sleep(), eventually(), maybe XMLHTTPRequest, so we'd have to be careful.
3. If the component saves its own local reference to setTimeout, then it's not possible to override it, but this is not a big deal really.
4. A component could rely on timeouts executing in a certain order, e.g. when you do setTimeout(a, 10); setTimeout(b, 30); setTimeout(c, 20) you expect the callbacks to run in a sequence: a, c, b instead of a, b, c. Fake timers violate that invariant, and you cannot be sure that the component works correctly even if the tests pass.

PR #547, opened 2018-06-04
Create private drivers for Input and Pagination

alisey:
And move things around in the IframesManager

### DIFF
--- a/packages/wix-ui-core/src/components/Pagination/Pagination.private.driver.ts
+++ b/packages/wix-ui-core/src/components/Pagination/Pagination.private.driver.ts
@@ -1,0 +1,58 @@
+import {Simulate} from 'react-dom/test-utils';
+
+export class PaginationDriver {
+  constructor(public root: HTMLElement) { }
+
+  get previousButton() {
+    return this.root.querySelector('[data-hook=previous]');
+  }
+
+  get nextButton() {
+    return this.root.querySelector('[data-hook=next]');
+  }
+
+  get firstButton() {
+    return this.root.querySelector('[data-hook=first]');
+  }
+
+  get lastButton() {
+    return this.root.querySelector('[data-hook=last]');
+  }
+
+  get input(): HTMLInputElement {
+    return this.root.querySelector('[data-hook=page-input]');
+  }
+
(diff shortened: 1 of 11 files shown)


packages/wix-ui-core/src/components/Pagination/Pagination.private.driver.ts
```diff
@@ -0,0 +1,58 @@
...
+  get firstButton() {
+    return this.root.querySelector('[data-hook=first]');
+  }
+
+  get lastButton() {
+    return this.root.querySelector('[data-hook=last]');
+  }
+
+  get input(): HTMLInputElement {
+    return this.root.querySelector('[data-hook=page-input]');
```
shlomitc: create a common function that query for the data-hook

alisey: Something like this?
```
get input() {
  return queryHook('page-input', this.root);
}
```

Or this?
```
get input() {
  return this.root.querySelector(hook('page-input'));
}
```

In my order of preference, from ok to less ok, it would be:
1. querySelector('[data-hook=page-input]')
2. querySelector(hook('page-input'))
3. queryHook('page-input', this.root) + queryHookAll('page-input', this.root)

I was thinking about a mixin that takes a list of hook names and creates getters for you automatically, but not sure how to make that play well with TypeScript, and also there are common exceptions, e.g. when you need to specify a type of return value, such as HTMLInputElement, or when you need to use querySelectorAll.

shlomitc: The third option is the one we widely use in many projects so let's stick with it.

shlomitc: we can stick currently to a simple usage and improve it later

alisey: It's so wrong on so many levels, but sure. Gonna add it to wix-ui-test-utils.

shlomitc:
please update the branch

alisey:
@shlomitc all checks have passed

PR #554, opened 2018-06-05
Popover tests: fix failures when running in the browser, refactor

alisey:
1. Reduced global state.
2. Separated tests for portal logic and attachment point logic.
3. Removed dependency on Enzyme.
4. Made the titles clearer.
5. Reduced timeouts for animation tests.
6. Switched to async rendering API.

### DIFF
--- a/packages/wix-ui-core/src/components/Popover/Popover.spec.tsx
+++ b/packages/wix-ui-core/src/components/Popover/Popover.spec.tsx
@@ -1,218 +1,211 @@
 import * as React from 'react';
-import {popoverDriverFactory} from './Popover.driver';
-import {createDriverFactory} from 'wix-ui-test-utils/driver-factory';
-import {Popover, PopoverProps, AppendTo} from './';
-import {Boundary} from 'popper.js';
-import {mount, ReactWrapper} from 'enzyme';
+import {Simulate} from 'react-dom/test-utils';
+import {queryHook} from 'wix-ui-test-utils/dom';
+import {Popover, PopoverProps} from './';
 import {ReactDOMTestContainer} from '../../../test/dom-test-container';
 import * as eventually from 'wix-eventually';
-import {popoverTestkitFactory as enzymePopoverTestkitFactory} from '../../testkit/enzyme';
-import {isEnzymeTestkitExists} from 'wix-ui-test-utils/enzyme';
-import {isTestkitExists} from 'wix-ui-test-utils/vanilla';
-import {popoverTestkitFactory} from '../../testkit';
 import styles from './Popover.st.css';
 
+const popoverWithProps = (props: PopoverProps) => (
+  <Popover {...props}>
+    <Popover.Element>
+      <div>Element</div>
+    </Popover.Element>
+    <Popover.Content>
+      <div>Content</div>
+    </Popover.Content>
+  </Popover>
+);
+
+// Since Popover.Content can render outside the component's root, let's query
+// the entire document with the assumption that we don't render more than one
+// popover at a time.
+const queryPopoverElement = () => queryHook<HTMLElement>(document, 'popover-element');
+const queryPopoverContent = () => queryHook<HTMLElement>(document, 'popover-content');
+const queryPopoverArrow   = () => queryHook<HTMLElement>(document, 'popover-arrow');
+const queryPopoverPortal  = () => queryHook<HTMLElement>(document, 'popover-portal');
+
 describe('Popover', () => {
-  let wrapper;
+  describe('Display', () => {
+    const container = new ReactDOMTestContainer().destroyAfterEachTest();
+
+    it(`doesn't display popup by default`, async () => {
+      await container.render(popoverWithProps({
+        placement: 'bottom',
+        shown: false
+      }));
   
-  afterEach(() => {
-    if (wrapper) {
-      wrapper.unmount();
-    }
+      expect(queryPopoverElement()).toBeTruthy();
+      expect(queryPopoverContent()).toBeNull();
+    });
+
+    it(`displays popup when shown={true}`, async () => {
+      await container.render(popoverWithProps({
+        placement: 'bottom',
+        shown: true
+      }));
+      
+      expect(queryPopoverContent()).toBeTruthy();
+    });
   });
 
-  const container = new ReactDOMTestContainer().unmountAfterEachTest();
+  describe('Events', () => {
+    const container = new ReactDOMTestContainer().destroyAfterEachTest();
 
-  const render = container.createLegacyRenderer(popoverDriverFactory);
+    it(`calls mouseEnter and mouseLeave callbacks`, async () => {
+      const onMouseEnter = jest.fn();
+      const onMouseLeave = jest.fn();
+      await container.render(popoverWithProps({
+        placement: 'bottom',
+        shown: false,
+        onMouseEnter,
+        onMouseLeave,
+      }));
 
-  const renderWithEnzyme = (jsx, {attachTo} = {attachTo: container.node}) => {
-    wrapper = mount<PopoverProps,{}>(jsx, {attachTo});
-    const driver = popoverDriverFactory({
-      element: wrapper.find(Popover).getDOMNode(),
-      eventTrigger: null
+      Simulate.mouseEnter(container.componentNode);
+      Simulate.mouseLeave(container.componentNode);
+      expect(onMouseEnter).toBeCalled();
+      expect(onMouseLeave).toBeCalled();
     });
-    return {wrapper, driver};
-  };
+  });
 
-  const createPopover = (props: Partial<PopoverProps> = {}) => (
-    <Popover placement="top" showArrow shown={false} {...props}>
-      <Popover.Element>
-        <div>
-          Element
+  describe('Position', () => {
+    const container = new ReactDOMTestContainer().destroyAfterEachTest();
+
+    it(`offsets the arrow by specified amount`, async () => {
+      await container.render(popoverWithProps({
+        placement: 'bottom',
+        shown: true,
+        showArrow: true,
+        moveArrowTo: 10
+      }));
+
+      expect(queryPopoverArrow().style.left).toBe('10px');
+    });
+  });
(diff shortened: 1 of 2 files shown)


alisey:
E2e tests are failing in the master, but this is still good to review. This is the last blocker before I can try running browser tests on CI.

shlomitc:
@hepiyellow - please review

packages/wix-ui-core/src/components/Popover/Popover.spec.tsx
```diff
@@ -1,218 +1,211 @@
...
+    <Popover.Element>
+      <div>Element</div>
+    </Popover.Element>
+    <Popover.Content>
+      <div>Content</div>
+    </Popover.Content>
+  </Popover>
+);
+
+// Since Popover.Content can render outside the component's root, let's query
```
hepiyellow: FYI: Actually, I had bugs related to multiple popovers in a page. I just didn't add the tests yet. I added tests that check that our implementation is correct, but not behavioral tests which test multiple popovers behavior. Anyways, I think the Popover driver was searching the body, in the same way you do.

packages/wix-ui-core/src/components/Popover/Popover.spec.tsx
```diff
@@ -1,218 +1,211 @@
...
+const queryPopoverPortal  = () => queryHook<HTMLElement>(document, 'popover-portal');
+
 describe('Popover', () => {
-  let wrapper;
+  describe('Display', () => {
+    const container = new ReactDOMTestContainer().destroyAfterEachTest();
+
+    it(`doesn't display popup by default`, async () => {
+      await container.render(popoverWithProps({
+        placement: 'bottom',
```
hepiyellow: That's why I had `createPopover` which accepts Partial<PopoverProps>, so I don't have to pass props which are not related to a test.

packages/wix-ui-core/src/components/Popover/Popover.spec.tsx
alisey: > I had bugs related to multiple popovers in a page

Bugs when you render multiple popovers and they don't play well together, or when you mount one popover, but two are mounted for some reason?

packages/wix-ui-core/src/components/Popover/Popover.spec.tsx
```diff
@@ -1,218 +1,211 @@
...
+      <div>Content</div>
+    </Popover.Content>
+  </Popover>
+);
+
+// Since Popover.Content can render outside the component's root, let's query
+// the entire document with the assumption that we don't render more than one
+// popover at a time.
+const queryPopoverElement = () => queryHook<HTMLElement>(document, 'popover-element');
+const queryPopoverContent = () => queryHook<HTMLElement>(document, 'popover-content');
```
hepiyellow: A variant for this `queryPopoverContent ` exists in the Popover.driver as `getContentElement`, why don't you use it?

packages/wix-ui-core/src/components/Popover/Popover.spec.tsx
alisey: I actually like passing all props explicitly, as opposed to having to figure out what are the defaults and how my props are going to be merged with those, and what to do if I want to omit some of the defaults.

alisey: here I pass the props, and I know that the component is going to receive those props and nothing else.

packages/wix-ui-core/src/components/Popover/Popover.spec.tsx
hepiyellow: rendering multiple popovers and the don't play together.

packages/wix-ui-core/src/components/Popover/Popover.spec.tsx
```diff
@@ -1,218 +1,196 @@
...
-    if (wrapper) {
-      wrapper.unmount();
-    }
-  });
-
-  const container = new ReactDOMTestContainer().unmountAfterEachTest();
-
-  const render = container.createLegacyRenderer(popoverDriverFactory);
+      expect(queryPopoverElement()).toBeTruthy();
+      expect(queryPopoverContent()).toBeNull();
```
hepiyellow: For readability, I'd rather having eplicit method for `isPopoverContentExists()`.

In anyway, I would rename `queryPopoverContent()` to `getPopoverContentElement()`.

alisey: Good point on `getPopoverContentElement`. About `isPopoverContentExists`, for readability I would rather not add it.

alisey: Wait, having `queryPopoverElement` and `getPopoverContentElement` would be confusing.

packages/wix-ui-core/src/components/Popover/Popover.spec.tsx
alisey: Because it's so much more clear to have these functions directly in the test file, and not to go through all the levels of pointless indirection.

alisey: Also drivers are going to be for consumers, and we're not going to use them in the unit tests.

packages/wix-ui-core/src/components/Popover/Popover.spec.tsx
```diff
@@ -1,218 +1,196 @@
...
+        showArrow: true,
+        moveArrowTo: 10
+      }));
 
-  it('should call mouse leave callback', () => {
-    const onMouseLeave = jest.fn();
-    const driver = render(createPopover({onMouseLeave}));
-    driver.mouseLeave();
-    expect(onMouseLeave).toBeCalled();
+      expect(queryPopoverArrow().style.left).toBe('10px');
```
hepiyellow: The reason we use drivers is to abstract implementation away. you are exposing the fact that we use `style.left`, we could have used `translate`. Why don't you use the `PopoverDriver`?

packages/wix-ui-core/src/components/Popover/Popover.spec.tsx
alisey: > rendering multiple popovers and they don't play together

If we ever decide to write tests for this, we'll change Popover to expose refs for the portal, content, and element, and we'll use those in tests. Not too difficult.

packages/wix-ui-core/src/components/Popover/Popover.spec.tsx
alisey: Abstract implementation from unit tests? No. Unit tests should know about implementation, and should test that implementation.

Suppose I have `isDisabled()` method in my driver, and suppose I check it in my unit test, what exactly am I checking? That the prop is being passed to the component, or that it applies a certain style state, or that the component is not clickable by mouse, or that it cannot be tabbed into by keyboard, or that the native input receives `disabled` attribute? Those all are different unit tests that each has its own concept of what `isDisabled` is. Any abstraction just obscures that and gives you false confidence.

In this specific case I agree that it would be better to take the position from `getBoundingClientRect`. But it's a small detail not relevant to the goal of this PR.

packages/wix-ui-core/src/components/Popover/Popover.spec.tsx
hepiyellow: I was talking about renaming `queryPopoverElement` -> `getPopoverContentElement`

packages/wix-ui-core/src/components/Popover/Popover.spec.tsx
hepiyellow: Create a private driver. Just let the test (`it()`) be readable and free of implementation.

packages/wix-ui-core/src/components/Popover/Popover.spec.tsx
alisey: We have `<Popover.Element>` and `<Popover.Content>`.

`queryPopoverElement` finds `Popover.Element`
`queryPopoverContent` finds `Popover.Content`

In this picture renaming queryPopoverElement -> getPopoverContentElement seems to add to confusion.

packages/wix-ui-core/src/components/Popover/Popover.spec.tsx
alisey: We seem to have very different ideas of readable. Can you please just approve the PR, and then change anything you don't like.

packages/wix-ui-core/src/components/Popover/Popover.tsx
```diff
@@ -261,6 +261,7 @@ export class Popover extends React.Component<PopoverType, PopoverState> {
     this.appendToNode = getAppendToNode({appendTo, targetRef: this.targetRef});
     if (this.appendToNode) {
       this.portalNode = document.createElement('div');
+      this.portalNode.setAttribute('data-hook', 'popover-portal');
```
hepiyellow: nice!

packages/wix-ui-core/src/components/Popover/Popover.spec.tsx
```diff
@@ -1,218 +1,196 @@
...
+      );
+
+      expect(queryPopoverPortal().parentElement).toBe(container.node.firstChild);
     });
-  });
 
-  describe('enzyme testkit', () => {
-    it('should exist', () => {
-      expect(isEnzymeTestkitExists(createPopover(), enzymePopoverTestkitFactory, mount)).toBe(true);
+    it(`adds the portal to the specified element when appendTo={element}`, async () => {
```
hepiyellow: This was tested already. Can remove it.

alisey: Right, I wasn't sure about whether to keep this test since it basically duplicates the one before, but then I thought: what if `renders the popup into a portal when given appendTo prop` used `viewport` for the portal, then I would need this test here, and why one test's implementation should affect whether another test exists.

alisey: And removed.

PR #555, opened 2018-06-05
Adding constants to the timepicker exports

guybenron:


packages/wix-ui-core/src/components/TimePicker/index.ts
```diff
@@ -1 +1,3 @@
 export {TimePicker, TimePickerProps} from './TimePicker';
+import * as TimePickerConstants from './constants';
```
lbelinsk: Can Use 
```
export * from './constants' or something similar 
```

But why do we need to export this?

PR #556, opened 2018-06-05
Migrated to svg2react-icon v3

Faradey27:
Screenshots
v2:
http://prntscr.com/jr6weu
http://prntscr.com/jr6wir
v3:
http://prntscr.com/jr6vbc
http://prntscr.com/jr6vht

from my perspective looks fine

### DIFF
--- a/packages/wix-ui-icons-common/package.json
+++ b/packages/wix-ui-icons-common/package.json
@@ -45,8 +45,8 @@
     "protractor": "^5.3.0",
     "react": "^16.3.1",
     "react-dom": "^16.3.1",
-    "svg2react-icon": "^2.0.0",
     "typescript": "~2.8.3",
+    "svg2react-icon": "^3.0.1",
     "wix-storybook-utils": "^1.0.0",
     "wix-ui-test-utils": "^1.0.0",
     "yoshi": "^1.2.0"


shlomitc:
maybe we can add some eyes tests screenshot to make sure it is not broken?
will probably need to be in a different PR

Faradey27:
@shlomitc will create separate pull request with eyes tests and after merging it let merge this one

alisey:
Don't merge just yet, there's a bug here: https://github.com/wix/svg2react-icon/blob/master/src/lib/svg2component.js#L16

alisey:
Another bug here: https://github.com/wix/svg2react-icon/blob/master/src/lib/svg2component.js#L43

`${widthFromSvg}` needs to be in quotes, otherwise if the SVG doesn't have width, the compiled code is going to be this:
```
width={props.size || 1em}
```

alisey:
Another problem here: https://github.com/wix/svg2react-icon/blob/master/src/lib/svg2component.js#L29
The size prop ends up directly on the SVG element, we should exclude it from props like this: `({size, ...props}) => {...}`

Sorry for not giving a more thorough review earlier today, I was tired and distracted.

Faradey27:
@alisey https://github.com/wix/svg2react-icon/pull/17

Faradey27:
Pull request update to svg2react-icon@^3.0.1 with fixes

@shlomitc

shlomitc:
CI is failing now (issue with npm registry), let's make it pass and see that tests are running and then we can merge

PR #560, opened 2018-06-12
`<Popover/>` - Add placement exampleProps to story

hepiyellow:


hepiyellow:
Already done in another PR

PR #563, opened 2018-06-13
Mocha runner: bump the version

alisey:
Forgot to do this in the previous PR :(


shlomitc:
Instead of bumping manually, you can use `wnpm-ci` that will auto-bump the version in CI
https://github.com/wix/wnpm-ci

I'll merge this one, but consider using it instead

alisey:
@shlomitc thanks, didn't know about that

PR #573, opened 2018-06-14
Browser test runner: the Yoshi way

alisey:
So `wix-ui-mocha-runner` works fine when `wix-ui-core` installs all its dependencies, but when it's linked to the other `wix-ui-*` repos there is some issue with duplicate `.d.ts` files, probably related to how the monorepo is configured, I gave up on figuring out to solve it. Using Karma now.

### DIFF
--- a/packages/wix-ui-core/karma.conf.js
+++ b/packages/wix-ui-core/karma.conf.js
@@ -1,0 +1,10 @@
+module.exports = {
+  frameworks: ['mocha'],
+  reporter: 'spec',
+  browsers: ['ChromeHeadless'],
+  client: {
+    mocha: {
+      reporter: 'html',
+    }
+  }
+};
--- a/packages/wix-ui-core/test/karma-setup.js
+++ b/packages/wix-ui-core/test/karma-setup.js
@@ -1,0 +1,12 @@
+import expect from 'expect';
+import JestMock from 'jest-mock';
+import Enzyme from 'enzyme';
+import EnzymeAdapter from 'enzyme-adapter-react-16';
+
+// Jest compatibility
+window.expect = expect;
+window.jest = JestMock;
+window.beforeAll = window.before;
+window.afterAll = window.after;
+
+Enzyme.configure({adapter: new EnzymeAdapter()});
--- a/packages/wix-ui-core/package.json
+++ b/packages/wix-ui-core/package.json
@@ -26,7 +26,7 @@
     "test": "npm run test:unit && npm run test:e2e",
     "posttest": "npm run lint",
     "test:watch": "yoshi test --jest --watch",
-    "test:unit": "yoshi test --jest",
+    "test:unit": "yoshi test --jest --karma",
     "test:e2e": "yoshi test --protractor",
     "lint": "yoshi lint",
     "yoshi-start": "yoshi start --no-test",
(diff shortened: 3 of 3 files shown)


packages/wix-ui-core/test/karma-setup.js
```diff
@@ -0,0 +1,12 @@
+import expect from 'expect';
```
shlomitc: where is this `expect` coming from in the dependency tree?

packages/wix-ui-core/karma.conf.js
```diff
@@ -0,0 +1,10 @@
+module.exports = {
+  frameworks: ['mocha'],
```
shlomitc: we can use `jasmine` and keep the same `jest` syntax and matchers

packages/wix-ui-core/test/karma-setup.js
alisey: from jest

packages/wix-ui-core/karma.conf.js
alisey: What do you mean by syntax? What's the benefit of using Jasmine instead of expect?

alisey: Tried Jasmine. There are incompatibilities in `expect` and `describe` functions.

alisey:
This is ready for review/merge.

packages/wix-ui-core/karma.conf.js
shlomitc: hmm, so if we use the `expect` from `jest` we're maintaining the same matchers, so that's actually working already. nevermind

shlomitc:
looks good. are there any differences regarding testing runtime, deubgging, sourcemaps, etc..?

alisey:
Debugging experience is utter crap, which is why I didn't consider Karma up to this point. First, you need to manually adjust karma.conf.js to launch in windowed mode or to enable remote debugging. Source maps are not working, reporting is very awkward, even when running in windowed mode failures are logged into terminal without any formatting, coloring or line numbers.

We don't have many options though. People are writing new tests that don't work in browser, and we need some kind of automatic checking. My puppetteer runner provides an OK dev experience, but it would need to be merged into Yoshi to work with this project. And I've no idea how much time that's going to take.

So the hope is that until I merge it or find a way to configure Karma to be not awful we can have this setup which at least prevents regressions in the tests. Most of the problems can be debugged in Node + Jest, and for the few remaining browser-only issues `console.log` is our friend.

shlomitc:
@ranyitz - I think this requires

shlomitc:
Thanks, @alisey.
@ranyitz - I think this requires some thinking if Karma is still a relevant tool for our use case or should we consider adding the `mocha-loader` solution into yoshi

alisey:
Karma is faster, and it already exists in Yoshi, adding Puppeteer would duplicate existing functionality and make `npm install` take more time for everyone, even if they don't use Puppetter.

But I'm not sure how much Karma can be customized to our needs. Here's what we get from `wix-ui-mocha-runner`:
* Headless mode
  * uses Mocha spec reporter
  * writes to the terminal in color
  * pretty-prints proxied console output
  * exits with non-zero code on uncaught exceptions / page errors / webpack compilation errors, logs the error in a readable format
* Watch mode
  * opens the browser in windowed mode
  * uses HTML Mocha reporter
  * exits with non-zero code on webpack compilation errors and page errors
  * reports webpack progress during compilation
  * allows to run subset of tests by adding grep pattern to the URL (mocha HTML reporter allows to do this automatically by clicking on one of the tests / groups)
  * allows to set breakpoint in devtools (currently Karma changes the page to about:blank when tests are completed, which makes this impossible)
  * doesn't require opening multiple tabs to see the list of running / completed / failed tests
  * allows to access the test page from a phone / other computer in the same network
  * allows to customize CSS for Mocha / use custom reporters
* Source maps
* Suppressed React Devtools warning

PR #574, opened 2018-06-14
fix styles

shl3vi:


packages/wix-ui-core/src/components/LinearProgressBar/LinearProgressBar.tsx
```diff
@@ -56,17 +56,18 @@ const normalizeProps = (props: LinearProgressBarProps) => {
   return {...props, value};
 }
 
-export const LinearProgressBar = (_props: LinearProgressBarProps) => {
-  const {error, showProgressIndication} = _props;
-  const props = normalizeProps(_props);
+export const LinearProgressBar = (props: LinearProgressBarProps) => {
+  const {error, showProgressIndication} = props;
```
lbelinsk: you are destructuring this showProgressIndication but not using it. You can use it in line 69

PR #579, opened 2018-06-17
wix-ui-core: added propTypes to LinearProgressBar

bchmn:


lbelinsk:
What about default prop for `value`? do you guys think that having default prop = 0 makes sense?

lbelinsk:
build fails.
http://pullrequest-tc.dev.wixpress.com/viewLog.html?buildId=499011&buildTypeId=FedInfra_WixUi_Packages_WixUiCode&tab=buildLog&_focus=251

lbelinsk:
Oh, maybe just add a test for the default value = 0?

PR #584, opened 2018-06-19
`<CircularProgressBar/>` - new component

aaronvine:


packages/wix-ui-core/stories/CircularProgressBar/style.st.css
```diff
@@ -0,0 +1,29 @@
...
+
+.root {
+  -st-extends: CircularProgressBar;
+}
+
+.root::statusIndicator, .root::progressIndicator {
+  color: rgb(50,83,106);
+}
+
+.root::backArc {
```
lbelinsk: The story looks great. the thing is that it looks too good :)
This is the core, let's demonstrate this maybe with grey/darker colors which are the minimal requirements in order to make the story relevant.
Maybe some darker gray on top of lighter gray.

packages/wix-ui-core/src/components/CircularProgressBar/Arc.tsx
```diff
@@ -0,0 +1,71 @@
+import * as React from 'react';
```
lbelinsk: What are the differences from the Loader's one from wsr?
This is just something we would like to think about later when we start working on the Loader...
It is ok to have it here now

packages/wix-ui-core/src/components/CircularProgressBar/CircularProgressBar.protractor.driver.ts
```diff
@@ -0,0 +1,31 @@
...
+  const findByDataHook = dataHook => element.$(`[data-hook="${dataHook}"]`);
+  const foregroundArc = () => findByDataHook('progressarc-foreground');
+  const getElementAttribute = (e: ElementFinder, attr) => e.getAttribute(attr);
+  const progressIndication = () => findByDataHook('progress-indicator');
+  const getElementIntAttribute = (e: ElementFinder, attr) => new promise.Promise<number>(async resolve => resolve(parseInt(await getElementAttribute(e, attr))));
+
+  return {
+    element: () => element,
+    exists: () => element.isPresent(),
+    getArcValue: () => getElementIntAttribute(foregroundArc(), 'data-value'),
```
alisey: We should stop this practice of adding attributes to the DOM just for tests. If you want to test that the component receives the correct prop use Enzyme. But if you want to check that the arc has the correct appearance, then using a data attribute is just fooling yourself, even if you receive the correct value here, it absolutely does not mean that you're generating a correct `d` attribute for the path. Checking appearance is what screenshot tests are for.

I suggest either using Enzyme or adding this value as a title attribute to the SVG. Then it would at least serve some purpose for the screen readers.

packages/wix-ui-core/src/components/CircularProgressBar/CircularProgressBar.protractor.driver.ts
```diff
@@ -0,0 +1,31 @@
+import {BaseDriver, DriverFactory} from './../../common/BaseDriver.protractor';
+import {promise, ElementFinder} from 'protractor';
+import { GlobalsNumber } from 'csstype';
+
+export interface CircularProgressBarDriver extends BaseDriver {
+  /** Returns true if the root element is present */
+  exists: () => promise.Promise<boolean>;
+  /** Get the foreground arc value (percentage) */
+  getArcValue: () => promise.Promise<number>;
```
lbelinsk: Let's try to keep the names as similar as possible to the other driver. `getArcValue` --> `getValue` would be less low level for the end user.

packages/wix-ui-core/src/components/CircularProgressBar/CircularProgressBar.protractor.driver.ts
```diff
@@ -0,0 +1,31 @@
...
+  const foregroundArc = () => findByDataHook('progressarc-foreground');
+  const getElementAttribute = (e: ElementFinder, attr) => e.getAttribute(attr);
+  const progressIndication = () => findByDataHook('progress-indicator');
+  const getElementIntAttribute = (e: ElementFinder, attr) => new promise.Promise<number>(async resolve => resolve(parseInt(await getElementAttribute(e, attr))));
+
+  return {
+    element: () => element,
+    exists: () => element.isPresent(),
+    getArcValue: () => getElementIntAttribute(foregroundArc(), 'data-value'),
+    isProgressIndicationDisplayed: () => progressIndication().isPresent(),
```
lbelinsk: I think that for e2e's `isProgressIndicationDisplayed` is not very useful. You are not using it internally also.

packages/wix-ui-core/src/components/CircularProgressBar/CircularProgressBar.st.css
```diff
@@ -0,0 +1,47 @@
...
+  width: 100%;
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+}
+
+.progressIndicator {
+  position: relative;
+  height: 100%;
+  margin-top: 18px;
```
alisey: The base style sheet should only define selectors, states and styles common for all themes. It should not define colors, spacing, fonts - anything that depends on the theme.

packages/wix-ui-core/src/components/CircularProgressBar/CircularProgressBar.e2e.ts
```diff
@@ -0,0 +1,59 @@
...
+  eyes.it('should progress as value increases', async () => {
+    const expectedProgress = 90;
+
+    await autoExampleDriver.setProps({value: expectedProgress});
+    const foregroundArcValue = await driver.getArcValue();
+
+    expect(foregroundArcValue).toBe(expectedProgress);
+  });
+
+  eyes.it('should show empty progress as value less than 0', async () => {
```
lbelinsk: This is really a unit test which you already covers in the jest tests.
It is not rendering anything new for eyes here so I believe we can remove it

lbelinsk: I think one e2e which is enough here, with some percentage

packages/wix-ui-core/src/components/CircularProgressBar/CircularProgressBar.tsx
```diff
@@ -0,0 +1,103 @@
...
+  /** an indication icon (any react component) that will be presented when 'error' and 'showProgressIndication' are set to true */
+  errorIcon?: JSX.Element;
+  /** an indication icon (any react component) that will be presented when 'showProgressIndication' are set to true and 'value' is 100 */
+  successIcon?: JSX.Element;
+}
+
+const FULL_PROGRESS = 100;
+const NO_PROGRESS = 0;
+
+const DEFAULT_SIZE = 54;
```
alisey: wix-ui-core is supposed to contain components that can be styled by the consumer - mainly by extending the base stylesheet. By embedding a bunch of inline styles you completely defeat the purpose of the library.

packages/wix-ui-core/src/components/CircularProgressBar/CircularProgressBar.spec.tsx
```diff
@@ -0,0 +1,111 @@
...
+import { CircularProgressBar } from './';
+import { CircularProgressBarProps } from './CircularProgressBar';
+import { runTestkitExistsSuite } from '../../common/testkitTests';
+import { circularProgressBarTestkitFactory } from '../../testkit';
+import { circularProgressBarTestkitFactory as enzymeCircularProgressBarTestkitFactory } from '../../testkit/enzyme';
+
+describe('CircularProgressBar', () => {
+
+  const createDriver = new ReactDOMTestContainer().unmountAfterEachTest()
+    .createLegacyRenderer(circularProgressBarDriverFactory);
```
alisey: It's called legacy renderer because it's legacy and should not be used for new components. Drivers are supposed to be for consumers, for this component consumers probably only need one method: `getValue`, other things can be added later if the need arises.

In the unit tests you can query DOM directly, like in Popover tests, or define a bunch of helper functions like in Pagination/Input unit tests.

packages/wix-ui-core/src/components/CircularProgressBar/CircularProgressBar.spec.tsx
```diff
@@ -0,0 +1,111 @@
...
+      expect(driver.isErrorIconDisplayed()).toBe(true);
+      expect(driver.getValue()).toBe('Failed');
+    })
+
+    it('should show percentages value while in progress', () => {
+      driver = createDriver(<CircularProgressBar {...{ ...props, value: 50 }} />);
+      expect(driver.getValue()).toBe('50%');
+    })
+
+    it('should show percentages value of 0 when passing value below 0', () => {
```
lbelinsk: Check please also for more than 100

packages/wix-ui-core/src/components/CircularProgressBar/CircularProgressBar.spec.tsx
```diff
@@ -0,0 +1,111 @@
...
+      props = { ...defaultProps, showProgressIndication: true };
+    });
+
+    it('should show success icon when reaching 100% and value passed as string', () => {
+      driver = createDriver(<CircularProgressBar {...{ ...props, value: '100' as any, successIcon: <div /> }} />);
+      expect(driver.isSuccessIconDisplayed()).toBe(true);
+    })
+
+    it('should show success icon and percentage when reaching 100%', () => {
+      driver = createDriver(<CircularProgressBar {...{ ...props, value: 100, successIcon: <div /> }} />);
```
alisey: You probably don't need this double spread.

packages/wix-ui-core/src/components/CircularProgressBar/CircularProgressBar.st.css
lbelinsk: I agree. all the `height: 100%` kind of sizes are ok though, just not the specific spacing, line-height, font-weight...

packages/wix-ui-core/src/components/CircularProgressBar/CircularProgressBar.tsx
lbelinsk: Better to put it in the stylesheet instead of inline

packages/wix-ui-core/src/components/CircularProgressBar/CircularProgressBar.tsx
```diff
@@ -0,0 +1,103 @@
...
+  const {error, showProgressIndication} = props;
+  const _props = normalizeProps(props);
+  const success = _props.value === FULL_PROGRESS;
+  const value = error && _props.errorLabel ? _props.errorLabel : `${Math.floor(_props.value)}%`;
+
+  return (
+    <div {...style('root', {error, success}, _props)}>
+      {renderArcs(_props)}
+      {showProgressIndication &&
+        <div data-hook="progress-indicator" className={style.progressIndicator}>
```
lbelinsk: I think we don't need to have 3 layers here. why not just one div / span?

packages/wix-ui-core/src/components/CircularProgressBar/CircularProgressBar.protractor.driver.ts
alisey: Another option is to have a test for function that generates the arc path for the component, and then that known-to-work-correctly function can be used in other tests where you need to check that the value was passed correctly all the way through.

packages/wix-ui-core/src/components/CircularProgressBar/CircularProgressBar.tsx
aaronvine: agree about inline, but we still need it for the arc svg viewbox

aaronvine: agree about inline, but we still need it for the arc svg viewbox

alisey: Right, but shouldn't it be a prop then? Something like: `<CircularProgressBar size={54} />`

packages/wix-ui-core/src/components/CircularProgressBar/CircularProgressBar.tsx
```diff
@@ -0,0 +1,111 @@
...
+  /** circle size in pixels */
+  size?: number | string;
+}
+
+const FULL_PROGRESS = 100;
+const NO_PROGRESS = 0;
+
+const VIEWBOX = 54;
+
+const ANGLE_NORMALIZE_VAL = 0.1;
```
alisey: Instead of doing these conversions between percents and angles, couldn't you just pass a value in the range [0..1] to the Arc component, where 0 means empty, and 1 means full? And then do a single multiplication there.

So instead of `<Arc angle={FULL_PROGRESS_ANGLE} />` you'd have `<Arc value={1} />`

Just a suggestion. I think it would simplify the code a bit, but up to you.

packages/wix-ui-core/src/components/CircularProgressBar/Arc.tsx
```diff
@@ -0,0 +1,71 @@
...
+}
+
+const describeArc = (x, y, radius, startAngle, endAngle) => {
+  const start = polarToCartesian(x, y, radius, endAngle);
+  const end = polarToCartesian(x, y, radius, startAngle);
+
+  const largeArcFlag = endAngle - startAngle <= 180 ? '0' : '1';
+
+  const d = [
+    'M', start.x, start.y,
```
alisey: Here you could do 'start.x + 0.1' and then you wouldn't need `ANGLE_NORMALIZE_VAL`.

packages/wix-ui-core/src/components/CircularProgressBar/CircularProgressBar.tsx
```diff
@@ -0,0 +1,113 @@
+import * as React from 'react';
+import {bool, element, oneOfType, number, string} from 'prop-types';
+import style from './CircularProgressBar.st.css';
+import Arc from './Arc';
+
+export interface CircularProgressBarProps {
+  /** represent the progress state in percentages (0 - no progress, 100 - progress completed) */
+  value?: number | string;
```
alisey: I see no reason why we should accept strings here.

packages/wix-ui-core/src/components/CircularProgressBar/Arc.tsx
```diff
@@ -0,0 +1,67 @@
...
+    'A', radius, radius, 0, largeArcFlag, 0, end.x, end.y
+  ].join(' ');
+
+  return d;
+}
+
+export interface ArcProps {
+  angle: number;
+  value?: number;
```
alisey: Oh, it already accepts value. Then it certainly should not accept angle.

packages/wix-ui-core/src/components/CircularProgressBar/CircularProgressBar.tsx
aaronvine: we were instructed to handle it in LinearProgressBar so it's pretty much the same here

packages/wix-ui-core/src/components/CircularProgressBar/CircularProgressBar.spec.tsx
aaronvine: added

packages/wix-ui-core/src/components/CircularProgressBar/CircularProgressBar.tsx
aaronvine: yeah it's redundant

packages/wix-ui-core/src/components/CircularProgressBar/Arc.tsx
```diff
@@ -0,0 +1,67 @@
...
+  const viewBox = `-${viewBoxSize / 2} -${viewBoxSize / 2} ${viewBoxSize} ${viewBoxSize}`;
+
+  return (
+    <svg
+      xmlns="http://www.w3.org/2000/svg"
+      viewBox={viewBox}
+      className={className}
+      data-hook={value !== undefined ? 'progressarc-foreground' : 'progressarc-background'}
+    >
+      {value !== undefined && <title>{`Arc percentage ${value}%`}</title>}
```
alisey: Title should be an attribute, as in ``, and it should be on the `<svg>` element. Also I think it should simply say `${value}%`, because that's what it represents, just some percentage.

packages/wix-ui-core/src/components/CircularProgressBar/Arc.tsx
aaronvine: it's pretty much the same component and we can use it later in loader, just added some typescript wrapping

packages/wix-ui-core/src/components/CircularProgressBar/Arc.tsx
alisey: Oh, you're totally right, I didn't know that SVG supports title tag.

aaronvine: i'll try integrating the simplified arc one, looks cool

alisey: One thing I'm not sure about: if we plan to reuse the `Arc` component for the `Loader`, then it would probably need to accept two values, for the start angle and for the end angle. But we can sort it out later.

packages/wix-ui-core/src/components/CircularProgressBar/Arc.tsx
```diff
@@ -0,0 +1,48 @@
...
+  value /= 100;
+  const angle = 2 * Math.PI * (value - 1/4);
+  const x = r * Math.cos(angle);
+  const y = r * Math.sin(angle);
+  const sweep = Math.round(value);
+  return `M .01 ${-r} A ${r} ${r} 0 ${sweep} 1 ${x} ${y}`;
+}
+
+export interface ArcProps {
+  value: number;
```
lbelinsk: If there is no `?` then in the proptype validation put `.required`

packages/wix-ui-core/src/components/CircularProgressBar/Arc.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+  return (
+    <svg
+      className={className}
+      width={size}
+      height={size}
+      viewBox={viewBox}
+      data-hook={title !== undefined ? 'progressarc-foreground' : 'progressarc-background'}
+    >
+      {title && <title>{title}</title>}
```
lbelinsk: I think it will be better to remove the `title`. We can depend on eyes. Let's render the component in the e2e in different values, and that should be good enough.

packages/wix-ui-core/src/components/CircularProgressBar/CircularProgressBar.spec.tsx
lbelinsk: have you tried removing it?

packages/wix-ui-core/src/components/CircularProgressBar/CircularProgressBar.tsx
```diff
@@ -0,0 +1,105 @@
...
+
+const renderArcs = (props: CircularProgressBarProps) => {
+  const { value, size } = props;
+  const normalizedSize = normalizeSize(size);
+  const normalizedValue = normalizeValue(value);
+  return (
+    <div className={style.arcsContainer}>
+      {resolveIndicationElement(props)}
+      <Arc value={FULL_PROGRESS} className={style.backArc} strokeWidth={4} size={normalizedSize} />
+      <Arc value={normalizedValue} title={`${normalizedValue}%`} className={style.foreArc} strokeWidth={4} size={normalizedSize} />
```
lbelinsk: you have title here. don't forget it

packages/wix-ui-core/src/components/CircularProgressBar/CircularProgressBar.tsx
lbelinsk: ?

PR #585, opened 2018-06-19
fix(createHOC): use class or function name when displayName is unavailable

argshook:
1. some tests to cover displayName functionality of `createHOC`
1. use `Component.displayName || Component.name || 'WixComponent'`

this fixes displayName for components that don't explicitly set it.
However, for components that are anonymous functions, it's gonna
fallback to `'WixComponent'` which may or may not be desired behaviour.


shlomitc:
why not just force to always have `displayName`?

argshook:
we can, if there's a way to do it (lint rule?) but this additional `||` anyway does more good than harm

PR #587, opened 2018-06-20
RadioButton - add focus-visible state

JoMarton:


shlomitc:
@hepiyellow - any comments regarding this `focus-visible` feature?

packages/wix-ui-core/src/components/RadioButton/RadioButton.driver.ts
```diff
@@ -19,6 +20,7 @@ export const radioButtonDriverFactory = ({element, eventTrigger}) => {
     labelExists: () => !!getLabel(),
     isChecked: () => domUtils.hasStyleState(element, 'checked'),
     isFocused: () => domUtils.hasStyleState(element, 'focused'),
-    isDisabled: () => domUtils.hasStyleState(element, 'disabled')
+    isDisabled: () => domUtils.hasStyleState(element, 'disabled'),
+    isFocuseVisible: () => domUtils.hasStyleState(element, 'focus-visible')
```
alisey: Typo, should be `isFocusVisible`

packages/wix-ui-core/src/components/RadioButton/RadioButton.spec.tsx
```diff
@@ -87,4 +87,20 @@ describe('RadioButton', () => {
 
     expect(radio.iconExists()).toBeTruthy();
   });
+
+  it('get NON VISIBLE focus if clicked', () => {
```
alisey: get -> gets

packages/wix-ui-core/src/components/RadioButton/RadioButton.tsx
```diff
@@ -105,12 +115,17 @@ export class RadioButton extends React.Component<RadioButtonProps, RadioButtonSt
...
   }
 
   onInputBlur = () => {
-    this.setState({focused: false});
+    this.setState({focused: false, focusVisible: false});
+    this.focusedByMouse = false;
   }
 
+  handleInputKeyDown = () => {
+    this.setState({focusVisible: true})
```
alisey: Semicolon

packages/wix-ui-core/src/components/RadioButton/RadioButton.driver.ts
JoMarton: good catch! :)

packages/wix-ui-core/src/components/RadioButton/RadioButton.tsx
```diff
@@ -91,6 +96,11 @@ export class RadioButton extends React.Component<RadioButtonProps, RadioButtonSt
     );
   }
 
+  handleClick = (event: React.MouseEvent<HTMLDivElement>) => {
```
alisey: I'm not sure this is going to work as expected, because `click` occurs after `focus`. In ToggleSwitch we're using mouseDown event: https://github.com/wix/wix-ui/blob/master/packages/wix-ui-core/src/components/ToggleSwitch/ToggleSwitch.tsx#L114

alisey: Just checked in the storybook, `focusedByMouse = true` should be set in `handleMouseDown` instead, `handleInputChange` should stay in `handleClick`.

JoMarton: This case is a bit different, onClick happens on the root and only then focus is called on the native input child...

alisey: ah, ok, but I noticed in the storybook that, unlike in native controls and ToggleSwitch, `focus-visible` is removed if you first focus by keyboard, and then click. I can live with that.

JoMarton:
@shlomitc  @hepiyellow 
I can haz merge?

hepiyellow:
This implementation has holes in it, please see FocusableHOC in WSR for reference.
Only Issue I see now is quite esoteric:
- in case a TAB is made from the browser's address-bar and focuses on a RadioButton, then you won't get the focus-visible, since you don'y get the keyDown of the TAB, you rather get the keyUp.

alisey:
@hepiyellow is `FocusableHOC` compatible with wix-ui-core? If it is, then let's merge this PR, and as a separate task move FocusableHOC from wix-style-react to wix-ui-core, and update this file to use it.

I don't think that esoteric edge cases should be a blocker for this PR. There's plenty of more pressing issues that need to be fixed in order to launch StarRating.

alisey:
Oh, cool, didn't notice you accepted.

PR #596, opened 2018-06-22
Improve TimePicker component

alexgoncharwix:
Add more functions to TimePicker.driver and TimePicker.protractor.driver.
Add 'width?: string' to Input component.


alisey:
> Add 'width?: string' to Input component.

Why not add inline style support instead? It's standard, it should be supported anyway, and it provides more flexibility.

argshook:
@alisey yeah why not, that's a good idea also, i'll brush it up in another PR :)

PR #597, opened 2018-06-22
WIP: <Badge/> - add max-width support

Faradey27:
It is a proposal pull request to resolve issues: https://github.com/wix-private/wsr-issues/issues/134 and https://github.com/wix/wix-style-react/issues/1622

My points:
- Add support of max-width and badge+tooltip to core
- style tooltip in wix-ui-backoffice
- done

@shlomitc let discuss this pull request and after some contract, I will improve pull and add tests

### DIFF
--- a/packages/wix-ui-core/src/components/Label/Label.tsx
+++ b/packages/wix-ui-core/src/components/Label/Label.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import {string, bool, node} from 'prop-types';
+import {string, bool, oneOfType, node, number} from 'prop-types';
 import style from './Label.st.css';
 
 export interface LabelProps {
@@ -14,6 +14,8 @@ export interface LabelProps {
   ellipsis?: boolean;
   /** Is the Label disabled */
   disabled?: boolean;
+  /** Label max width */
+  maxWidth?: number | string;
 }
 
 const defaultProps: LabelProps = {
@@ -25,7 +27,16 @@ const defaultProps: LabelProps = {
  */
 export const Label: React.SFC<LabelProps> = props => {
   const {id, children, ellipsis, disabled} = props;
-  return <label {...style('root', {ellipsis, disabled}, props)} htmlFor={props.for} id={id}>{children}</label>;
+  return (
+    <label
+      {...style('root', {ellipsis, disabled}, props)}
+      htmlFor={props.for}
+      id={id}
+      style={{maxWidth: props.maxWidth}}
+    >
+      {children}
+    </label>
+  );
 };
 
 Label.displayName = 'Label';
@@ -42,7 +53,9 @@ Label.propTypes = {
   /** should the text be ellipsed or not */
   ellipsis: bool,
   /** Is the Label disabled */
-  disabled: bool
+  disabled: bool,
+  /** Label max width */
+  maxWidth: oneOfType([number, string]),
 };
 
 Label.defaultProps = defaultProps;
--- a/packages/wix-ui-core/src/components/StylableBadge/Badge.tsx
+++ b/packages/wix-ui-core/src/components/StylableBadge/Badge.tsx
@@ -1,26 +1,92 @@
 import * as React from 'react';
-import {object, any} from 'prop-types';
+import * as ReactDOM from 'react-dom';
+import {any, oneOfType, number, string} from 'prop-types';
+/*
+  To resolve issue, we need to import Requireable
+  error TS4026: Public static property 'propTypes' of exported class has or is using name 'Requireable'
+  from external module
+  "/Users/andriit/Documents/Projects/wix-ui/packages/wix-ui-core/node_modules/@types/prop-types/index"
+  but cannot be named.
+*/
+import { Requireable } from 'prop-types';
+import {Label} from './../Label';
+import {Tooltip} from './../Tooltip';
+import {isShallowEqual} from './../../utils/isShallowEqual.tsx';
 import style from './Badge.st.css';
 
 export interface BadgeProps {
   children?: React.ReactNode;
   className?: string;
+  maxWidth?: number | string;
+}
+
+export interface BadgeState {
+  isEllipsisActive: boolean;
 }
 
 /**
  * Badge
  */
-export const Badge: React.SFC<BadgeProps> = props => (
-  <span {...style('root', {}, props)}>
-    {props.children}
-  </span>
-);
+export class Badge extends React.Component<BadgeProps, BadgeState> {
+  static displayName = 'Badge';
+  static propTypes = {
+    /** Any node to be rendered (usually text node) */
+    children: any,
 
-Badge.displayName = 'Badge';
-Badge.propTypes = {
-  /** Any node to be rendered (usually text node) */
-  children: any,
+    /** className to place on the root of the rendered element */
+    className: any,
 
-  /** className to place on the root of the rendered element */
-  className: any
+    /** When a maximum width is set and the badge content did not fit, a tooltip will become visible */
+    maxWidth: oneOfType([number, string]),
+  }
+
+  state = {
+    isEllipsisActive: false
+  }
+
+  componentDidMount() {
+    this.setState({isEllipsisActive: this.isEllipsisActive()});
+  }
+
+  componentDidUpdate(prevProps) {
+    // if props changed, then we want to re-check node for ellipsis state
+    // and we can not do such check in render, because we want to check already rendered node
+    if (!isShallowEqual(prevProps, this.props)) {
+      this.setState({isEllipsisActive: this.isEllipsisActive()});
+    }
+  }
+
+  isEllipsisActive = () => {
+    const node = ReactDOM.findDOMNode(this) as HTMLElement;
+    if (!node) {
+      return false;
+    }
+    return node.offsetWidth < node.scrollWidth;
+  }
+
+  render() {
+    if (!this.state.isEllipsisActive) {
+      return (
+        <Label
+          {...style('root', {}, this.props)}
+          ellipsis
+          maxWidth={this.props.maxWidth}
+        >
+          {this.props.children}
+        </Label>
+      );
+    }
+
+    return (
+      <Tooltip content={this.props.children}>
+        <Label
+          {...style('root', {}, this.props)}
+          ellipsis
+          maxWidth={this.props.maxWidth}
+        >
+          {this.props.children}
+        </Label>
+      </Tooltip>
+    )
+  }
 };
--- a/packages/wix-ui-core/src/components/StylableBadge/Badge.tsx
+++ b/packages/wix-ui-core/src/components/StylableBadge/Badge.tsx
@@ -41,8 +41,6 @@ export class Badge extends React.Component<BadgeProps, BadgeState> {
     maxWidth: oneOfType([number, string]),
   }
 
-  private badgeNode: Node;
-
   state = {
     isEllipsisActive: false
   }
@@ -59,16 +57,24 @@ export class Badge extends React.Component<BadgeProps, BadgeState> {
     }
   }
 
-  setBadgeNode = node => this.badgeNode = node;
-
-  updateEllipsesState = () => this.setState({isEllipsisActive: isEllipsisActive(this.badgeNode)});
+  updateEllipsesState = () => this.setState({isEllipsisActive: isEllipsisActive(ReactDOM.findDOMNode(this) as HTMLElement)});
 
   render() {
-    return (
-      <Tooltip content={this.props.children} visible={this.state.isEllipsisActive}>
+    if (!this.state.isEllipsisActive) {
+      return (
         <span
           {...style('root', {}, this.props)}
-          ref={this.setBadgeNode}
+          style={{ maxWidth: this.props.maxWidth }}
+        >
+          {this.props.children}
+        </span>
+      );
+    }
+
+    return (
+      <Tooltip content={this.props.children}>
+        <span
+          {...style('root', {}, this.props)}
           style={{maxWidth: this.props.maxWidth}}
         >
           {this.props.children}
--- a/packages/wix-ui-core/src/components/StylableBadge/Badge.tsx
+++ b/packages/wix-ui-core/src/components/StylableBadge/Badge.tsx
@@ -41,6 +41,8 @@ export class Badge extends React.Component<BadgeProps, BadgeState> {
     maxWidth: oneOfType([number, string]),
   }
 
+  private badgeNode: Node;
+
   state = {
     isEllipsisActive: false
   }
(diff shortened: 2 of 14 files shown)


packages/wix-ui-core/src/components/StylableBadge/Badge.tsx
```diff
@@ -1,26 +1,92 @@
...
+  }
+
+  componentDidMount() {
+    this.setState({isEllipsisActive: this.isEllipsisActive()});
+  }
+
+  componentDidUpdate(prevProps) {
+    // if props changed, then we want to re-check node for ellipsis state
+    // and we can not do such check in render, because we want to check already rendered node
+    if (!isShallowEqual(prevProps, this.props)) {
```
lbelinsk: Maybe if you extend `React.PureComponent` instead of `React.Component` you will get the shallow comparison out of the box.

packages/wix-ui-core/src/components/StylableBadge/Badge.tsx
```diff
@@ -1,26 +1,92 @@
...
+      this.setState({isEllipsisActive: this.isEllipsisActive()});
+    }
+  }
+
+  isEllipsisActive = () => {
+    const node = ReactDOM.findDOMNode(this) as HTMLElement;
+    if (!node) {
+      return false;
+    }
+    return node.offsetWidth < node.scrollWidth;
```
lbelinsk: Can prevent the `if` here:
```
return node && node.offsetWidth < node.scrollWidth;
```

lbelinsk: I think that this ability to know when to display the ellipsis or not is very useful and should not sit inside the Badge. For example, breadcrumbs, ProgressBar text, ... It can be used internally and outside of the lib for the end users.

packages/wix-ui-core/src/components/StylableBadge/Badge.tsx
```diff
@@ -1,26 +1,92 @@
...
+        <Label
+          {...style('root', {}, this.props)}
+          ellipsis
+          maxWidth={this.props.maxWidth}
+        >
+          {this.props.children}
+        </Label>
+      </Tooltip>
+    )
+  }
```
lbelinsk: Missing tests. You can choose e2e / component testing.

packages/wix-ui-core/src/components/StylableBadge/Badge.tsx
```diff
@@ -1,26 +1,92 @@
...
+    if (!node) {
+      return false;
+    }
+    return node.offsetWidth < node.scrollWidth;
+  }
+
+  render() {
+    if (!this.state.isEllipsisActive) {
+      return (
+        <Label
```
lbelinsk: Why are we using Label here?
Maybe Text or just a span?

packages/wix-ui-core/src/components/StylableBadge/Badge.tsx
```diff
@@ -1,26 +1,92 @@
...
+          ellipsis
+          maxWidth={this.props.maxWidth}
+        >
+          {this.props.children}
+        </Label>
+      );
+    }
+
+    return (
+      <Tooltip content={this.props.children}>
```
lbelinsk: I have the exact same usage now in my ProgressBar, where I want to conditionally render a tooltip that will wrap the root. I would really like us to have a prop for the tooltip like:

```
<Tooltip content={this.props.children} visible={false}>
```

And then the condition will happen inside of it, hidden from us.

Let's talk about it in the weekly.

packages/wix-ui-core/src/utils/isShallowEqual.tsx
```diff
@@ -0,0 +1,15 @@
+export const isShallowEqual = (v, o) => {
```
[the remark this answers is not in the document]
Faradey27: because we already had such function in a code)

will update to lodash

packages/wix-ui-core/src/components/StylableBadge/Badge.tsx
```diff
@@ -1,26 +1,92 @@
...
-  /** className to place on the root of the rendered element */
-  className: any
+export class Badge extends React.Component<BadgeProps, BadgeState> {
+  static displayName = 'Badge';
+  static propTypes = {
+    /** Any node to be rendered (usually text node) */
+    children: any,
+
+    /** className to place on the root of the rendered element */
+    className: any,
```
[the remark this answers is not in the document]
Faradey27: +

packages/wix-ui-core/src/components/Label/Label.tsx
```diff
@@ -42,7 +53,9 @@ Label.propTypes = {
   /** should the text be ellipsed or not */
   ellipsis: bool,
   /** Is the Label disabled */
-  disabled: bool
+  disabled: bool,
+  /** Label max width */
+  maxWidth: oneOfType([number, string]),
```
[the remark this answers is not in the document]
Faradey27: why? I remember that in a lot of components we have such `number or string` condition

also, number is completely valid value, same as 50px, but '50' not valid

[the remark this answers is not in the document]
Faradey27: @lbelinsk

packages/wix-ui-core/src/utils/isShallowEqual.tsx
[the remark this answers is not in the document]
Faradey27: @lbelinsk lodash does not have such method https://github.com/lodash/lodash/issues/2340

they have deep equal

packages/wix-ui-core/src/components/Label/Label.tsx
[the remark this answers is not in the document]
lbelinsk: Since it is going to be a css property, it is legal to pass numbers as well, so I guess it is ok

packages/wix-ui-core/src/utils/isShallowEqual.tsx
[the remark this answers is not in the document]
lbelinsk: right. ok, so let's just have a test to it

packages/wix-ui-core/src/components/StylableBadge/Badge.tsx
lbelinsk: About style duplications, we intend to have some useful mixins to apply typography and stuff. this is a great place to use them

packages/wix-ui-core/src/components/StylableBadge/Badge.tsx
lbelinsk: Feels like it can be a nice HOC. Take a badge, wrap it with HOC, and then it will have the behavior of conditional ellipsis.

`withEllipsis(Badge)`

packages/wix-ui-core/src/components/StylableBadge/Badge.tsx
lbelinsk: hmmm, ok, not sure i have a better idea at the moment

packages/wix-ui-core/src/components/StylableBadge/Badge.tsx
lbelinsk: I updated it in the issue: 
https://github.com/wix-private/wsr-issues/issues/130

packages/wix-ui-core/src/components/Label/Label.tsx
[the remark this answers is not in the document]
Faradey27: added

packages/wix-ui-core/src/components/StylableBadge/Badge.tsx
Faradey27: added

packages/wix-ui-core/src/components/Tooltip/Tooltip.tsx
```diff
@@ -128,7 +131,7 @@ export class TooltipComponent extends React.PureComponent<TooltipProps & Injecte
       <Popover
         {...style('root', {}, this.props)}
         placement={placement}
-        shown={isOpen}
+        shown={this.props.visible && isOpen}
```
lbelinsk: it is not that simple. The tooltip can be opened programatically with his static member function `open` - https://github.com/wix/wix-ui/blob/db90948a9c23a8f6e3ee4dc7205e9d564cfda0df/packages/wix-ui-core/src/components/Tooltip/Tooltip.tsx#L112

What if we do `ref.open()` and the prop is false?
We need to think about it

lbelinsk: Anyhow, i would take care of the tooltip in a separate pr

packages/wix-ui-core/src/utils/isShallowEqual.tsx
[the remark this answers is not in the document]
alisey: It's an inefficient implementation, it needlessly iterates twice and doesn't have an early return for objects of different lengths.
Here's a better one: https://github.com/wix/stylable-components/blob/751462afea5430fbc98896c1b1e5129eb36dfd4b/src/utils/objects-shallow-equal.ts

packages/wix-ui-core/src/components/Tooltip/Tooltip.tsx
Faradey27: ok, it is just a proposal)

with my way, by default everything by default will left same, as visible: true, by default.
If someone will set visible: false and then call `open`, from my perspective nothing should happend

@lbelinsk

packages/wix-ui-core/src/components/Label/Label.tsx
```diff
@@ -14,6 +14,8 @@ export interface LabelProps {
   ellipsis?: boolean;
   /** Is the Label disabled */
   disabled?: boolean;
+  /** Label max width */
+  maxWidth?: number | string;
```
alisey: What's the advantage of this over `style` props? Are we going to have props for minWidth, maxWidth, width, minHeight, maxHeight, height, etc?

packages/wix-ui-core/src/components/StylableBadge/Badge.tsx
```diff
@@ -1,26 +1,79 @@
...
+
+  componentDidUpdate(prevProps) {
+    // if props changed, then we want to re-check node for ellipsis state
+    // and we can not do such check in render, because we want to check already rendered node
+    if (!isShallowEqual(prevProps, this.props)) {
+      this.updateEllipsesState();
+    }
+  }
+
+  setBadgeNode = node => this.badgeNode = node;
```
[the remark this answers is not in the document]
shlomitc: we need to support React 15

packages/wix-ui-core/src/components/StylableBadge/Badge.tsx
```diff
@@ -1,26 +1,79 @@
...
+
+  updateEllipsesState = () => this.setState({isEllipsisActive: isEllipsisActive(this.badgeNode)});
+
+  render() {
+    return (
+      <Tooltip content={this.props.children} visible={this.state.isEllipsisActive}>
```
shlomitc: I would suggest not rendering a tooltip at all it such case instead of controlling the visibility of it

packages/wix-ui-core/src/utils/isShallowEqual.tsx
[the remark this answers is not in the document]
Faradey27: @alisey why your algo is better?

in your algo we have:
const aKeys = Object.keys(a) - O(n)
const bKeys = Object.keys(b); - O(n)
return (aKeys.length === bKeys.length) ? aKeys.every(k => a[k] === b[k]) : false; - O(n)

3 loops

in my just 2 loops

packages/wix-ui-core/src/components/Label/Label.tsx
Faradey27: @alisey because we are using approach where we not allow user to modify everything, we just want to give him minimal ability of customization

1) lib dev could override everything with styleble
2) user could modify just things that was allowed to modify

if we want to have another approach, better to talk with @shlomitc

packages/wix-ui-core/src/utils/isEllipsisActive.tsx
```diff
@@ -0,0 +1 @@
+export const isEllipsisActive = (node: HTMLElement) => node && node.offsetWidth < node.scrollWidth;
```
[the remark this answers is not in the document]
Faradey27: @alisey they name is fine, this function is always true when elipses is active and false when elipses is not active.

@shlomitc @lbelinsk

packages/wix-ui-core/src/components/StylableBadge/Badge.tsx
Faradey27: yea, it was previous way, will return

packages/wix-ui-core/src/utils/isEllipsisActive.tsx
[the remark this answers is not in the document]
lbelinsk: I think there is no problem with any suggested name:
`hasOverflowX`, `hasHorizontalOverflow`, `isEllipsisActive`, `shouldApplyEllipsis`...

packages/wix-ui-core/src/components/StylableBadge/Badge.tsx
lbelinsk: wait, why?
I expect the tooltip to hide this inside of it. So the Tooltip can render only the children if it got false prop. no css hiding..

packages/wix-ui-core/src/utils/isEllipsisActive.tsx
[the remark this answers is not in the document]
lbelinsk: Our intention is to use it for the ellipsis, but yeah, in general, I agree with the name you offer since it is more generic and describes better what it is actually doing.

packages/wix-ui-core/src/components/Label/Label.tsx
lbelinsk: wix-ui-core should be as flexible as possible, that is the plan, nothing changed.

packages/wix-ui-core/src/utils/isShallowEqual.tsx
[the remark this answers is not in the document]
Faradey27: @alisey no, i just created this test https://jsfiddle.net/to4jhsyy/6/ and my function is 2 times faster then yours) - http://prntscr.com/jzz5lb

Maybe in your test you worked with same objects or with small objects but large amount of times and webkit has applied some JIT optimisations

packages/wix-ui-core/src/components/Label/Label.tsx
Faradey27: @alisey ok, so should we everywhere use style obj? instead of all custom properties? @shlomitc @lbelinsk

packages/wix-ui-core/src/utils/isShallowEqual.tsx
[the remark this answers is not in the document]
alisey: Unlikely to do anything with JIT, because running in loop gives the same results. Depends on the data, I guess. I take back my words about inefficient.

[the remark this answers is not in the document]
alisey: For the sake of completeness [here's React's implementation](https://github.com/facebook/react/blob/72434a7686035b4af766ee7d06c070d7f5d6a5f2/packages/shared/shallowEqual.js).

packages/wix-ui-core/src/utils/isEllipsisActive.tsx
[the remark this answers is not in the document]
alisey: Maybe it shouldn't be exposed as a general utility then. Checking if there's *really* an ellipsis could be messy and unnecessary for your use case.

Personally, I wouldn't even add this as a private utility, and would just inline the comparison.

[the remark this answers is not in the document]
Faradey27: @alisey at start it was, but as i and @lbelinsk need same logic, we decided to move it to utils

Faradey27:
@lbelinsk @alisey @shlomitc I am closing pull request as issue taking to much time for discussing without any progress

alisey:
@Faradey27 treat everything I say as suggestions, I'm not requesting any changes actually.

PR #598, opened 2018-06-24
`<Popover/>` - Fix Popover.spec to use Popover.driver

hepiyellow:
As part of mocha runner refactoring, @alisey rewrote the Popover specs, and removed the usage of the Popover.driver. 
I fixed it so it uses the driver, so that our driver is verified.
- moved the `getArrowOffest` driver method to a private driver (I didn't see anyone using it in core or backoffice, and I don't think consumers need it).

### DIFF
--- a/packages/wix-ui-core/src/common/BaseDriver.ts
+++ b/packages/wix-ui-core/src/common/BaseDriver.ts
@@ -1,0 +1,21 @@
+import { Simulate } from 'react-dom/test-utils';
+
+export interface DriverArgs {
+  element: HTMLElement;
+  eventTrigger: typeof Simulate;
+}
+
+export class BaseDriver {
+  readonly element : HTMLElement;
+  readonly eventTrigger: typeof Simulate;
+
+  constructor({element, eventTrigger}: DriverArgs) {
+    this.element = element;
+    this.eventTrigger = eventTrigger || Simulate;
+  }
+
+  exists = () => !!this.element;
+  mouseEnter = () =>  this.eventTrigger.mouseEnter(this.element);
+  mouseLeave = () =>  this.eventTrigger.mouseLeave(this.element);
+  click = () => this.eventTrigger.click(this.element);
+}
--- a/packages/wix-ui-core/src/common/BaseDriver.ts
+++ b/packages/wix-ui-core/src/common/BaseDriver.ts
@@ -2,7 +2,7 @@ import { Simulate } from 'react-dom/test-utils';
 
 export interface DriverArgs {
   element: HTMLElement;
-  eventTrigger: typeof Simulate;
+  eventTrigger?: typeof Simulate;
 }
 
 export class BaseDriver {
--- a/packages/wix-ui-core/src/common/BaseDriver.ts
+++ b/packages/wix-ui-core/src/common/BaseDriver.ts
@@ -1,21 +1,21 @@
 import { Simulate } from 'react-dom/test-utils';
 
-export interface DriverArgs {
+export interface DriverBuilderParams {
   element: HTMLElement;
   eventTrigger?: typeof Simulate;
 }
 
+/**
+ * Base class for public drivers (exposed to consumers)
+ */
 export class BaseDriver {
-  readonly element : HTMLElement;
-  readonly eventTrigger: typeof Simulate;
+  protected readonly element : HTMLElement;
+  protected readonly eventTrigger: typeof Simulate;
 
-  constructor({element, eventTrigger}: DriverArgs) {
+  constructor({element, eventTrigger}: DriverBuilderParams) {
     this.element = element;
     this.eventTrigger = eventTrigger || Simulate;
   }
 
   exists = () => !!this.element;
-  mouseEnter = () =>  this.eventTrigger.mouseEnter(this.element);
-  mouseLeave = () =>  this.eventTrigger.mouseLeave(this.element);
-  click = () => this.eventTrigger.click(this.element);
 }
--- a/packages/wix-ui-core/src/components/Popover/Popover.driver.private.ts
+++ b/packages/wix-ui-core/src/components/Popover/Popover.driver.private.ts
@@ -1,20 +1,30 @@
+import { Simulate } from 'react-dom/test-utils';
 const queryDocumentOrElement = (element, query) => ((element && element.querySelectorAll(query)[0]) || document && document.querySelector(query));
 const getTargetElement = (element: Element | undefined) => element && element.querySelectorAll('[data-hook="popover-element"]')[0];
 const getContentElement = (element: Element | undefined) => queryDocumentOrElement(element, '[data-hook="popover-content"]');
 const getArrowElement = (element: Element | undefined) => queryDocumentOrElement(element, '[data-hook="popover-arrow"]');
 const getPortalElement = (element: Element | undefined) => queryDocumentOrElement(element, '[data-hook="popover-portal"]');
 
-export const popoverPrivateDriverFactory = ({element, eventTrigger}) => ({
-  getTargetElement: () => getTargetElement(element),
-  getContentElement: () => getContentElement(element),
-  isTargetElementExists: () => !!getTargetElement(element),
-  isContentElementExists: () => !!getContentElement(element),
-  mouseEnter: () => eventTrigger.mouseEnter(element),
-  mouseLeave: () => eventTrigger.mouseLeave(element),
-  inlineStyles: () => element.style,
-  getArrowOffset: () => {
-    const {top, left, right, bottom} = (getArrowElement(element) as HTMLElement).style;
+export class PopoverDriverPrivate {
+  readonly element:HTMLElement;
+  readonly eventTrigger: typeof Simulate;
+
+  constructor({element, eventTrigger}) {
+    this.element = element;
+    this.eventTrigger = eventTrigger || Simulate;
+  }
+
+  getTargetElement = () =>  getTargetElement(this.element);
+  getContentElement = () =>  getContentElement(this.element);
+  isTargetElementExists = () =>  !!getTargetElement(this.element);
+  isContentElementExists = () =>  !!getContentElement(this.element);
+  mouseEnter = () =>  this.eventTrigger.mouseEnter(this.element);
+  mouseLeave = () =>  this.eventTrigger.mouseLeave(this.element);
+  inlineStyles = () =>  this.element.style;
+  getArrowOffset = () => {
+    const {top, left, right, bottom} = (getArrowElement(this.element) as HTMLElement).style;
     return {top, left, right, bottom};
-  },
-  getPortalElement: () => getPortalElement(element),
-});
+  }
+  getPortalElement = () =>  getPortalElement(this.element);
+
+};
--- a/packages/wix-ui-core/src/components/Popover/Popover.driver.private.ts
+++ b/packages/wix-ui-core/src/components/Popover/Popover.driver.private.ts
@@ -1,25 +1,21 @@
-import { Simulate } from 'react-dom/test-utils';
+import { BaseDriver , DriverArgs} from './../../common/BaseDriver';
+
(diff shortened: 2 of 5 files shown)


alisey:
1. I think we agreed that drivers are for consumers, they shouldn't be used for unit tests, and should only contain methods relevant to the consumers. To test a public driver you can define a new `describe()` specifically for it. Seeing two different drivers mixed together is confusing.

2. `popoverPrivateDriverFactory` - it's not a factory, in OOP terms this function is analogous to `new PopoverPrivateDriver(..., ...)`. It's just an object constructor, no reason to make things more complicated.

3. `isTargetElementExists`. So the component has two parts: `Popover.Element` and `Popover.Content`, none of which are called target. From reading tests it's not clear what target is. We shouldn't introduce new terminology not used in the component. If the component's terminology is confusing let's change that first.

4. `getArrowOffset` - checking style doesn't guarantee that the arrow is offset by specified amount, because there are many other things that could affect the arrow's position. Ideally this functionality should either be checked using screenshot tests, or `.getBoundingClientRect()` to get the computed position.

hepiyellow:
@alisey 
1.public driver testing -  I think having a separate test suite for the public drivers in unnecessary overhead. We believe we should use the public drivers in our component test. Sometime I add a few missing cases to a component test, specifically to cover the public driver.

2. "Factory" - I totally agree. I do it only to be consistent. I would like to have our drivers as actual ES6 classes. In most cases we could use class inheritence for the drivers, but in generic cases (like the FocusableHOC) we'd need to use mixins, and that brings a TS issue. So I am still waiting for a solution. When I find one, I will rename all the drivers.

3. You are right. But this is out of the scope of this PR.
 I think "Target" is more clear than just "Element". So I guess we could either rename `Popover.Element` (and fix all places that use it), OR add `Popover.Target` and "softly" deprecate `Popover.Element`, OR we can simply add documentation about `Popover.Element` and say it is the "Target Element".

4. You are right, But this is out of the scope of this PR.
We should add an e2e test with eyes.

alisey:
1. Ok, up to you. But I think it's a horrible idea. Like if you could compare tradeoffs of different approaches, and then combine the worst parts of each, this would be it.

> I think having a separate test suite for the public drivers in unnecessary overhead

It's just another describe in the test file, and most public drivers should have 1 or 2 methods. So we're talking 2 extra unit tests on average.

2. Pagination and Input already use E6 classes for their drivers. Since we're separating public and private drivers you wouldn't violate consistency by using a class for Popover too.

3. Imo, target is just as unclear as element. TriggerElement, or maybe Reference - as they call it in Popper? Anyway, if renaming element to target in the component is out of scope, then renaming it in the tests should also be out of scope.

hepiyellow:
1. I will talk to Shlomi about it. 
  - Separate test for drivers could have some value. I noticed that sometimes people forget to test a driver method (thoroughly), and it turns out broken. A dedicated test suit for the driver might have people more oriented into testing the driver (not the component) and they might not forget stuff.
1.a - Still I think that private and public driver usually do have common stuff, like the selectors of the component's sub-elements. Do you suggest to duplicate the selectors in both drivers? I once tried to use 1 driver with the private methods "hidden" inside a INTERNAL `Symbol` property.

2. I don't see any `class` drivers in `wix-ui-core` Pagination or Input. Can you point me to it?

3. "Target" - react-pooper used to call it `Target` in their API version 0.x (which we are using). In version 1.x they renamed it to `Reference`. Anyway, the test should be consistent with the component. When we rename it, we will rename ALL.

alisey:
2. ES6 class private drivers: [Pagination](https://github.com/wix/wix-ui/blob/master/packages/wix-ui-core/src/components/Pagination/Pagination.private.driver.ts), [Input](https://github.com/wix/wix-ui/blob/master/packages/wix-ui-core/src/components/Input/Input.private.driver.ts).

3. Cool.

shlomitc:
Reagarding the drivers issue:
1. Drivers are used to simplify repeated actions and name them to make it simple to reuse and read.
1. We expose drivers to our users to let them test common use cases without knowing the internal parts of the component. We must test this driver so we can assure they actually work.

As both private and public drivers share functionality it sounds to me like one should use the other. Here's a suggestion how to solve it:

```ts
import {getArrowElement, popoverDriver} from './Popover.driver.ts';

export const popoverPrivateDriver = (args) => {
  const publicDriver = popoverDriver(args)
  const {element, eventTrigger} = args;
  return {
     getArrowOffset: () => {...},
     ...publicDriver
  }
}
```

That way we have only one driver we use in our test and we can easily declare which part goes to the consumers and which are only for our tests.

alisey:
@shlomitc I thought the plan is to move to public drivers with agnostic async API that can work with Protractor/Selenium/jsdom/Puppeteer. [For example](https://github.com/wix-incubator/unidriver).

This would make APIs of public and private drivers different and incompatible.

Also, as I said before, public drivers are unlikely to have many methods, and there wouldn't be any serious overhead in defining the same method twice - for public and private driver.

As a concrete example let's take input components: calendar, number input, slider, dropdown. For such components public drivers need only getValue and setValue. Can we reuse those for unit tests? Let's see, in calendar there are two ways to set the value: by entering it into an input, and by opening a dropdown and clicking a date in that dropdown. If you were to use `publicDriver.setValue()` in a unit test, how do you know which method it uses, and as a consequence what exactly is being tested?

I agree that there should be tests for public drivers, I just don't agree that the way to achieve this goal is to use public drivers for unit tests. It's like trying to test two different things at the same time using each other. You're not sure that the component is working, and you're not sure that the driver is working, but somehow combining them proves that both are working.

packages/wix-ui-core/src/components/Popover/Popover.driver.private.ts
```diff
@@ -1,12 +1,26 @@
-const queryDocumentOrElement = (element, query) => ((element && element.querySelectorAll(query)[0]) || document && document.querySelector(query));
+import { BaseDriver , DriverArgs} from './../../common/BaseDriver';
 
+const queryDocumentOrElement = (element, query) => ((element && element.querySelectorAll(query)[0]) || document && document.querySelector(query));
```
alisey: Here you could use `element.querySelector`

packages/wix-ui-core/src/components/Popover/Popover.driver.private.ts
```diff
@@ -1,12 +1,26 @@
...
+const getTargetElement = (element: Element | undefined) => element && element.querySelectorAll('[data-hook="popover-element"]')[0];
```
alisey: And here

packages/wix-ui-core/src/components/Popover/Popover.driver.private.ts
```diff
@@ -1,12 +1,26 @@
...
+const getContentElement = (element: Element | undefined) => queryDocumentOrElement(element, '[data-hook="popover-content"]');
 const getArrowElement = (element: Element | undefined) => queryDocumentOrElement(element, '[data-hook="popover-arrow"]');
 const getPortalElement = (element: Element | undefined) => queryDocumentOrElement(element, '[data-hook="popover-portal"]');
 
-export const popoverPrivateDriverFactory = ({element, eventTrigger}) => ({
-  getArrowOffset: () => {
-    const {top, left, right, bottom} = (getArrowElement(element) as HTMLElement).style;
+export class PopoverDriverPrivate extends BaseDriver {
+
+  constructor(driverArgs: DriverArgs) {
```
alisey: I don't think you need to define this if you're just calling super.

packages/wix-ui-core/src/components/Popover/Popover.driver.ts
```diff
@@ -1,20 +1,30 @@
...
+
+  get p() {
+    return this[PRIVATE_DRIVER];
+  }
+
+  getTargetElement = () => this.p.getTargetElement();
+  getContentElement = () => this.p.getContentElement();
+  isTargetElementExists = () => this.p.isTargetElementExists();
+  isContentElementExists = () => this.p.isContentElementExists();
+  inlineStyles = () => this.element.style;
```
alisey: Don't think consumers should need inline styles.

packages/wix-ui-core/src/components/Popover/Popover.spec.tsx
```diff
@@ -59,9 +60,9 @@ describe('Popover', () => {
         onMouseEnter,
         onMouseLeave,
       }));
-
-      Simulate.mouseEnter(container.componentNode);
-      Simulate.mouseLeave(container.componentNode);
+      const driver = createDriver();
+      driver.mouseEnter();
```
alisey: This is a good example of what I'm talking about abstraction level too high for unit tests.
Does this simulate an event on the root element, or on the trigger element, or on the popup element? There are three levels of indirection I need to go through to understand what is actually going on in this test.

I prefer the previous version much more. It's exactly the level of abstraction suitable for unit tests. You can see what even you are dispatching and what element you are dispatching it to.

packages/wix-ui-core/src/components/Popover/Popover.spec.tsx
```diff
@@ -89,10 +90,10 @@ describe('Popover', () => {
       await container.render(popoverWithProps(
         {placement: 'bottom', shown: false, timeout: 10}
       ));
-  
-      expect(queryPopoverContent()).toBeTruthy();
+      const driver = createDriver();
+      expect(driver.isContentElementExists()).toBeTruthy();
       await eventually(() => {
-        expect(queryPopoverContent()).toBeNull();
+        expect(driver.isContentElementExists()).toBeFalsy();
```
alisey: Don't use `toBeFalsy()` when a more strict assertion is possible. When the element doesn't exist it's not "falsy", it's null. And if it's turns out to be some other falsy value, then something went wrong and needs to be fixed.

packages/wix-ui-core/src/components/Popover/Popover.spec.tsx
```diff
@@ -128,9 +130,11 @@ describe('Popover', () => {
         appendTo: portalContainer.node
       }));
 
-      expect(queryPopoverContent().parentElement).toBe(queryPopoverPortal());
-      expect(queryPopoverPortal().parentElement).toBe(portalContainer.node);
-      expect(queryPopoverPortal().classList).toContain(styles.root);
+      const driver = createDriver();
```
alisey: Wait, what? Is this a mistake?

packages/wix-ui-core/src/components/Popover/Popover.spec.tsx
```diff
@@ -179,8 +187,78 @@ describe('Popover', () => {
           </div>
         </div>
       );
+      
+      const privateDriver = createDriver();
+      expect(privateDriver.getPortalElement().parentElement).toBe(container.node.firstChild);
+    });
+  });
+
+  describe('public driver', () => {
```
alisey: cool

packages/wix-ui-core/src/common/BaseDriver.ts
```diff
@@ -0,0 +1,21 @@
+import { Simulate } from 'react-dom/test-utils';
+
+export interface DriverArgs {
+  element: HTMLElement;
+  eventTrigger?: typeof Simulate;
+}
+
+export class BaseDriver {
```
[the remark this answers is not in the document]
hepiyellow: What do you mean "interface of base driver is too restrictive"?
you mean like is defines a `click()` method, and the extending class wants to define a `click(index)` method?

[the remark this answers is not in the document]
alisey: That could also be an issue, but I meant something different. [This interface](https://github.com/wix/wix-ui/blob/630625f02288c441392cccfc968dd0f62ce2e90a/packages/wix-ui-test-utils/src/driver-factory/createDriverFactory.tsx#L13) doesn't accept instance of the component, and people, maybe because changing the API is too much work, resort do hacks to overcome this limitation.

So I'm just saying it's not so clear to me how the interface of the base driver should look.

shlomitc:
It's been open for a while now without a decision.
I'm closing it but keeping the branch alive to refer to it later.

PR #601, opened 2018-06-26
<Tooltip /> - add new "visible" prop, if set to false, then tooltip will be hidden

Faradey27:
`<Tooltip visible={false} />` - such tooltip will be never visible.
`<Tooltip visible={true} />` - will work as currently(as simple tooltip)

By default, visible === true


shlomitc:
Hey. so this is a classic case of having the state of the component in the hands of the consumer.
currently, `isOpen` was only an internal state and now you want to make it controlled value.

Please see [this PR](https://github.com/wix/wix-style-react/pull/1826#issuecomment-397889503) it was closed but I explain there how "controllable" props should be handled.

In this case, we should expose the `isOpen` value. maybe we can give it a better name like `show`.

Tell me if you have any questions

PR #605, opened 2018-06-29
feat(Input): support `style` prop

argshook:
* remove explicit prop for `width`
* support inline style from `style` prop instead

### DIFF
--- a/packages/wix-ui-core/src/components/Input/Input.spec.tsx
+++ b/packages/wix-ui-core/src/components/Input/Input.spec.tsx
@@ -95,7 +95,7 @@ describe('Input', () => {
       const styles = {
         width: '100px',
         background: 'blue',
-        position: 'relative' as 'relative',
+        color: 'green',
         display: 'inline-flex'
       };
       const driver = await render(<Input style={styles}/>);
--- a/packages/wix-ui-core/src/components/Input/Input.spec.tsx
+++ b/packages/wix-ui-core/src/components/Input/Input.spec.tsx
@@ -95,7 +95,7 @@ describe('Input', () => {
       const styles = {
         width: '100px',
         background: 'blue',
-        position: 'relative',
+        position: 'relative' as 'relative',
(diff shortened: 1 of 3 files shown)


packages/wix-ui-core/src/components/Input/Input.tsx
```diff
@@ -34,7 +34,7 @@ export interface InputProps {
   value?: string;
   id?: string;
   name?: string;
-  width?: string;
+  style?: object;
```
alisey: Should be `React.CSSProperties`

packages/wix-ui-core/src/components/Input/Input.spec.tsx
```diff
@@ -95,7 +95,7 @@ describe('Input', () => {
       const styles = {
         width: '100px',
         background: 'blue',
-        position: 'relative',
+        position: 'relative' as 'relative',
```
alisey: What happened here, does TS complain about something?

argshook: yes, it' throws up

```
Type '{ style: { width: string; background: string; position: string; display: string; }; }' is not assignable to type 'IntrinsicAttributes & IntrinsicClassAttributes<Input> & Readonly<{ children?: ReactNode; }> & Rea...'.
  Type '{ style: { width: string; background: string; position: string; display: string; }; }' is not assignable to type 'Readonly<InputProps>'.
    Types of property 'style' are incompatible.
      Type '{ width: string; background: string; position: string; display: string; }' is not assignable to type 'CSSProperties'.
        Types of property 'position' are incompatible.
          Type 'string' is not assignable to type 'PositionProperty'.
```

related https://github.com/Microsoft/TypeScript/issues/11465

alisey: Maybe we could simply remove 'relative' from here in the meanwhile.

PR #608, opened 2018-07-02
AddressInput - Adding suffix and fixedFooter props

giladartzi:


packages/wix-ui-core/src/components/AddressInput/AddressInput.tsx
```diff
@@ -319,11 +327,13 @@ export class AddressInput extends React.PureComponent<AddressInputProps, Address
...
             placeholder,
             disabled: readOnly,
-            value
+            value,
+            suffix
         };
 
         const states = {};
-        const timeout = options.length > 0 ? 150 : 0;
+        const hasOptions = options.length > 0;
```
lbelinsk: Correct me if I am wrong, but isn't the internal Dropdown takes care of not rendering if the options are empty?
https://github.com/wix/wix-ui/blob/07ce1724cb9cba62413fa875ba2e667ecfdde36d/packages/wix-ui-core/src/baseComponents/Dropdown/Dropdown.tsx#L200

packages/wix-ui-core/src/components/AddressInput/AddressInput.spec.tsx
```diff
@@ -489,6 +489,44 @@ describe('AddressInput', () => {
...
+
+        it('Should NOT show suffix', () => {
+            init();
+            expect(driver.getSuffix()).toBeFalsy();
+        });
+    });
+
+    describe('fixedFooter', () => {
+        it('Should show fixedFooter', async () => {
+            init({fixedFooter: <div data-hook="fixed-footer"/>});
```
lbelinsk: In order to allow people to target the `fixedFooter` element from outside, we should expose the className for the container of the dom element where the fixedFooter will sit in:
https://github.com/wix/wix-ui/blob/07ce1724cb9cba62413fa875ba2e667ecfdde36d/packages/wix-ui-core/src/baseComponents/DropdownContent/DropdownContent.tsx#L156

You can add a className (also an empty one in its stylesheet)
```
.fixedFooter: {}
```
And add some selector to the driver.
You won't have to pass here in the test a data-hook as well if you do that, but this is not why I am raising it.

packages/wix-ui-core/src/components/AddressInput/AddressInput.tsx
giladartzi: That's correct. Though in this case we use the `hasOptions` variable to determine whether pass the fixedFooter through - to avoid a situation where the `contentElement` is visible although there are no suggestions (and only showing the footer).

packages/wix-ui-core/src/components/AddressInput/AddressInput.spec.tsx
giladartzi: In my opinion, the `fixedFooter` element is determined by the user, and thus they can add whatever class name (or data-hook) to the element as they wish

lbelinsk: That is ok, they can add whatever they want, but I think i case they would have to change the container of it they would want to have access to the footer.
That is a different task, not directly related to your pr, so all good.

packages/wix-ui-core/src/components/AddressInput/AddressInput.tsx
lbelinsk: ok

PR #610, opened 2018-07-02
wix-storybook-utils - Add generic renderStory config

hepiyellow:
## Main motivation:
Allow rendering a test page instead of the AutoDocs story (for e2e tests)

## Secondary outcome
As an outcome of opening a `renderStory` property, it makes sense to pass the `renderStory` all the AutoDocs "context" , this object:
```js
{
    isTestMode: isE2E,
    component,
    displayName,
    componentProps,
    examples,
    exampleProps,
    exampleImport,
    codeExample,
    _config,
    _metadata
    }
```
So that we give total freedom to render a custom story page which may or may not use wix-storybook-utils component such as `<AutoExample/>`.

### DIFF
--- a/packages/wix-storybook-utils/src/Story/index.js
+++ b/packages/wix-storybook-utils/src/Story/index.js
@@ -16,6 +16,7 @@ export default ({
   exampleProps,
   exampleImport,
   codeExample,
+  renderStory,
   _config,
   _metadata
 }) =>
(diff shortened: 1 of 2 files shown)


argshook:
@hepiyellow 
the idea may be nice but implementation is incomplete.

moreover, if all you need is a custom story, why do all this trickery, why use wix-storybook-utils at all? just do good old custom story with `storiesOf`

if you want to get parsed data, then i'd argue that we should use `react-autodocs-utils` directly instead of trying to squeeze into wix-storybook-utils but not use any of its functionality

do i miss something important?

hepiyellow:
## Custom story
As I wrote, the motivation is for test examples (which can not use the AutoExample).
I just thought that a separate story brings 2 issues:
1) If we add a separate story for e2e tests, we needs to figure out:
 - how do we structure these stories? a separate category "Tests" with numbering like the original UX spec?
```
...
3. Inputs
3.1 TextField
4. Selection
99. Tests
99.3 Inputs
99.3 Inputs - TextField
...
```
This way good in "almost hiding" the test from the users. But we are limited to only 2 levels of depth in storybook, so it makes it difficult/ugly.

OR:

```
...
3. Inputs
3.1 TextField
3.1 + TextField [TEST]
4. Selection
...
```
This way the consumers will see this story.
If we want to hide this story, we need to do it in build time, so that when pushing a PR storybook to surge, we WILL have the test stories, but when building a storybook for the master WSR we will NOT have the test stories.

2)  overhead of code. Usually the extra code in the main story needs to be used in the test examples as well, so it is nice to have it together. 

Anyway, we already are mixing tests and the AutoDocs story, so I am not making things worse, just easier to use and more flexible. 

## Flexible Story Template
This part is less important for me now, I can remove it. We can discuss on it later.

hepiyellow:
Alternatively, this PR helps my need for test examples:
https://github.com/wix/wix-ui/pull/611

argshook:
> Anyway, we already are mixing tests and the AutoDocs story, so I am not making things worse, just easier to use and more flexible.

and that is no good, we're digging a hole for ourselves with it. i would
argue that PRs like these do actually make things a bit worse from
autodocs perpsective.

documentation and e2e must be separated and the more we work on it the more obvious it becomes.

i think now it's much better/faster/stronger/easier to simply do e2e stories under, say, `13. Work in progress` category, or even better just separate new `Test cases` category.
there's no harm if consumers see it.

later on we should go ahead and just separate e2e from autodocs. we
already had such usecase a year or so ago (a tiny separate express server, serves small page with same bundle from `build-storybook`, so no duplicate bundles or anything)

Autodocs was never meant to cover testing and `AutoexampleDriver` is obvious nice-to-have/hack. let's not hack more on it.

argshook:
closing, since we decided to day to use separate story page

PR #611, opened 2018-07-02
wix-storybook-utils - Add exampleIndex URL param

hepiyellow:
An alternativ for https://github.com/wix/wix-ui/pull/610

### DIFF
--- a/packages/wix-storybook-utils/src/Story/index.js
+++ b/packages/wix-storybook-utils/src/Story/index.js
@@ -23,31 +23,53 @@ export default ({
     .storiesOf(category, module)
     .add(
       storyName || _metadata.displayName,
-      () =>
-        isE2E ?
-          <div>
-            <AutoExample
-              isInteractive={false}
-              ref={ref => global.autoexample = ref}
-              component={component}
-              componentProps={componentProps}
-              parsedSource={_metadata}
+      () => {
+        if (isE2E) {
+          const exampleIndex = queryString.parse(window.location.search).exampleIndex;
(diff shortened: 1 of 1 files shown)


argshook:
please just do a separate custom story without autodocs where you have total freedom

hepiyellow:
We are already making tests on the AutoExample, why should we add another story.
It is easier when all tests use the same story.

argshook:
why e2e with autodocs is bad:

* `AutoexampleDriver` is a hack, it supports only a fraction of use cases (5, to be exact https://github.com/wix/wix-ui/blob/master/packages/wix-storybook-utils/src/AutoExample/protractor.driver.js and :crossed_fingers: for other)
* need to do `if else`s in wix-storybook-utils:
    * additional apis
    * need to test them (it becomes infra)
    * need to maintain it (because it's infra, many projects depend on it)
    * new unsupported use cases pop up every now and then (it's like fifth time already), so rinse & repeat this list
* missing api for resetting focus, resetting state
* possible clashes scrolling (e.g. to test tooltips, sticky headers etc.)
* stuff from autodocs may leak into test, this means more `if else`s, more complications
* a mix between wix-storybook-utils and wix-ui-test-utils (e.g. `getStoryUrl`, `withExamples` param)
* after all, it's autodocs, not autotest

short term decision: use separate story page for specific e2e
this means no changes to wix-storybook-utils, simply use existing tools, urgent to bcc

long term decision: separate e2e and documentation entirely. use same
bundle but serve it separately from storybook with some tiny express server. it already happened a while ago, not a big deal.

argshook:
closing, since we decided to day to use separate story page

PR #622, opened 2018-07-09
feat(TimePicker): support width prop

argshook:
useful in general and also required in wix-ui-backoffice

edit: needed for https://github.com/wix/wix-ui-backoffice/pull/178

### DIFF
--- a/packages/wix-ui-core/src/components/TimePicker/TimePicker.driver.ts
+++ b/packages/wix-ui-core/src/components/TimePicker/TimePicker.driver.ts
@@ -11,7 +11,7 @@ export const timePickerDriverFactory = ({element, eventTrigger}) => {
     getInputElement: () => inputDriver.getInput(),
 
     /** returns inline style width of TimePicker */
-    getInlineWidth: () => inputDriver.getInput().parentNode.style.width,
+    getInlineStyle: () => window.getComputedStyle(inputDriver.getInput().parentNode),
 
     /** returns a boolean indicating if the timePicker is disabled */
     isDisabled: () => inputDriver.isDisabled(),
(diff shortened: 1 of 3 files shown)


packages/wix-ui-core/src/components/TimePicker/TimePicker.driver.ts
```diff
@@ -6,26 +6,40 @@ export const timePickerDriverFactory = ({element, eventTrigger}) => {
   return {
     /** checks if element exists */
     exists: () => !!element,
+
     /** returns the input element */
     getInputElement: () => inputDriver.getInput(),
+
+    /** returns inline style width of TimePicker */
+    getInlineWidth: () => inputDriver.getInput().parentNode.style.width,
```
lbelinsk: Not too low level? maybe getWidth?
And also, maybe better to implement with `getComputedStyles`

packages/wix-ui-core/src/components/TimePicker/TimePicker.tsx
```diff
@@ -419,6 +422,7 @@ export class TimePicker extends React.PureComponent<TimePickerProps, TimePickerS
         onMouseMove = {this._onMouseMove}
         onClick     = {this._onClick}
         onDragStart = {e => e.stopPropagation()}
+        style       = {{width}}
```
lbelinsk: Maybe since this is a core component, we should just support `style` prop?
Otherwise, what would happen when we would need height, ...

We have some components that already supports inline styles in the core, due to santa limitations.

argshook: maybe we should, i wasn't sure about it so wanted someone to take a look. i guess it's better that way, i'll update

PR #623, opened 2018-07-10
Import typescript helper methods to reduce bundle size

yakirn:
See [tslib](https://www.npmjs.com/package/tslib) for more info.

### DIFF
--- a/packages/wix-ui-core/package.json
+++ b/packages/wix-ui-core/package.json
@@ -56,6 +56,7 @@
     "react-transition-group": "^2.2.1",
     "shallowequal": "1.1.0",
     "type-zoo": "3.1.1",
+    "tslib": "^1.9.3",
     "wix-ui-jss": "^1.0.0",
     "wix-ui-test-utils": "^1.0.0",
     "unidriver": "^1.0.0-alpha"
--- a/packages/wix-ui-core/tsconfig.json
+++ b/packages/wix-ui-core/tsconfig.json
@@ -2,6 +2,7 @@
   "compilerOptions": {
     "sourceMap": true,
     "declaration": true,
+    "importHelpers": true,
     "module": "commonjs",
     "moduleResolution": "node",
     "strict": true,


hepiyellow:
@yakirn 
- How much does it reduce the bundle size? (How much per component avg., since users are not using all components usually)
- I see [this issue](https://github.com/Microsoft/tslib/issues/49) with `tslib`, does it seem to be relevant for us? I think we are using the es5 polyfill.

I guess that if the bundle size reduction is very small (in %), then shouldn't use this, it only introduces risk.

@alisey @roymiloh  What do you think?

yakirn:
@hepiyellow it can reduce a lot, but it's really hard to tell.
For example, the following code will be duplicate for every module that has a class that extends another class (which is very common with React):
```
var __extends = (this && this.__extends) || (function () {
    var extendStatics = Object.setPrototypeOf ||
        ({ __proto__: [] } instanceof Array && function (d, b) { d.__proto__ = b; }) ||
        function (d, b) { for (var p in b) if (b.hasOwnProperty(p)) d[p] = b[p]; };
    return function (d, b) {
        extendStatics(d, b);
        function __() { this.constructor = d; }
        d.prototype = b === null ? Object.create(b) : (__.prototype = b.prototype, new __());
    };
})();
```
with `"importHelpers":true` it will be included in the bundle once, and imported instead of duplicated into each module that uses `extends`
In premium we use it in every project. I don't remember exact numbers but It could save 10s of KBs before gzip, and 2-3 after.
I discussed this with @shlomitc ,it could be more correct to make it the consumer's responsibility.
Maybe we should add this to the `wix-js` generator instead?

hepiyellow:
When you mean "make it a consumer's responsibility" do you mean using the webpack ProvidePlugin, as suggested here (see seconds answer)?
https://stackoverflow.com/questions/36556772/webpack-and-typescript-extends

@ranyitz @ronami Which way do you recommend?

yakirn:
@hepiyellow I actually meant that the consumer will be the one in charge of adding `"importHelpers":true` in their own `tsconfig`, But looks like what you're offering is better.
Anyway, I'm curious what the bundler masters have to say :)

roymiloh:
@hepiyellow I don't think we need to complicate it with Webpack stuff. This PR is great - let's merge:)

shlomitc:
@roymiloh - but is this effective for libraries

wix-ui-core outputs components which are consumed by wix-style-react and introduced to some consuming project.

should the consumer be responsible to do this optimization to his code?

roymiloh:
It is **very** effective for libs. Consumers shouldn't do anything, it just works.

alisey:
The main question is how much it actually reduces the bundle size in this specific project. `tslib` itself is 6kb minified, not too much, but not too little either.

hepiyellow:
@yakirn Can you make a measurement on how much `tslib` reduces the bundle size for `wix-ui-core`?

yakirn:
@hepiyellow yes, will do.

yakirn:
@hepiyellow it's hard to tell exactly as there is no single bundle for `wix-ui-core`, so I measured 2 components:

1. Button: reduced ~240 bytes.
2. Checkbox: reduce 716 bytes.

As I wrote before, every time we use class extends or spread operator a function is duplicated instead of being imported from a single location.
The amount of bytes you save will vary depends on how many components you used, and likely to grow over time.

lbelinsk:
Sounds like a good idea to add it. I say let's merge this

yakirn:
@lbelinsk I updated this branch with changes from master, there should be no conflicts now.

PR #628, opened 2018-07-15
Timepicker

guybenron:


packages/wix-ui-core/src/components/TimePicker/TimePicker.spec.tsx
```diff
@@ -233,4 +237,13 @@ describe('TimePicker', () => {
       expect(driver.getValue()).toEqual('--:59');
     });
   });
+
+  it('should support focus state', () => {
+    const driver = createDriver(<TimePicker/>);
+    driver.focus();
+    expect(stylableUtil.hasStyleState(driver.getTimePickerElement(), 'focus')).toBe(true);
```
hepiyellow: Use toBeTruthy

packages/wix-ui-core/src/components/TimePicker/TimePicker.driver.ts
```diff
@@ -7,6 +7,9 @@ export const timePickerDriverFactory = ({element, eventTrigger}) => {
     /** checks if element exists */
     exists: () => !!element,
 
+    /** returns the time picker element */
+    getTimePickerElement: () => element,
```
hepiyellow: Rename to element()

packages/wix-ui-core/src/components/TimePicker/TimePicker.spec.tsx
```diff
@@ -233,4 +237,13 @@ describe('TimePicker', () => {
...
+    driver.blur();
+    expect(stylableUtil.hasStyleState(driver.getTimePickerElement(), 'focus')).not.toBe(true);
```
hepiyellow: Use toBeFalsy()

packages/wix-ui-core/src/components/TimePicker/TimePicker.spec.tsx
```diff
@@ -233,4 +237,13 @@ describe('TimePicker', () => {
       expect(driver.getValue()).toEqual('--:59');
     });
   });
+
+  it('should support focus state', () => {
+    const driver = createDriver(<TimePicker/>);
+    driver.focus();
```
hepiyellow: Check that ha no govus state before focusing

packages/wix-ui-core/src/components/TimePicker/TimePicker.tsx
```diff
@@ -221,7 +225,8 @@ export class TimePicker extends React.PureComponent<TimePickerProps, TimePickerS
       this.setState({value}, () => { if (value !== this.props.value) { onChange(value); } });
     }
 
-    this.props.onBlur && this.props.onBlur();
+    this.setState({focus: false});
```
hepiyellow: In line 225, there is another `setState`.
It would be better to batch these 2 into 1 update, so that there won't be 2 renders.
You can either aggregate them to one call to `setState`, or use the `setState` callback (2nd param) to do the `focus` update.

hepiyellow: nevermind

packages/wix-ui-core/src/components/TimePicker/TimePicker.spec.tsx
```diff
@@ -233,4 +237,14 @@ describe('TimePicker', () => {
       expect(driver.getValue()).toEqual('--:59');
     });
   });
+
```
hepiyellow: Add test case for `onFocus`, `onBlur` callbacks being called.

guybenron: There are already tests for this (see line 58, don't know how to link directly here...)

PR #630, opened 2018-07-16
Address input - managing input state

giladartzi:
Maintaining `this.state.inputValue` instead of directly passing `this.props.value` to the `<Input/>` component. This is done in order to let the user see what they type as they go.

### DIFF
--- a/packages/wix-ui-core/src/components/AddressInput/AddressInput.spec.tsx
+++ b/packages/wix-ui-core/src/components/AddressInput/AddressInput.spec.tsx
@@ -1,4 +1,5 @@
 import * as React from 'react';
+import {Simulate} from 'react-dom/test-utils';
 import {ReactDOMTestContainer} from '../../../test/dom-test-container';
 import {addressInputDriverFactory} from './AddressInput.driver';
 import {AddressInput, Handler} from './AddressInput';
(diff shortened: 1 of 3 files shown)


packages/wix-ui-core/src/components/AddressInput/AddressInput.tsx
```diff
@@ -327,7 +333,7 @@ export class AddressInput extends React.PureComponent<AddressInputProps, Address
             onBlur: this._handleOnBlur,
             placeholder,
             disabled: readOnly,
-            value,
+            value: this.state.inputValue,
```
lbelinsk: why do you need to be both controlled & uncontrolled?
The state is just doing what the props.value is

packages/wix-ui-core/src/components/AddressInput/AddressInput.spec.tsx
```diff
@@ -249,6 +250,37 @@ describe('AddressInput', () => {
...
+            
+            const addressInputDriver = addressInputDriverFactory({element: wrapper.getDOMNode(), eventTrigger: Simulate});
+            addressInputDriver.setValue('n');
+            expect(addressInputDriver.getValue()).toBe('n');
+            const newValue = '321 Ibn Gabirol st.';
+            wrapper.setProps({value: newValue});
+            expect(addressInputDriver.getValue()).toBe(newValue);
+            addressInputDriver.setValue('n');
+            expect(addressInputDriver.getValue()).toBe('n');
+            wrapper.setProps({value: newValue});
```
lbelinsk: what?
You set the props, so willReceiveProps is being called, but the new value is not set?

guybenron: Setting the same value twice to check that it only replaces when the prop is different from the previous prop

PR #631, opened 2018-07-17
AddressInput - adding focus and blur methods

giladartzi:


packages/wix-ui-core/src/components/AddressInput/AddressInput.spec.tsx
```diff
@@ -475,6 +475,26 @@ describe('AddressInput', () => {
             expect(onBlur).toHaveBeenCalled();
         });
 
+        it('Should have a focus and blur method', () => {
```
lbelinsk: I can imagine that we would want onBlur and onFocus. It always comes to that :)
Your call.

giladartzi: For the moment let's leave it as it is, if the need arises, we can add it later.

alisey:
@lbelinsk @giladartzi this PR either should to be rolled back or the test should to be fixed - it unreasonably expects `#root` to be present in the document and to be empty. It also doesn't unmount the component after the test.

lbelinsk:
About the first part, I don't think it is that critical, but there is no reason why it can't just attach to some div element.

About the unmounting - good catch!

alisey:
> there is no reason why it can't just attach to some div element

It shouldn't attach to some div element. The tests have to share the page with other tests and the Mocha reporter. There are two methods that allow you to create a container with a guarantee that you're not going to interfere with other things on the page.

1. https://github.com/wix/wix-ui/blob/e7199fe9b1fa27e04193d5f3c3328f24d0104667/packages/wix-ui-core/test/dom-test-container.ts#L8

2. https://github.com/wix/wix-ui/blob/e7199fe9b1fa27e04193d5f3c3328f24d0104667/packages/wix-ui-core/test/dom-test-container.ts#L17

PR #635, opened 2018-07-19
InternalAddress - Using the correct data type according to product sp…

giladartzi:
…ecification documents

### DIFF
--- a/packages/wix-ui-core/src/clients/GoogleMaps/google2address/consts.ts
+++ b/packages/wix-ui-core/src/clients/GoogleMaps/google2address/consts.ts
@@ -1,0 +1,163 @@
+export const googleResponse = {
+    "address_components": [
+        {
+            "long_name": "137",
+            "short_name": "137",
+            "types": [
+                "street_number"
+            ]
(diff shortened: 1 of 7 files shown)


packages/wix-ui-core/src/clients/GoogleMaps/google2address/consts.ts
```diff
@@ -0,0 +1,163 @@
+export const googleResponse = {
```
shlomitc: rename to fixtures

packages/wix-ui-core/src/clients/GoogleMaps/google2address/google2address.ts
```diff
@@ -1,98 +1,6 @@
...
-
-    for (const key in result) {
-        if (isUndefined(result[key])) {
-            delete result[key];
-        }
-    }
-
-    return result;
-}
+const intersection = require('lodash/intersection');
```
shlomitc: try using `import intersection = require('lodash/intersection');`

PR #637, opened 2018-07-22
Add WSR Lint Plugins

yairhaimo:
This PR adds an `eslint` and a `tslint` plugin.
Both have only one rule, `no-full-wsr-lib`. 
This rule checks that a user does not import the entire `wix-style-react` library.
In the future, when WSR will support tree shaking, this rule can be removed.

### DIFF
--- a/packages/eslint-plugin-wix-style-react/package.json
+++ b/packages/eslint-plugin-wix-style-react/package.json
@@ -1,0 +1,26 @@
+{
+  "name": "eslint-plugin-wix-style-react",
+  "version": "0.0.0",
+  "description": "Wix Style React Lint Plugin",
+  "keywords": [
+    "eslint",
+    "eslintplugin",
+    "eslint-plugin"
+  ],
+  "author": "yairhaimo",
+  "main": "lib/index.js",
+  "scripts": {
+    "test": "mocha tests --recursive"
+  },
+  "dependencies": {
+    "requireindex": "~1.1.0"
+  },
+  "devDependencies": {
+    "eslint": "~3.9.1",
+    "mocha": "^3.1.2"
+  },
+  "engines": {
+    "node": ">=0.10.0"
+  },
+  "license": "ISC"
+}
--- a/packages/tslint-plugin-wix-style-react/package.json
+++ b/packages/tslint-plugin-wix-style-react/package.json
@@ -1,0 +1,36 @@
+{
+  "name": "tslint-plugin-wix-style-react",
+  "version": "1.0.0",
+  "description": "",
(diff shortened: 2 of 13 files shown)


packages/eslint-plugin-wix-style-react/package.json
```diff
@@ -0,0 +1,26 @@
+{
+  "name": "eslint-plugin-wix-style-react",
+  "version": "0.0.0",
```
shlomitc: is this ok to be `0.0.0`? or should it be `1.0.0`?

packages/tslint-plugin-wix-style-react/package.json
```diff
@@ -0,0 +1,36 @@
+{
+  "name": "tslint-plugin-wix-style-react",
+  "version": "1.0.0",
+  "description": "",
+  "main": "dist/",
```
shlomitc: shouldn't there be one main index file? or `dist/src/noFullWsrLibRule.js`?

packages/tslint-plugin-wix-style-react/package.json
```diff
@@ -0,0 +1,36 @@
...
+  "scripts": {
+    "build": "tsc",
+    "test": "jest"
+  },
+  "keywords": [],
+  "author": "",
+  "license": "ISC",
+  "dependencies": {
+    "tslint": "^5.10.0"
```
shlomitc: shouldn't it be a `peerDependency`?

yairhaimo: yep

packages/tslint-plugin-wix-style-react/package.json
yairhaimo: Good catch. i changed it to be the `tslint.json` of the package which setups the `ruleDirectory` to be dist.
A consumer (`tslint-config-yoshi` for example) would just need to do this now:
```json
{
  "extends": [
    "tslint-plugin-wix-style-react"
  ]
}
```

packages/tslint-plugin-wix-style-react/package.json
```diff
@@ -0,0 +1,36 @@
...
+  "main": "tslint.json",
+  "scripts": {
+    "build": "tsc",
+    "test": "jest"
+  },
+  "keywords": [],
+  "author": "",
+  "license": "ISC",
+  "peerDependencies": {
+    "tslint": "^5.10.0"
```
shlomitc: So it should also be a devDep, otherwise tests won’t run.
And maybe typescript should be here as well

PR #642, opened 2018-07-24
AddressInputSpec - using DOMContainer

giladartzi:


packages/wix-ui-core/src/components/AddressInput/AddressInput.spec.tsx
```diff
@@ -493,6 +494,8 @@ describe('AddressInput', () => {
             expect(document.activeElement).toBe(input);
             instance.blur();
             expect(document.activeElement).not.toBe(input);
+
+            domContainer.remove();
```
alisey: React won't clean up event handlers correctly if you do this. You need to call `wrapper.unmount()` first.

Alternatively you do `{attachTo: container.node}` without using `createDOMContainer`, I think that should work.

packages/wix-ui-core/src/components/AddressInput/AddressInput.spec.tsx
```diff
@@ -1,6 +1,6 @@
 import * as React from 'react';
 import {Simulate} from 'react-dom/test-utils';
-import {ReactDOMTestContainer} from '../../../test/dom-test-container';
+import {ReactDOMTestContainer, createDOMContainer} from '../../../test/dom-test-container';
```
alisey: Don't need the second import here.

PR #643, opened 2018-07-24
chore(protractor.conf.js): double default jasmine timeout

argshook:


argshook:
ok, im reanimating this PR cause we keep getting timed out builds and until we go about investigating root causes, i say let's just increase timeout and sleep well

PR #645, opened 2018-07-24
Adding ui-autotools to wix-ui-core

Zombiefruit:


packages/wix-ui-core/.gitignore
```diff
@@ -19,4 +19,7 @@ storybook-static
 /*.d.ts
 !wallaby.js
 !protractor.conf.js
-index.st.css
\ No newline at end of file
+index.st.css
+
+# Don't ignore the meta.webpack.config.js file, necessary for sanity testing
+!meta.webpack.config.js
```
alisey: Maybe create a `config` folder?

packages/wix-ui-core/package.json
```diff
@@ -23,11 +23,12 @@
   "scripts": {
     "build": "yoshi build && npm run generate-stylable-index && npm run transpile-mixins && npm run import-path && build-storybook",
     "pr-postbuild": "npx teamcity-surge-autorelease@^1.0.0",
-    "test": "npm run test:unit && npm run test:e2e",
+    "test": "npm run test:unit && npm run autotools && npm run test:e2e",
     "posttest": "npm run lint",
     "test:watch": "yoshi test --jest --watch",
     "test:unit": "yoshi test --jest && wix-ui-mocha-runner",
     "test:e2e": "yoshi test --protractor",
+    "autotools": "npx autotools sanity",
```
alisey: If autotools is a binary we shouldn't need npx here

packages/wix-ui-core/package.json
```diff
@@ -65,9 +66,12 @@
     "@types/react-dom": "^16.0.5",
     "@types/react-onclickoutside": "^6.0.2",
     "@types/react-transition-group": "^2.0.6",
+    "@ui-autotools/registry": "^1.0.2",
+    "@ui-autotools/scripts": "^1.0.6",
     "airbnb-prop-types": "^2.10.0",
     "babel-cli": "^6.26.0",
     "babel-preset-env": "^1.6.1",
+    "chai": "^4.1.2",
```
shlomitc: why is chai needed?

packages/wix-ui-core/meta.webpack.config.js
```diff
@@ -0,0 +1,38 @@
+const path = require('path');
```
shlomitc: why is this configuration not part of the autotools package?

packages/wix-ui-core/package.json
Zombiefruit: It's a peer-dependency of the autotools. Can discuss making it a dependency

packages/wix-ui-core/meta.webpack.config.js
Zombiefruit: Because you may need to configure webpack specifically for your project (e.g. add a stylable loader if you're using stylable)

shlomitc: it should be more of a zero-configuration tool, so I suggest having this encapsulated in the tool or a different module.

besides - why do you need to create a bundle?
we can create a bundle inside our project and just pass it to `autotools`

alisey: It cannot be a zero-configuration tool unless all projects `autotools` are used with have identical webpack configs. And by identical I also mean identical versions of all webpack plugins, stylable, and typescript.

> we can create a bundle inside our project and just pass it to autotools

This might be a good idea, but wouldn't it be the opposite of zero configuration? Also there's a question of how this would work in combination with watch mode. But it's definitely something we should consider.

packages/wix-ui-core/.autotools/webpack.config.js
```diff
@@ -0,0 +1,15 @@
+const wixStorybookConfig = require('yoshi/config/webpack.config.storybook');
+const packagePath = __dirname;
```
alisey: `__dirname` is `.autotools`, the package path is one level up.

packages/wix-ui-core/.autotools/webpack.config.js
```diff
@@ -1,5 +1,6 @@
 const wixStorybookConfig = require('yoshi/config/webpack.config.storybook');
-const packagePath = __dirname;
+const path = require('path');
+const packagePath = path.join(__dirname, '..');
```
alisey: should be path.resolve

PR #646, opened 2018-07-25
AddressInput - adding disabled prop in addition to readOnly prop

giladartzi:


packages/wix-ui-core/src/components/AddressInput/AddressInput.spec.tsx
```diff
@@ -444,6 +444,11 @@ describe('AddressInput', () => {
 
         it('Should pass readOnly prop', () => {
             init({readOnly: true});
+            expect(driver.isReadOnly()).toBeTruthy();
+        });
+
+        it('Should pass disabled prop', () => {
```
lbelinsk: don't you think it is important to check that by default it is readonly=false?

PR #652, opened 2018-07-25
Focusable hoc

Faradey27:


packages/wix-ui-core/package.json
```diff
@@ -76,6 +76,7 @@
     "gh-pages-auto-release": "latest",
     "identity-obj-proxy": "^3.0.0",
     "import-path": "latest",
+    "jsdom-global": "^3.0.2",
```
hepiyellow: I think we might not need jsdom-global in wix-ui-core, since the tests are running with mocha-runner ina real dom.

packages/wix-ui-core/src/hocs/Focusable/Focusable.scss
```diff
@@ -0,0 +1,2 @@
+$focusable-focus: "[data-focus]";
```
hepiyellow: WE certainly don't need a Sass stylesheet here. These `[data-focus]` and `[data-focus-visible]` should be replaced with Stylable pseudo states.

packages/wix-ui-core/src/hocs/Focusable/FocusableMixins.scss
```diff
@@ -0,0 +1,11 @@
+@mixin FocusBox($borderColor: $B10, $boxShadow: 0 0 0 3px $F00) {
```
hepiyellow: Need to convert to Stylable classes (which can also be used as mixins)

packages/wix-ui-core/src/hocs/Focusable/FocusableHOC.tsx
```diff
@@ -0,0 +1,132 @@
+import * as React from 'react';
+import hoistNonReactMethods from 'hoist-non-react-methods';
+import {wrapDisplayName, getDisplayName, isStatelessComponent} from '../utils';
+
+/**
+ * Use this method to spread the focusable focus states onto the a component's root element.
+ * @param {object} props
+ */
+export function focusableStates(props) {
```
hepiyellow: There is no need for this function, since With Stylable ww already use the Runtime Stylesheet as a function , to spread the states onto a component's root element.

packages/wix-ui-core/src/hocs/Focusable/FocusableHOC.tsx
```diff
@@ -0,0 +1,132 @@
...
+      window.addEventListener('keyup', () => this.setMethod('keyboard'));
+    }
+  }
+
+  /**
+   * Subscribe to inputMethod change events
+   * @param {*} target used as a key to the subscribers map
+   * @param {*} callback optional to be called when the input method changes
+   */
+  subscribe = (target, callback) => this.subscribers.set(target, callback);
```
hepiyellow: missing types

packages/wix-ui-core/src/hocs/Focusable/FocusableHOC.tsx
```diff
@@ -0,0 +1,132 @@
...
+
+  /**
+   * Unsubscribe to inputMethod change events
+   * @param {*} target used as a key to the subscribers map
+   */
+  unsubscribe = target => this.subscribers.delete(target)
```
hepiyellow: missing types

packages/wix-ui-core/src/hocs/Focusable/FocusableHOC.tsx
```diff
@@ -0,0 +1,132 @@
...
+    'data-focus-visible': props.focusableIsFocusVisible
+  };
+}
+
+/**
+ * Singleton for managing current input method (keyboard or mouse).
+ */
+const inputMethod = new class {
+  // Default is keyboard in case an element is focused programmatically.
+  method = 'keyboard';
```
hepiyellow: Missing type: 
`'mouse' | 'keyboard'`

packages/wix-ui-core/src/hocs/Focusable/FocusableHOC.tsx
```diff
@@ -0,0 +1,132 @@
...
+  subscribers = new Map();
```
hepiyellow: missing type of Map

packages/wix-ui-core/src/hocs/Focusable/FocusableHOC.tsx
```diff
@@ -0,0 +1,132 @@
...
+  if (isStatelessComponent(Component)) {
+    throw new Error(`FocusableHOC does not support stateless components. ${getDisplayName(Component)} is stateless.`);
+  }
+
+  interface IFocusableHOCState {
+    focus: boolean;
+    focusVisible: boolean;
+  }
+
+  class FocusableHOC extends React.PureComponent<any, IFocusableHOCState> {
```
hepiyellow: About extending PureComponent, in WSR it doesn't really matter since, all components extends WixComponent which is a PureComponent. I think we don't really need this optimization here.

packages/wix-ui-core/src/hocs/Focusable/FocusableHOC.tsx
```diff
@@ -0,0 +1,132 @@
...
+    };
+
+    render() {
+      return (
+        <Component
+          ref={ref => this.wrappedComponentRef = ref}
+          {...this.props}
+          focusableOnFocus={this.onFocus}
+          focusableOnBlur={this.onBlur}
+          focusableIsFocused={this.state.focus || null}
```
hepiyellow: Again, these states should now be Stylable states. so they can be passed by the spreading of  `style('root',...` 
So the render function would look like:
```
import style from './Focusable.st.css`
...
render() {
   return <Component 
      ref={ref => this.wrappedComponentRef = ref}
      {...this.props}
      focusableOnFocus={this.onFocus}
      focusableOnBlur={this.onBlur}
      {...style('root', {isFocused:this.state.focus || null, isFocusVisible: this.state.focusVisible || null}, this.props)}
/>
);
```

packages/wix-ui-core/package.json
shlomitc: `jest` contains `jsdom` internally, why do you need it here?

Faradey27: no need, it was added during experiment

just forgot to remove

packages/wix-ui-core/src/hocs/Focusable/FocusableHOC.tsx
Faradey27: why? it is pretty useful to have focusableIsFocus as prop, expecially for external users

@hepiyellow

packages/wix-ui-core/src/hocs/Focusable/FocusableHOC.tsx
Faradey27: @hepiyellow can you update wix-ui-core tscofig config with noAny prop? so people will be not able to build such code

packages/wix-ui-core/src/hocs/Focusable/Focusable.scss
Faradey27: what about backcompability? i plan to proxy this via wix-style-react

hepiyellow: I’d rather not try and make it Sass compatible as well.

packages/wix-ui-core/src/hocs/Focusable/FocusableHOC.tsx
hepiyellow: What did you mean by focusablesFocus? Do you mean the isFocused state? If so, it is not a prop that the consumer needs to pass. What is the “useful” case you are thinking of?

Faradey27: currently, when I will do withFocusable(MyAwesomeComponent), ill receive inside MyAwesomeComponent prop isFocusableVisible which I can use not just for styling but also for some logic, etc.

If we will stop to send this prop and we will just use styleble states then we will prevent such thing and we will do a breaking change(which i do not really want to do)

@hepiyellow

Faradey27:
@shlomitc  can you take a look?

PR #659, opened 2018-07-29
wix-storybook-utils: Do not use the first option as a default in <List/>

yardnsm:
The `<List/>` component (used in `<AutoExample/>`), uses the first option (`0` index) as the default option. This causes a small issue when trying to choose the first option, when the field is empty. 

For example, in: https://wix-wix-style-react.surge.sh/?selectedKind=1.%20Foundation&selectedStory=1.5%20Loader&full=0&addons=0&stories=1&panelRight=0, in order to select the `tiny` option, we first need to select a different option.

screen shot 2018-07-29 at 15 46 00


argshook:
hey @yardnsm thanks for taking time on this!

however, i think im going to do [this](https://github.com/wix/wix-ui/pull/661) instead. that one chooses correct initial selected id and also fixes a long standing `[object Object]` bug

PR #662, opened 2018-07-30
TimePicker - fixing focus and blur behavior for the tickers

guybenron:


packages/wix-ui-core/src/components/TimePicker/Tickers.tsx
```diff
@@ -14,27 +14,40 @@ export interface TickersProps {
...
+    onDecrement: func,
+    /** up ticker icon */
+    tickerUpIcon: node,
+    /** down ticker icon */
+    tickerDownIcon: node
+  };
 
-Tickers.displayName = 'Tickers';
+  handleIncrement = e => {
+    e.preventDefault();
```
argshook: why do you need to prevent default?

packages/wix-ui-core/src/components/TimePicker/TimePicker.protractor.driver.ts
```diff
@@ -29,5 +34,8 @@ export const timePickerDriverFactory: DriverFactory<TimePickerDriver> = componen
     mouseEnter: async () => mouseEnter(component),
     pressKeyArrowRight: () => inputDriver.pressKey(protractor.Key.ARROW_RIGHT),
     pressKeyArrowLeft: () => inputDriver.pressKey(protractor.Key.ARROW_LEFT),
+    clickTickerUp: async () => tickerUp.click(),
+    clickTickerDown: async () => tickerDown.click(),
+    getText: () => inputDriver.getText(),
```
argshook: perhaps should be called `getValue` as this is public api and so need to be semantic to some extent

packages/wix-ui-core/src/components/TimePicker/TimePicker.tsx
```diff
@@ -390,20 +389,20 @@ export class TimePicker extends React.PureComponent<TimePickerProps, TimePickerS
...
-    if (isValidTime(value)) { onChange(value); }
-  }
-
-  _decrement(field?: FIELD) {
-    let {value} = this.state;
-    const {step, onChange} = this.props;
-    value = decrement({value, field: field || this._lastFocusedField || FIELD.MINUTE, step});
-    this.setState({value});
-    if (isValidTime(value)) { onChange(value); }
+  _tick(action: Function) {
```
argshook: you know if you do `_tick = (action: Function) => ...` then you don't need any of the `bind`s in constructor

packages/wix-ui-core/src/components/Input/Input.tsx
```diff
@@ -168,6 +168,9 @@ export class Input extends React.Component<InputProps, InputState> {
   focus() { this.input.focus(); }
   blur() { this.input.blur(); }
   select() { this.input.select(); }
+  getSelectionStart() { return this.input.selectionStart; }
```
argshook: i wouldn't add these here and just access properties directly where needed

packages/wix-ui-core/src/components/TimePicker/Tickers.tsx
guybenron: So that the input won't lose focus if a ticker is clicked

packages/wix-ui-core/src/components/Input/Input.tsx
guybenron: There is no outside access to the <input/> component, so unless you want to expose it, specific attributes must be accessed via getters/setters

packages/wix-ui-core/src/components/TimePicker/TimePicker.tsx
guybenron: Yes but I don't think it warrants a rewrite?

PR #665, opened 2018-07-31
wix-ui-core: Popover - update internal structure

lbelinsk:


hepiyellow:
Looks ok. But why?

lbelinsk:
Long discussion. the tooltip styles API is wrong. I am going to push some change in the tooltip bo, not really need a review for it, consulted with Shlomi

PR #670, opened 2018-08-01
Revert usage of withFocusable hoc inside of the button

Faradey27:


hepiyellow:
What was the reason for the revert?

PR #679, opened 2018-08-06
AddressInput - Marking value as 'dirty' upon user interaction, to let…

giladartzi:
… the user reselect the current value

### DIFF
--- a/packages/wix-ui-core/src/components/AddressInput/AddressInput.spec.tsx
+++ b/packages/wix-ui-core/src/components/AddressInput/AddressInput.spec.tsx
@@ -257,7 +257,7 @@ describe('AddressInput', () => {
             expect(driver.getValue()).toBe('n');
         });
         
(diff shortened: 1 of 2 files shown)


shlomitc:
As discussed, there are few issues:

1. re-selecting an option is not possible. to reproduce:
- type `a`. 
- pick the first result.
- clear and type `a` again.
- pick the first result.
- `_onSelect` is not being called internally thus value isn't updated

This should probably be fixed by letting `inputWithOptions` know that the selection should be cleared right after picking the value.

2. story bug. the `onSelect` method get's wrong value
- type `a`. 
- pick the first result.
- clear and type `a` again.
- pick the second result.
- `onSelect` is called but the `value.address.formatted` is from the first result (`1 East Broadway, New York, NY, USA` instead of `114 N 6th St, Brooklyn, NY 11249, USA`)

PR #681, opened 2018-08-07
EllipsedTooltip - fix text overflow issues

yardnsm:
This PR fixes an issue with the ellipsed tooltip: when using the ellipsed tooltip within a container with `display: flex`, the ellipsed test would overflow out of the container.

screen shot 2018-08-07 at 15 08 50


packages/wix-ui-core/src/hocs/EllipsedTooltip/withEllipsedTooltip.tsx
```diff
@@ -10,7 +10,8 @@ import debounce = require('lodash/debounce');
 
 type EllipsedTooltipProps = {
   component: React.ReactElement<any>,
-  showTooltip?: boolean
+  showTooltip?: boolean,
+  [propName: string]: any
```
lbelinsk: why?
Can't you just add the style?

PR #686, opened 2018-08-13
Adding open() and close() methods to Dropdown and InputWithOptions. A…

giladartzi:
…dding close() to AddressInput


packages/wix-ui-core/src/baseComponents/InputWithOptions/InputWithOptions.tsx
```diff
@@ -108,6 +109,14 @@ export class InputWithOptions extends React.PureComponent<InputWithOptionsProps>
     this._onKeyDown = this._onKeyDown.bind(this);
   }
 
+  open() {
+    this.dropDownRef.getInstance().open();
```
lbelinsk: I would just add a comment here that `getInstance` is here because `closeOutside` hoc

PR #687, opened 2018-08-14
DO NOT MERGE!!! Moved components from baseComponent folder to component folder

RashelleKosib:


packages/wix-ui-core/src/components/NavStepper/NavStep.tsx
```diff
@@ -1,6 +1,6 @@
 import * as React from 'react';
 import style from './NavStep.st.css';
-import {StepProps} from '../../baseComponents/Stepper';
+import {StepProps} from '../../components/Stepper';
```
JoMarton: you can have a shorter path here

shlomitc:
What is the strategy? keeping backward compatibility? releasing a major version?
because this current solution will break all other libraries.
also - let's not forget all the consuming libraries that import `.st.css` files from dist

JoMarton:
@shlomitc you're totally correct. We incorrectly assumed that `baseComponents` were only used internally by comps from `components`.
Given that isn't the case we'll move to map the direct import from dist in depending libs and try to fix those before changing directories

PR #691, opened 2018-08-16
wix-ui-core - migrate to @stylable/dom-test-kit

yardnsm:


packages/wix-ui-core/package.json
```diff
@@ -60,6 +60,7 @@
   "devDependencies": {
     "@storybook/addon-options": "4.0.0-alpha.14",
     "@storybook/react": "4.0.0-alpha.14",
+    "@stylable/dom-test-kit": "^0.1.0",
```
shlomitc: should be dependency

packages/wix-ui-core/src/globals.d.ts
```diff
@@ -1,43 +1,32 @@
-declare module '*.st.css' {
-  const stylesheet: RuntimeStylesheet;
-  export default stylesheet;
-}
+type StateValue = boolean | number | string;
```
shlomitc: changing this file might problematic.
Other componet libraries use it in their `tsconfig.json` and rely on the data.
for example:
https://github.com/wix/wix-ui-backoffice/blob/master/tsconfig.json#L17

can you make `wix-ui-backoffice` will work with this new change?
also - was this file change required when migrating to the test-utils or you just did it on the way?

yardnsm: Yeah, this change was required when migrating to the new test-utils. @tomrav helped me with it.

Also, there is a related PR in `wix-ui-backoffice`: https://github.com/wix/wix-ui-backoffice/pull/213. These changes are required for it to pass as well.

packages/wix-ui-core/src/types/BaseProps.ts
```diff
@@ -1,6 +1,6 @@
 /**
  * Base props for any core component
  */
-export interface BaseProps extends RuntimeStylesheetBaseProps {
+export interface BaseProps extends InheritedAttributes {
```
hepiyellow: @yardnsm Why did you change this?
It opens all component's Props type to any kind of prop.
```
interface InheritedAttributes {
    className?: string;
    [props: string]: any;
}
```

PR #692, opened 2018-08-19
ui-core: dropdown will be closed on another click

eranshabi:


packages/wix-ui-core/src/baseComponents/Dropdown/Dropdown.tsx
```diff
@@ -203,7 +210,7 @@ export class DropdownComponent extends React.PureComponent<DropdownProps & Injec
         shown={forceContentElementVisibility || (isOpen && !disabled && hasContent)}
         showArrow={showArrow}
         timeout={timeout}
-        onClick={!disabled && openTrigger === CLICK ? () => this.open() : undefined}
+        onClick={!disabled && openTrigger === CLICK ? () => this.onPopoverClick() : undefined}
```
lbelinsk: Small refactor suggestion: Can drop the arrow (your call)
```
onClick={!disabled && openTrigger === CLICK ? this.onPopoverClick : undefined}
```

packages/wix-ui-core/src/components/AddressInput/AddressInput.spec.tsx
```diff
@@ -386,6 +387,7 @@ describe('AddressInput', () => {
             GoogleMapsClientStub.setAddresses([helper.ADDRESS_2]);
             const secondRequest = GoogleMapsClientStub.setGeocodePromise(helper.GEOCODE_2);
             driver.click();
```
lbelinsk: wait, this might affect production?
In the past, there was no need to click twice in order for this behavior to take place...

lbelinsk:
This pr is changing the behavior of the Dropdown - clicking twice on the Input will close it, unlike it does today.

Don't think this is a breaking change or anything, but this is a dependency of wix-ui-santa AddressInput component. @giladartzi is this a problem by any chance?

Most of the dropdowns in the world, including the wsr one, are closing when clicking the Input twice, so that is why this pr is reasonable.

giladartzi:
fine by me @lbelinsk @eranshabi

PR #695, opened 2018-08-22
Add new internal icons: AddItemMedium, AddItemLarge, AddItemSmall

mykas:
New internal icons for AddItem component.

### DIFF
--- a/packages/wix-ui-icons-common/src/system/raw/AddItemLarge.svg
+++ b/packages/wix-ui-icons-common/src/system/raw/AddItemLarge.svg
@@ -1,0 +1,7 @@
+<svg width="63px" height="63px" viewBox="0 0 63 63" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
+    <g id="Page-1" stroke="none" stroke-width="1" fill="none" fill-rule="evenodd">
+        <g id="Add_Item_Large" fill="#3899EC" fill-rule="nonzero">
+            <polygon points="31 31 31 0 32 0 32 31 63 31 63 32 32 32 32 63 31 63 31 32 0 32 0 31"></polygon>
+        </g>
+    </g>
+</svg>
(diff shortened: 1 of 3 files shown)


alisey:
Was this merged without a review? The icon files should **not** contain IDs, class names, stroke and fill colors. Please remove them.

shlomitc:
@alisey - do you have guidelines how to add a raw SVG icon?
Your comment makes sense but there is no way to verify it

alisey:
We talked about this, but I don't think I ever documented it. Here's a relevant snippet from the fed-infra channel:

> IDs are supposed to be unique on the page, so if multiple SVG files define the same ID it’s going to cause problems. Monochrome SVGs don’t need ids, defs, masks. They should contain only paths and shapes. It’s just a matter of exporting SVGs correctly from Illustrator/Sketch/Figma. Merge all layers into one, and apply the masks, which will result in an SVG with a single path. If you know who designed the icons you can ask them to export the icons correctly, otherwise Domas could help I think.

shlomitc:
@milkyfruit - I suggest we should align according to Alexey's suggestion.
@alisey - is this a good example for an svg? https://github.com/wix/wix-ui/blob/master/packages/wix-ui-icons-common/src/general/raw/Chat.svg

alisey:
@shlomitc - yes.

argshook:
@alisey @shlomitc we have `svg2react-icon` which internally uses `svgo` which optimizes icons and we don't need to care about it ourselves. this is ok there's no need to revert anything

alisey:
@argshook in theory yes, but in practice no, here's what AddItemLarge compiles to in `wix-ui-icons-common`:

```
React.createElement("svg",
  {
    viewBox: "0 0 63 63",
    fill: "currentColor",
    width: size || "63",
    height: size || "63"
  },
  React.createElement("g",
    {
      id: "Page-1",
      fillRule: "evenodd"
    },
    React.createElement("path",
      {
        d: "M31 31V0h1v31h31v1H32v31h-1V32H0v-1z",
        id: "Add_Item_Large",
        fillRule: "nonzero"
      }
    )
  )
)
```

Our SVGO configuration does not remove ID and style attributes, masks, groups. And if we added the rules to do so it would screw up some of the icons. Unfortunately SVO is boo basic / unreliable, and the designers still need to think about exporting SVGs in a way that is compatible with our use.

argshook:
@alisey i see id removal was intentionally disabled recently. i'll remove those from these icons then. why do you worry about masks and groups btw?

alisey:
1. They are unnecessary
2. Masks don't play well with SVGO
3. It shouldn't be hard for designers to export SVGs as a single path and avoid all kinds of issues

From these 3 images you can safely remove the groups and just keep the polygons.

milkyfruit:
I agree with @alisey. A designer must remove all "transform", "mirror", "border thickness" and similar attributes before handing it to our library. I will talk to Zilvinas to optimise those icons, so we can update them here.

PR #700, opened 2018-08-23
Adding accessibility a11 y for components

RashelleKosib:
Added accessibility to a variety of ui-core components. 
Fixed violations by adding aria-labels. 
Passing default aria-labels via meta files.
Added a script to the test command (dev mode).

### DIFF
--- a/packages/wix-ui-core/src/components/AddressInput/AddressInput.meta.ts
+++ b/packages/wix-ui-core/src/components/AddressInput/AddressInput.meta.ts
@@ -18,6 +18,7 @@ addressInputMeta
       apiKey: '',
       lang: 'en',
       Client,
-      onSelect: noop
+      onSelect: noop,
+      'aria-label': 'adressinput'
     }
   });
--- a/packages/wix-ui-core/src/components/AddressInput/AddressInput.meta.ts
+++ b/packages/wix-ui-core/src/components/AddressInput/AddressInput.meta.ts
@@ -19,6 +19,6 @@ addressInputMeta
       lang: 'en',
       Client,
       onSelect: noop,
-      'aria-label': 'adressinput'
+      'aria-label': 'Insert you address here'
     }
   });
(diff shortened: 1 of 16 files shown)


JoMarton:
@RashelleKosib 
This PR currently has no affect on the library since no change was made to the build process

packages/wix-ui-core/src/components/TimePicker/TimePicker.tsx
```diff
@@ -51,6 +51,9 @@ export type TimePickerProps = Pick<InputProps, 'disabled' | 'readOnly'> & {
 
   /** custom width of component. Goes into inline style so any css distance value allowed */
   style?: React.CSSProperties;
+
+  /** aria-label - Accessibility */
+  'aria-label'?: string;
```
JoMarton: `TimePicker` already supports `aria-label` through `InputProps`

packages/wix-ui-core/src/components/AddressInput/AddressInput.meta.ts
```diff
@@ -18,6 +18,7 @@ addressInputMeta
       apiKey: '',
       lang: 'en',
       Client,
-      onSelect: noop
+      onSelect: noop,
+      'aria-label': 'adressinput'
```
Zombiefruit: Not sure if relevant after Jonathan's comments, but "address" is missing a "d" here

Zombiefruit: Not sure if relevant after Jonathan's comments, but "address" is missing a "d" here

shlomitc:
is `aria-label` and other `aria-*` are mandatory for components?
I think it should be optional `?`

JoMarton:
@shlomitc 
Sure they should be optional. Is there a place here it is added as required?

shlomitc:
https://github.com/wix/wix-ui/pull/700/files#diff-7f91bea5eb002f2aa38dbd4ed389a957R180

shlomitc:
sorry, it's proptypes which is ok

RashelleKosib:
@shlomitc  it's optional as a prop, but I did gave it a default value (through meta files only) mainly because of how we test it with A11Y, as a stand-alone component.

JoMarton:
@shlomitc 
That's the propTypes syntax. to make it required it must be `string.isRequired`.
The typescript props are `"aria-label"?: string`

PR #703, opened 2018-08-27
Address input client - adding ability to use clientId

giladartzi:


lbelinsk:
looks like you have backward compatibility, so we can merge this after the build will pass

PR #711, opened 2018-08-29
fixed  Pagination and RadioButton axe-core issues

RashelleKosib:
fixed Pagination and RadioButton axe-core issues 
video is still broken


shlomitc:
thanks for the PR.
one error is left
```
[11:14:20][npm run test] 1. Video Simulation with default props - img: (Impact: minor)
[11:14:20][npm run test] Fix any of the following:
[11:14:20][npm run test]   role button  is not allowed for given element
```

JoMarton:
@shlomitc  Video is giving us problems since the error is deep in the player. We resorted to setting axe-core to the previous working version and we'll handle the issue with video guys

JoMarton:
@shlomitc  don't know how to re-run the appVeyor jobs but it should be ok

PR #715, opened 2018-08-31
Teamcity Surge Deploy failed. Temp fix.

mykas:
This is temp fix until we find out what is the reason of npx failure together with teamcity-surge-autorelease.


mykas:
Forgot. Made another pr for it.

PR #719, opened 2018-09-03
optional wrapper in protractor testkit factory

ChenRoth:
Add an optional `wrapper` when creating a testkit.
this is highly needed when there are multiple elements with the same data-hook (e.g., rows in a table) and you're aiming for a specific one.

### DIFF
--- a/packages/wix-ui-test-utils/src/protractor/protractor.ts
+++ b/packages/wix-ui-test-utils/src/protractor/protractor.ts
@@ -1,5 +1,6 @@
 import {$, ElementFinder} from 'protractor';
 
 export function protractorTestkitFactoryCreator<T> (driverFactory: (e: ElementFinder) => T) {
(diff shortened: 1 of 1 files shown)


shlomitc:
@ChenRoth - we usually recommend giving different data-hooks instead of a wrapper. so if you have a list of items, each item will have a unique data-hook. does this solve your issue?

ChenRoth:
it's a possible solution, but i don't see a reason not to allow a different approach, which happens to already exist in the enzyme kit

shlomitc:
@hepiyellow - I think you worked last time on this part, can you please take a look?

lbelinsk:
@hepiyellow any updates?

hepiyellow:
looking

hepiyellow:
Oh, yeah. I wanted to have that for a long time.
Thanks Chen !!!

PR #720, opened 2018-09-04
StylableBadge - extract layout and color mixins

Noam3kCH:
keeping backwards compatibility of BadgeStyle Mixin.

motivation - to allow state combinations of layout. for size(padding) modifications etc.
specifically for for badge size prop.


shlomitc:
We decided not to change the core component as it is a simple css class applied on a div. it can be removed after migrating other wrapping components (wix-ui-backoffice for example).

closing

PR #724, opened 2018-09-05
<ButtonNext> component

mykas:
### API
```jsx
<ButtonNext
  className="string"
  disabled={true | false}
  error={true | false}
  prefixIcon={node}
  suffixIcon={node}
  type="button | submit | reset"
  onClick={func}
>
  Click me
</ButtonNext>
```
Questions:

1. Should we have basic styles for `error` and `disabled` states?
2. Should we support `loading` state as in UX book specified?

- [x] sanity pass
- [x] unit tests pass
- [x] a11y tests pass
- [x] lint pass


packages/wix-ui-core/src/components/Button-next/button-next.protractor.driver.ts
```diff
@@ -0,0 +1,25 @@
+import {ElementFinder} from 'protractor';
+import {DriverFactory, BaseDriver} from '../../common/BaseDriver.protractor';
+
+export interface ButtonNextDriver extends BaseDriver {
+  /** returns true if the root element is present */
+  exists: () => Promise<boolean>;
+  /** returns the Button's text content */
+  getButtonTextContent: () => Promise<string>;
```
argshook: it's a button driver, so `button` in method name is redundant imo:
`getTextContent`, `isDisabled`

packages/wix-ui-core/src/components/Button-next/button-next.protractor.driver.ts
```diff
@@ -0,0 +1,25 @@
...
+  /** checks wether the button is disabled */
+  isButtonDisabled: () => Promise<boolean>;
+}
+
+export const buttonNextDriverFactory: DriverFactory<ButtonNextDriver> = element => ({
+  element: () => element,
+  exists: async () => element.isPresent(),
+  getButtonTextContent: async () => element.getText(),
+  click: async () => element.click(),
+  isButtonDisabled: () => hasAttribute(element, 'disabled'),
```
argshook: should also be async?

packages/wix-ui-core/src/components/Button-next/button-next.tsx
```diff
@@ -0,0 +1,71 @@
...
+      prefixIcon,
+      children,
+      onClick,
+      disabled,
+      error,
+      ...rest
+    } = this.props;
+    return (
+      <button
+        {...rest}
```
argshook: you had a question about `type` to be exposed. i  think it's okay to expose it and even by default you should say `type="button"` to avoid accidental form submission

packages/wix-ui-core/stories/Button-next/Button-next.story.tsx
```diff
@@ -0,0 +1,18 @@
+import { ButtonNext } from "../../src/components/Button-next/";
+
+export default {
+  category: "Components",
+  storyName: "ButtonNext",
+  component: ButtonNext,
+  componentPath: "../../src/components/Button-next/button-next.tsx",
```
argshook: noticed here `Button-next` with capital `B`. is it okay according to latest convention?

packages/wix-ui-core/src/components/Button-next/button-next.protractor.driver.ts
Guar1: I think so.

packages/wix-ui-core/stories/Button-next/Button-next.story.tsx
Guar1: camelCase is better

packages/wix-ui-core/src/components/Button-next/button-next.driver.ts
```diff
@@ -0,0 +1,20 @@
+import {
+  ComponentFactory,
+  DriverFactory
+} from "wix-ui-test-utils/driver-factory";
+
+export interface ButtonDriver {
+  exists: () => boolean;
+  /** click on the button root element */
+  click: () => any;
```
lbelinsk: I think 
click: () => void;
Is more accurate

packages/wix-ui-core/src/components/Button-next/button-next.driver.ts
```diff
@@ -0,0 +1,20 @@
+import {
+  ComponentFactory,
+  DriverFactory
+} from "wix-ui-test-utils/driver-factory";
+
+export interface ButtonDriver {
+  exists: () => boolean;
```
lbelinsk: Better to extend BaseDriver which already gives you `exsits`.
https://github.com/wix/wix-ui/blob/585303d2988035e0e5b7ab0e8d90bd3d70d64e5f/packages/wix-ui-test-utils/src/driver-factory/createDriverFactory.tsx#L10

packages/wix-ui-core/src/components/Button-next/button-next.driver.ts
```diff
@@ -0,0 +1,20 @@
...
+  /** click on the button root element */
```
lbelinsk: Let's keep the description high level for end users - `click on the button`

packages/wix-ui-core/src/components/Button-next/button-next.protractor.driver.ts
lbelinsk: We don't have `getTextContent ` in the unit driver, or isDisabled, so better to add them.

It is a good question to ask what should `getTextContent` should return...

mykas: Left it intentionally. Someone already might be using this API.

packages/wix-ui-core/src/components/Button-next/button-next.tsx
mykas: Thanks, added type prop from React.ButtonHTMLAttributes with default value button.

packages/wix-ui-core/src/components/Button-next/button-next.protractor.driver.ts
```diff
@@ -0,0 +1,25 @@
...
+});
+
+// TODO: use this from wix-ui-test-utils, once it is added there
```
lbelinsk: You can add it :)

packages/wix-ui-core/stories/index.tsx
```diff
@@ -1,43 +1,44 @@
...
-import {storiesOf} from '@storybook/react';
-import {PaginationStory} from './Pagination/pagination-story';
-import {DividerStory} from './Divider/divider-story';
-import {GoogleMapsIframeClientStory} from './clients/GoogleMapsIframeClient-story';
-import {CheckboxStory} from './Checkbox/Checkbox-story';
-import {TooltipStory} from './Tooltip/custom';
-import {RadioButtonStory} from './RadioButton';
-import EllipsedText from './EllipsedTooltip';
-import Focusable from './Focusable';
+import * as React from "react";
```
lbelinsk: Our linting is shit. once we will update yoshi we will get the correct linting. But in general, we are using ' and not "...
Not critical.

packages/wix-ui-core/src/components/Button-next/button-next.spec.tsx
```diff
@@ -0,0 +1,40 @@
...
+describe("ButtonNext", () => {
+  const createDriver = new ReactDOMTestContainer()
+    .unmountAfterEachTest()
+    .createLegacyRenderer(buttonDriverFactory);
+
+  const createPrivateDriver = new ReactDOMTestContainer()
+    .unmountAfterEachTest()
+    .createLegacyRenderer(buttonPrivateDriverFactory);
+
+  describe("onClick prop", () => {
```
lbelinsk: Missing test - should not call onClick when button is disabled

packages/wix-ui-core/src/components/Button-next/button-next.spec.tsx
```diff
@@ -0,0 +1,40 @@
...
+      expect(driver.suffixExists()).toBeTruthy();
+      expect(driver.prefixExists()).toBeFalsy();
+    });
+
+    it("should render prefix when given", () => {
+      const driver = createPrivateDriver(<ButtonNext prefixIcon={prefix} />);
+      expect(driver.prefixExists()).toBeTruthy();
+      expect(driver.suffixExists()).toBeFalsy();
+    });
+  });
```
lbelinsk: Missing test for testkits

packages/wix-ui-core/src/components/Button-next/button-next.st.css
```diff
@@ -0,0 +1,14 @@
+@namespace "core-button";
+
+.root {
+  -st-states: error, disabled;
```
lbelinsk: disabled is a native pseudo selector. I believe you can drop it, like in the poc.

lbelinsk: Also, you can specify the default cursor for disabled button

packages/wix-ui-core/stories/Button-next/Button-next.story.tsx
mykas: @argshook Yep.

packages/wix-ui-core/src/components/Button-next/button-next.tsx
```diff
@@ -0,0 +1,76 @@
...
+import { BaseProps } from "../../types/BaseProps";
+import style from "./button-next.st.css";
+
+export interface ButtonProps
+  extends BaseProps,
+    React.ButtonHTMLAttributes<any> {
+  prefixIcon?: React.ReactElement<any>;
+  suffixIcon?: React.ReactElement<any>;
+  onClick?: React.EventHandler<React.MouseEvent<HTMLButtonElement>>;
+  disabled?: boolean;
```
lbelinsk: disabled and onClick are native ones, you don't need to specify it manually

packages/wix-ui-core/src/components/Button-next/button-next.tsx
```diff
@@ -0,0 +1,76 @@
...
+    type: 'button'
+  };
+
+  static propTypes = {
+    className: string,
+    disabled: bool,
+    error: bool,
+    prefixIcon: node,
+    suffixIcon: node,
+      /** Type of the button - submit / button / reset */
```
lbelinsk: No need for the comment imo

packages/wix-ui-core/stories/Button-next/Button-next.story.tsx
mykas: @Guar1 Hey, even though it feels like camelCase looks better, we want solve an issue with git in windows ( git cares about casing, windows doesn’t).

packages/wix-ui-core/src/components/Button-next/button-next.tsx
```diff
@@ -0,0 +1,76 @@
...
+      children,
+      onClick,
+      disabled,
+      error,
+      type,
+      ...rest
+    } = this.props;
+    return (
+      <button
+        {...rest}
```
lbelinsk: rest can include `type`

packages/wix-ui-core/src/components/Button-next/meta-components/Icon.tsx
```diff
@@ -0,0 +1,12 @@
+import * as React from "react";
```
lbelinsk: what is this icon and why is it here?

packages/wix-ui-core/stories/Button-next/Button-next.story.tsx
lbelinsk: since windows users have issues with git and case-sensitive letters, the guidelines specify to use kebab cased style, so it should be `button-next`

packages/wix-ui-core/src/components/Button-next/meta-components/Icon.tsx
mykas: The icon is for autotools showcase for testing purposes. Should we use it already or maybe lets wait until autotools will be ready with some kind of guidelines?

packages/wix-ui-core/stories/index.tsx
mykas: Yep, but I guess I can't change it now. Otherwise our current yoshi lint will fail?

lbelinsk:
@nadavwix

packages/wix-ui-core/src/components/Button-next/button-next.driver.ts
mykas: Yep, changed it

packages/wix-ui-core/src/components/Button-next/button-next.driver.ts
mykas: Done

packages/wix-ui-core/src/components/Button-next/button-next.driver.ts
mykas: done

packages/wix-ui-core/src/components/Button-next/button-next.protractor.driver.ts
mykas: done

packages/wix-ui-core/src/components/Button-next/button-next.spec.tsx
mykas: done

packages/wix-ui-core/src/components/Button-next/button-next.spec.tsx
mykas: done

packages/wix-ui-core/src/components/Button-next/button-next.st.css
mykas: done

packages/wix-ui-core/src/components/Button-next/button-next.tsx
mykas: done

packages/wix-ui-core/src/components/Button-next/button-next.tsx
mykas: done

packages/wix-ui-core/src/components/Button-next/button-next.tsx
mykas: what do you mean? I think if you specify rest at the end of object destructuring it will not include `type` if its already destructured

packages/wix-ui-core/src/components/button-next/button-next.protractor.driver.ts
```diff
@@ -0,0 +1,23 @@
+import { DriverFactory, BaseDriver } from "../../common/BaseDriver.protractor";
```
lbelinsk: this will be removed when the unidriver will be updated..

packages/wix-ui-core/src/components/button-next/button-next.spec.tsx
```diff
@@ -0,0 +1,69 @@
...
+    });
+  });
+
+  describe(`'suffixIcon' and 'prefixIcon' props`, () => {
+    const suffix = <div>suffix</div>;
+    const prefix = <div>prefix</div>;
+
+    it(`should render 'suffix' when given`, async () => {
+      const driver = createPrivateDriver(<ButtonNext suffixIcon={suffix} />);
+      expect(await driver.suffixExists()).toBeTruthy();
```
lbelinsk: Can also check that the actual prefix or suffix component is rendered by providing a data-hook on them and then expecting for the datahook to be on the dom

packages/wix-ui-core/src/components/button-next/button-next.st.css
```diff
@@ -0,0 +1,18 @@
+@namespace "core-button";
```
lbelinsk: let's not use this for now. If we will decide to use it, we should do it for all the components.

packages/wix-ui-core/src/components/button-next/button-next.tsx
```diff
@@ -0,0 +1,72 @@
...
+    prefixIcon: node,
+    suffixIcon: node,
+    type: oneOf(["submit", "button", "reset"])
+  };
+
+  addPrefix = prefixIcon =>
+    prefixIcon &&
+    React.cloneElement(prefixIcon, {
+      className: style.prefix,
+      "data-hook": "prefix"
```
lbelinsk: Since there is no wrapping div, you are blocking the user from passing data-hook on the prefix. Overriding it might be annoying for the user when he write his tests, since he is expecting for this data-hook to be on the dom.

why do you need this custom data-hook? for the driver?
Either don't add data-hook, or wrap with extra div/span

packages/wix-ui-core/src/components/button-next/button-next.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+  render() {
+    const {
+      suffixIcon,
+      prefixIcon,
+      children,
+      onClick,
+      disabled,
+      error,
+      type,
```
lbelinsk: you don't need to destructure type. it will be past as `...rest`

packages/wix-ui-core/src/components/button-next/meta-components/Icon.tsx
```diff
@@ -0,0 +1,12 @@
+import * as React from "react";
```
lbelinsk: why isn't this icon in wix-ui-icons-common?

packages/wix-ui-core/src/testkit/enzyme.ts
```diff
@@ -61,6 +67,15 @@ export const buttonTestkitFactory = enzymeTestkitFactoryCreator<ButtonDriver>(
 );
 export { ButtonDriver };
 
+import {
```
lbelinsk: For sanity, check that as a consumer you are able to use the enzyme + react-test-utils testkit with this new unidriver implementation

packages/wix-ui-core/stories/button-next/button-next.story.tsx
```diff
@@ -0,0 +1,18 @@
+import { ButtonNext } from "../../src/components/button-next/";
```
lbelinsk: According to the new contribution guide, we don't need to create a story, or at least not in a seperate place.
Please write what are the issues that still blocks you from doing this here, and we will talk about it.

packages/wix-ui-core/src/components/button-next/button-next.protractor.driver.ts
Guar1: **

packages/wix-ui-core/package.json
```diff
@@ -95,6 +95,7 @@
     "tslint-config-prettier": "^1.12.0",
     "tslint-react": "^3.5.1",
     "typescript": "2.9.2",
+    "unidriver": "^1.0.0-alpha",
```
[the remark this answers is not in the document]
mykas: ok

packages/wix-ui-core/src/components/button-next/button-next.driver.ts
```diff
@@ -0,0 +1,18 @@
+import { UniDriver } from "unidriver";
+
+export interface ButtonNextDriver {
+  /** returns true if button exists */
+  exists: () => Promise<boolean>;
```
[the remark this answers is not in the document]
mykas: I  think first we need to upgrade `BaseDriver` with async await values. Otherwise it complains about type mismatch (`Promise<boolean>` with `boolean`)

packages/wix-ui-core/src/components/button-next/button-next.driver.ts
```diff
@@ -0,0 +1,18 @@
...
+  /** click on the button */
+  click: () => Promise<void>;
+  /** returns button text */
+  getTextContent: () => Promise<string>;
+}
+
+export const buttonNextDriver = (base: UniDriver): ButtonNextDriver => {
+  return {
+    exists: async () => await base.$("button").exists(),
```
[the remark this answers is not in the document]
mykas: In order to achieve this we need to use driverfactory to pass data hook somehow. This is now blocked until we resolve dataHook handler.

packages/wix-ui-core/src/components/button-next/button-next.spec.tsx
mykas: true

packages/wix-ui-core/src/components/button-next/button-next.tsx
mykas: true, fixed.

packages/wix-ui-core/src/components/button-next/button-next.tsx
mykas: kk

packages/wix-ui-core/stories/button-next/button-next.story.tsx
mykas: Hm, then I guess I need to import stories from components.

packages/wix-ui-core/src/testkit/enzyme.ts
mykas: Blocked `enzyme` adapter for unidriver.

lbelinsk:
@mykas the testkit tests are working now. they are not using the one liner `runTestkitExistsSuite`, but instead it is more explicitly right now. It should be really easy to add this, but I wrote it to myself. That will be added in a separated pr.

packages/wix-ui-core/src/components/button-next/button-next.driver.ts
```diff
@@ -0,0 +1,20 @@
...
+export interface ButtonNextDriver extends BaseUniDriver {
+  /** returns button text */
+  getButtonTextContent: () => Promise<string>;
+  /** returns true if button disabled */
+  isButtonDisabled: () => Promise<any>;
+}
+
+export const buttonNextDriverFactory = (base: UniDriver): ButtonNextDriver => {
+  return {
+    ...BaseUniDriverFactory(base),
```
lbelinsk: should not start with capital letter, it is a function

packages/wix-ui-core/src/components/button-next/button-next.spec.tsx
```diff
@@ -0,0 +1,70 @@
...
+import { ButtonNext } from "./";
+import { isUniTestkitExists } from 'wix-ui-test-utils/vanilla';
+import { isUniEnzymeTestkitExists } from 'wix-ui-test-utils/enzyme';
+import { mount } from 'enzyme';
+import { buttonNextPrivateDriverFactory } from './button-next.driver.private';
+
+describe("ButtonNext", () => {
+  const createDriver = new ReactDOMTestContainer()
+    .unmountAfterEachTest()
+    .createUniRenderer(buttonNextPrivateDriverFactory);
```
lbelinsk: @nadavov @JoMarton here we provide the private driver because we wish to have all the private methods internally.
This driver is not going to be exposed for the consumers, as their testkit will be constructed from the public driver.

packages/wix-ui-core/src/components/button-next/button-next.spec.tsx
```diff
@@ -0,0 +1,70 @@
...
+    });
+  });
+
+  describe(`'suffixIcon' and 'prefixIcon' props`, () => {
+    const suffix = <div data-hook="suffix">suffix</div>;
+    const prefix = <div data-hook="prefix">prefix</div>;
+
+    it(`should render 'suffix' when given`, async () => {
+      const driver = createDriver(<ButtonNext suffixIcon={suffix} />);
+      expect(await driver.suffixExists()).toBeTruthy();
```
lbelinsk: @nadavov see here for example, `suffixExists` is a private method

packages/wix-ui-core/src/components/button-next/button-next.spec.tsx
```diff
@@ -0,0 +1,70 @@
...
+    });
+
+    it(`should render 'prefix' when given`, async () => {
+      const driver = createDriver(<ButtonNext prefixIcon={prefix} />);
+      expect(await driver.prefixExists()).toBeTruthy();
+      expect(await driver.suffixExists()).toBeFalsy();
+    });
+  });
+
+  describe('testkits', () => {
```
lbelinsk: This is something worth discussing. We should automate this kind of test, that checks that the consumers will get a valid testkit.

We construct both enzyme and react-test-utils testkits out of the public unidriver, and this is a sanity test that simulates the construction of the testkit, and calling the method `exists`.

It is important to run such kind of tests automatically in some way.

packages/wix-ui-core/src/testkit/enzyme.ts
lbelinsk: Added it in wix-ui-test-utils. not blocked anymore.
Added the sanity test, so this is resolved.

packages/wix-ui-core/test/dom-test-container.ts
```diff
@@ -84,6 +87,15 @@ export class ReactDOMTestContainer {
       });
     };
   }
+
+  // Adapter for react based uni driver
+  public createUniRenderer<T>(driverFactory: (base: UniDriver) => T): (element: JSX.Element) => T {
```
lbelinsk: Even if not in this pr, we should find a better name :)

packages/wix-ui-core/src/components/button-next/button-next.meta.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+buttonMetadata.addSim({
+  title: "Simulation with suffix icon",
+  props: {
+    suffixIcon: <Add />,
+    children: "Button"
+  }
+});
+
+buttonMetadata.addSim({
```
shlomitc: @JoMarton - do these kinds of simulations makes sense?
We would like to test these as `eyes` test to verify styles, but is every prop required for the sanity ones?

packages/wix-ui-core/src/components/button-next/button-next.spec.tsx
```diff
@@ -0,0 +1,70 @@
...
+    it(`should not call 'onClick' when 'disabled'`, async () => {
+      const onClick = jest.fn();
+      const driver = createDriver(<ButtonNext onClick={onClick} disabled />);
+      await driver.click();
+      expect(onClick).not.toBeCalled();
+    });
+  });
+
+  describe(`'children' prop`, () => {
+    it("should render text", async () => {
```
shlomitc: should it accept a react node as well?

packages/wix-ui-core/src/components/button-next/button-next.spec.tsx
shlomitc: I'm not sure it should be part of these component tests.
for example - where would you test that the puppeteer driver is exposed?

packages/wix-ui-core/src/components/button-next/button-next.tsx
```diff
@@ -0,0 +1,68 @@
...
+  error?: boolean;
+}
+/**
+ * ButtonNext
+ */
+export class ButtonNext extends React.Component<ButtonProps> {
+  static displayName = "ButtonNext";
+
+  static defaultProps = {
+    disabled: false,
```
shlomitc: shouldn't we comment above the props for autodocs to read it?

packages/wix-ui-core/src/components/button-next/button-next.tsx
```diff
@@ -0,0 +1,68 @@
...
+  extends BaseProps,
+    React.ButtonHTMLAttributes<any> {
+  prefixIcon?: React.ReactElement<any>;
+  suffixIcon?: React.ReactElement<any>;
+  error?: boolean;
+}
+/**
+ * ButtonNext
+ */
+export class ButtonNext extends React.Component<ButtonProps> {
```
shlomitc: The fact that we call it `ButtonNext` is because `Button` is already taken in the file structure.
I suggest changing it to `Button` (and all other possible references).

packages/wix-ui-core/src/components/button-next/button-next.tsx
```diff
@@ -0,0 +1,68 @@
...
+  static propTypes = {
+    className: string,
+    disabled: bool,
+    error: bool,
+    prefixIcon: node,
+    suffixIcon: node,
+    type: oneOf(["submit", "button", "reset"])
+  };
+
+  addPrefix = prefixIcon =>
```
shlomitc: I'd name it `_addPrefix`, as a convention for private methods, same for the `addSuffix`

packages/wix-ui-core/stories/index.tsx
shlomitc: let's fix it.
this component should be an example, meaning it should lint perfectly

packages/wix-ui-core/src/components/button-next/button-next.spec.tsx
lbelinsk: Agree. I wrote this comment to think about what should we do. This is the way we do it today. Better to delete this entirely?
I added to my backlog to think about this.

packages/wix-ui-core/src/components/button-next/button-next.spec.tsx
lbelinsk: Yes, children can be node, but it won't happen a lot I guess since we expose 2 separate props for suffix and prefix

packages/wix-ui-core/src/components/button-next/button-next.tsx
```diff
@@ -0,0 +1,68 @@
+import * as React from "react";
+import { string, node, bool, oneOf } from "prop-types";
+import { BaseProps } from "../../types/BaseProps";
+import style from "./button-next.st.css";
+
+export interface ButtonProps
+  extends BaseProps,
+    React.ButtonHTMLAttributes<any> {
+  prefixIcon?: React.ReactElement<any>;
```
lbelinsk: autodocs comments should be here

packages/wix-ui-core/src/components/button-next/button-next.tsx
lbelinsk: we should comment the interface, not defaultProps.

packages/wix-ui-core/stories/index.tsx
lbelinsk: To be honest, I am not sure what is the "correct" linting. is it the aggressive prettier that Yoshi introduces?
If so, it is possible to use the lint rules without yoshi, just config the tslint config

packages/wix-ui-core/src/components/button-next/button-next.tsx
shlomitc: right, I thought it's the `propTypes`.
by the way - where are they?

packages/wix-ui-core/src/components/button-next/button-next.tsx
mykas: Too much wiring work. Lets stay with this until we decide to merge with the real Button

packages/wix-ui-core/src/components/button-next/button-next.tsx
lbelinsk: right here:
https://github.com/wix/wix-ui/blob/34ba6692c4cca41468a697ea576c26423c20cc03/packages/wix-ui-core/src/components/button-next/button-next.tsx#L25

lbelinsk:
let's take care of the linting and renaming of ButtonNext to Button in a separated pr.
Please add the autodoc comments and let's merge

packages/wix-ui-core/src/components/button-next/button-next.tsx
mykas: Plus autotols sanity check will fails because of multiple components with the same name

packages/wix-ui-core/src/components/button-next/button-next.driver.ts
```diff
@@ -0,0 +1,20 @@
...
+  /** returns button text */
+  getButtonTextContent: () => Promise<string>;
+  /** returns true if button disabled */
+  isButtonDisabled: () => Promise<any>;
+}
+
+export const buttonNextDriverFactory = (base: UniDriver): ButtonNextDriver => {
+  return {
+    ...baseUniDriverFactory(base),
+    getButtonTextContent: async () => await base.text(),
```
hepiyellow: @mykas Why do we need to wrap the `base.text()`, can't we simply expose it as is (and with the same name)?
Why not expose all of the base driver?

@shlomitc @lbelinsk Can you comment on this as a guide-line?

mykas: Good point. We could add it to base driver

PR #725, opened 2018-09-06
wix-ui-test-utils: allow to pass react component to testkitFactoryCreator

Faradey27:
Motivation:
```
const wrapper = ReactTestUtils.renderIntoDocument(
      <SortableListContext
        className={containerId}
        dataHook={containerId}
        containerId={containerId}
        renderItem={renderItem}
        items={items}
        onDrop={onDrop}
        />
    );
    const driver = sortableListTestkitFactory({
      wrapper,
      dataHook: containerId
    });
```
currently, such code does not work, because testkitFactoryCreator expect to receive node(to do querySelector)

so users need to do smth like this:
```
const wrapper = ReactTestUtils.renderIntoDocument(
<div>
      <SortableListContext
        className={containerId}
        dataHook={containerId}
        containerId={containerId}
        renderItem={renderItem}
        items={items}
        onDrop={onDrop}
        />
</div>
    );
    const driver = sortableListTestkitFactory({
      wrapper,
      dataHook: containerId
    });
``` 
to make it work

But in general this extra div is useless and also it prevent us from accessing SortableListContext instance

### DIFF
--- a/packages/wix-ui-test-utils/src/vanilla/vanilla.tsx
+++ b/packages/wix-ui-test-utils/src/vanilla/vanilla.tsx
@@ -1,4 +1,5 @@
 import * as React from 'react';
+import * as ReactDOM from 'react-dom';
 import * as ReactTestUtils from 'react-dom/test-utils';
 import {reactEventTrigger} from '../react-helpers';
 import {DriverFactory, BaseDriver} from '../driver-factory';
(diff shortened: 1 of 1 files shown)


lbelinsk:
@Faradey27 any updates?
I am  going to refactor some stuff in ui-test-utils so better to handle this in order to prevent comflicts

PR #734, opened 2018-09-18
WIP - UniDriver. Protractor uni driver factory creator + adapter.

mykas:
This is initial protractor driver factory creator for reuse in testkits. Adapter should not probably be here, but this is temporal adapter implementation until guys from unidriver approve this adapter pullrequest.

### DIFF
--- a/packages/wix-ui-test-utils/src/protractor/protractor-adapter.ts
+++ b/packages/wix-ui-test-utils/src/protractor/protractor-adapter.ts
@@ -1,0 +1,122 @@
+import {Locator, UniDriverList, UniDriver, MapFn} from 'unidriver';
+import {ElementFinder} from 'protractor';
+import {waitFor} from './utils';
+
+type ElementGetter = () => Promise<ElementFinder | null>;
+type ElementsGetter = () => Promise<ElementFinder[]>;
+
+export const protractorUniDriverList = (
+  elems: ElementsGetter
+): UniDriverList<ElementFinder> => {
+  const map = async <T>(fn: MapFn<T>) => {
+    const els = await elems();
+    const promises = els.map((e, i) => {
+      const bd = protractorUniDriver(() => Promise.resolve(e));
+      return fn(bd, i);
+    });
+    return Promise.all(promises);
+  };
+
+  return {
+    get: (idx: number) => {
+      const elem = async () => {
+        const els = await elems();
+        return els[idx];
+      };
+      return protractorUniDriver(elem);
+    },
+    text: async () => {
+      return map(d => d.text());
+    },
+    count: async () => {
+      const els = await elems();
+      return els.length;
+    },
+    map,
+    filter: fn => {
+      return protractorUniDriverList(async () => {
+        const els = await elems();
+
+        const results = await Promise.all(
+          els.map((e, i) => {
+            const bd = protractorUniDriver(() => Promise.resolve(e));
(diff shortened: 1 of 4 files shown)


packages/wix-ui-test-utils/src/protractor/protractor.ts
```diff
@@ -1,6 +1,23 @@
 import {$, ElementFinder} from 'protractor';
+import {protractorUniDriver} from './protractor-adapter';
+import {UniDriver} from 'unidriver';
```
lbelinsk: You need to add unidriver to the package.json

packages/wix-ui-test-utils/src/protractor/protractor.ts
```diff
@@ -1,6 +1,23 @@
...
+      ? driverFactory(obj.wrapper.$(`[data-hook='${obj.dataHook}']`))
+      : driverFactory($(`[data-hook='${obj.dataHook}']`));
+}
+
+export function protractorUniTestkitFactoryCreator<T>(
+  driverFactory: (base: UniDriver) => T
+) {
+  return (obj: { dataHook: string }) => {
+    const base = protractorUniDriver(
+      async () => await $(`[data-hook='${obj.dataHook}']`)
```
lbelinsk: I don't think this should be under async await syntax. This is a lazy element selector getter and we expect the dev who uses this to wrap it with async await, not the getter itself

lbelinsk:
@mykas looks great. I will ask Boris to review the pr in unidriver.

packages/wix-ui-test-utils/src/protractor/protractor.ts
mykas: But TS complains..

lbelinsk: ok, so leave it.

packages/wix-ui-test-utils/src/protractor/protractor-adapter.ts
```diff
@@ -0,0 +1,122 @@
+import {Locator, UniDriverList, UniDriver, MapFn} from 'unidriver';
```
hepiyellow: @mykas Is this a copy of https://github.com/wix-incubator/unidriver/blob/master/src/lib/protractor/index.ts?
Why?
Please add comments if there is a good reason.

packages/wix-ui-test-utils/src/protractor/utils.ts
```diff
@@ -0,0 +1,43 @@
...
+  const val = await fn();
+  if (!val) {
+    const now = Date.now();
+    await delay(retryDelay);
+    const delta = Date.now() - now;
+    return waitFor(fn, timeout - delta, retryDelay);
+  }
+};
+
+export const eventually = async (
```
hepiyellow: - Why don't you use wix-eventually ? (https://github.com/wix/wix-eventually/blob/master/index.js)
- Please document what this function does.

packages/wix-ui-test-utils/src/protractor/utils.ts
```diff
@@ -0,0 +1,43 @@
+const delay = (ms: number) => new Promise(res => setTimeout(res, ms));
+
+export const waitFor = async (
```
hepiyellow: Missing documentation.

PR #735, opened 2018-09-20
WIP: select

argshook:


packages/wix-ui-core/src/components/select/__mocks__/popper.js.ts
```diff
@@ -0,0 +1,12 @@
+const PopperJs = require('popper.js');
```
lbelinsk: what is das?

packages/wix-ui-core/src/components/select/index.tsx
```diff
@@ -0,0 +1,21 @@
...
+  children?: React.ReactNode;
+}
+
+export class Select extends React.PureComponent<SelectProps> {
+  render() {
+    return (
+      <Downshift>
+        {({isOpen, toggleMenu}) => (
+          <div>
+            <button onClick={() => toggleMenu()}>menu</button>
```
lbelinsk: this should be wix-ui-core/button i guess

packages/wix-ui-core/src/components/select/menu.st.css
```diff
@@ -0,0 +1,3 @@
+.root {
+  padding: 10px;
```
lbelinsk: is this something we need as reset for any themed library?
padding is usually something that each themed lib would like to take care of, and it might be annoying for them to need to override existing padding

packages/wix-ui-core/src/components/select/option.st.css
```diff
@@ -0,0 +1,21 @@
+.root {
```
lbelinsk: Most of the stuff here (the colors) should not be part of the resets layer.
I think that it will be nicer to have this in the core story example, just to apply this dummy basic colored stylesheet in order to show an example that makes sense

packages/wix-ui-core/src/components/select/option.tsx
```diff
@@ -0,0 +1,24 @@
...
+  selected?: boolean;
+  disabled?: boolean;
+  highlighted?: boolean;
+}
+
+export class Option extends React.PureComponent<Props> {
+  static displayName = 'Option';
+
+  render() {
+    const {selected, disabled} = this.props;
```
lbelinsk: can merge these two :)

packages/wix-ui-core/src/components/select/__mocks__/popper.js.ts
argshook: popper does not play well inside jsdom and so instead of trying to make it work somehow i simply mock the dependency itself.
found suggestion [here](https://github.com/FezVrasta/popper.js/issues/478#issuecomment-341506071)

packages/wix-ui-core/src/components/select/menu.st.css
argshook: not at all, im still playing with stylable

packages/wix-ui-core/src/components/select/option.st.css
argshook: yes of course, nothing here resembles final result yet

packages/wix-ui-core/src/components/select/option.tsx
argshook: you mean use one prop for selected & disabled? how would that work? selected item could also be disabled

PR #736, opened 2018-09-20
add react-test-utils unidriver infrastructure

lbelinsk:


packages/wix-ui-test-utils/src/vanilla/vanilla.tsx
```diff
@@ -11,6 +12,14 @@ export function testkitFactoryCreator<T extends BaseDriver> (driverFactory: Driv
   };
 }
 
+export function uniTestkitFactoryCreator<T extends BaseDriver> (driverFactory: (base: UniDriver) => T) {
```
mykas: What about BaseDriver? Shouldn't it exist with Promise<boolean>? I know that react does not need async await stuff but unidriver uses this for adapter: https://github.com/wix-incubator/unidriver/blob/master/src/lib/index.ts#L27

lbelinsk: yeah, the base driver should be updated.
Maybe the best thing to do, that will allow us to move fast, is to create a new uniBaseDriver and define there all the exact types we need. that way we will not break anyone. what do you say?

mykas: Good idea

packages/wix-ui-test-utils/src/driver-factory/createDriverFactory.tsx
```diff
@@ -10,6 +10,10 @@ export interface BaseDriver {
   exists: () => boolean;
 }
 
+export interface BaseUniDriver {
```
hepiyellow: @lbelinsk 
NO need for this interface, since UniDriver already defines `exists` (and also `click`)

PR #739, opened 2018-09-20
wix-ui-test-utils. Base uni driver interface and factory

mykas:


packages/wix-ui-test-utils/src/base-driver/index.ts
```diff
@@ -0,0 +1,17 @@
+import {UniDriver} from 'unidriver';
+
+export interface BaseUniDriver {
+  exists: () => Promise<boolean>;
+  click: () => Promise<void>;
+}
+
+export interface BaseDriver {
```
lbelinsk: Leave BaseDriver in the old place. We don't know who uses it, and let's not mix stuff. this file should be the new thing, without deprecated stuff

packages/wix-ui-test-utils/src/vanilla/vanilla.tsx
```diff
@@ -1,35 +1,50 @@
 import * as React from 'react';
 import * as ReactTestUtils from 'react-dom/test-utils';
 import {reactEventTrigger} from '../react-helpers';
-import {DriverFactory, BaseDriver, BaseUniDriver} from '../driver-factory';
+import {DriverFactory, BaseDriver} from '../driver-factory';
+import {BaseUniDriver} from '../base-driver';
```
lbelinsk: Only lints other than that, right?

mykas: Nope, I have imported BaseUniDriver from base-driver folder instead of driver-factory.

PR #740, opened 2018-09-20
Add isUniTestkitExists helper

lbelinsk:


hepiyellow:
Why is this in the vannila file?

Isn’t the testkit.exists() an async function?

PR #743, opened 2018-09-21
Make enzyme use BaseUniDriver from new folder

mykas:
+ comments for autodocs on base-driver


packages/wix-ui-test-utils/src/base-driver/index.ts
```diff
@@ -1,7 +1,9 @@
 import {UniDriver} from 'unidriver';
 
 export interface BaseUniDriver {
+  /** returns true if component exists */
   exists: () => Promise<boolean>;
+  /** click on the button root element */
```
lbelinsk: this is a general func, not a button specifically. 
`click on the element` is good enough

PR #749, opened 2018-09-25
wix-ui-core: Don't create package-lock.json

kupriyanenko:
If you already ignoring `package-lock.json` in `.gitignore` it's better to not create this file at all.
After each git pull you need to remove local version of `package-lock.json` before `npm install/update` to have the actual dependencies. It's confusing and you can just forget to do it.


lbelinsk:
Thanks, sounds like a good idea...

PR #759, opened 2018-09-28
VOD-3825 Social player component

mokhovyk:


packages/wix-ui-core/src/components/SocialPlayer/SocialPlayer.tsx
```diff
@@ -0,0 +1,117 @@
...
+        volume={volume}
+        muted={muted}
+        playsinline={true}
+        controls={controls}
+        onPlay={onPlay}
+        onPause={onPause}
+        onEnded={onEnd}
+        onReady={() => console.log('onReady')}
+        onStart={() => console.log('onStart')}
+        onError={e => console.log('onError', e)}
```
kupriyanenko: please remove console.log

packages/wix-ui-core/src/components/SocialPlayer/SocialPlayer.tsx
```diff
@@ -0,0 +1,117 @@
...
+  muted?: boolean;
+  controls?: boolean;
+  showTitle?: boolean;
+
+  onPlay?(): void;
+  onPause?(): void;
+  onEnd?(): void;
+
+  playerRef?(any): void;
+
```
kupriyanenko: extra line

shlomitc:
@nadavwix - what do you think about such component?

We already have a `video` component based on Wix's internal `playable` solution.
This component is intended to solve the issue of having multiple providers (such as youtube).

Here is the mock I got from the guys:
image

In my opinion, the core component should contain only a single Video component, but this PR suggests adding a new component.

@mokhovyk - please add your thoughts why this should be a new component and not part of the existing `video`.

mokhovyk:
> @mokhovyk - please add your thoughts why this should be a new component and not part of the existing `video`.

My opinions are based on different behaviors and interfaces of these components.

**Behaviors**
The Video player is a component based on Wix's internal playable solution which provides playable API. It doesn’t allow to play external link.

The Social Player is a wrapper over several social platforms API (YouTube, Facebook, Vimeo, DailyMotion, etc ) based on an external solution. 

**Interface**
_Common properties:_
autoplay, loop, muted, volume

_Different properties:_
The Video Player:
video source, theme, title, poster, logo, visibility of some controls, play button 
The Social Player:
link on video page, config of specific player props

Also, these two components have a different purpose for users, so I believe that it is not advisable to combine them into one component.

PR #766, opened 2018-10-07
Add autotools snap to the testing flow of wix-ui-core

Zombiefruit:


packages/wix-ui-core/src/components/Button/Button.meta.ts
```diff
@@ -1,11 +1,21 @@
...
-  });
\ No newline at end of file
+
+buttonMetadata.exportedFrom({
+  path: 'src/components/Button/Button',
+  exportName: 'Button',
+  baseStylePath: 'src/components/Button/Button.st.css'
+});
+
+buttonMetadata.addStyle(style, {name: 'style', path: 'src/components/Button/ButtonStyle.st.css'});
```
shlomitc: please move the `ButtonStyle.st.css` file to the `stories/Button` folder to maintain same structure across the different files (like `circularProgressBar`)

packages/wix-ui-core/src/components/Checkbox/Checkbox.meta.ts
```diff
@@ -1,7 +1,17 @@
...
 
 const checkboxMetadata = Registry.getComponentMetadata(Checkbox);
+
+checkboxMetadata.exportedFrom({
+  path: 'src/components/Checkbox/Checkbox',
+  exportName: 'Checkbox',
+  baseStylePath: 'src/components/Checkbox/Checkbox.st.css'
+});
+
+checkboxMetadata.addStyle(style, {name: 'style', path: 'src/components/Checkbox/CheckboxStyle.st.css'});
```
shlomitc: same as `Button`

packages/wix-ui-core/src/components/Button/Button.meta.ts
Zombiefruit: Sure

packages/wix-ui-core/scripts/run-snap.js
```diff
@@ -0,0 +1,12 @@
+const {exec} = require('child_process');
+
+if (!process.env.APPVEYOR) {
```
shlomitc: please change this to
```js
if (process.env.EYES_API_KEY) {
  execSync('npm run snap', { stdio: 'inherit' });
}
```

Zombiefruit:
@hepiyellow That's the hope. We need to merge another PR and then this one is next

JoMarton:
@shlomitc @Zombiefruit 
Do you have a specific single screen size? Or you want it configurable? If configurable then I suggest we hold this feature to another PR. It requires discussing the API to do that and I don't want to hold up this PR.

I would also prefer to not hold up this PR over console colors but if it affects functionality your choice

shlomitc:
@JoMarton - We currently have a fixed size, but it can be configurable.
I agree with your suggestion - making the screen size as we have today and make it configurable in another PR

PR #768, opened 2018-10-07
AddressInput - Adding events - onClick, onDoubleClick, onMouseEnter, …

giladartzi:
…onMouseLeave

### DIFF
--- a/packages/wix-ui-core/src/components/AddressInput/AddressInput.driver.ts
+++ b/packages/wix-ui-core/src/components/AddressInput/AddressInput.driver.ts
@@ -1,1 +1,19 @@
-export {inputWithOptionsDriverFactory as addressInputDriverFactory} from '../InputWithOptions/InputWithOptions.driver';
+import {inputWithOptionsDriverFactory} from '../InputWithOptions/InputWithOptions.driver';
+
+export const addressInputDriverFactory = ({element, eventTrigger}) => {
+    const inputWithOptionsDriver = inputWithOptionsDriverFactory({
+        element,
+        eventTrigger
+    });
+    
+    const input = element.querySelector('input');
+
+    const driver = {
+        clickInput: () => eventTrigger.click(input),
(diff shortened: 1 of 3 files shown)


packages/wix-ui-core/src/components/AddressInput/AddressInput.driver.ts
```diff
@@ -1 +1,19 @@
-export {inputWithOptionsDriverFactory as addressInputDriverFactory} from '../InputWithOptions/InputWithOptions.driver';
+import {inputWithOptionsDriverFactory} from '../InputWithOptions/InputWithOptions.driver';
+
+export const addressInputDriverFactory = ({element, eventTrigger}) => {
+    const inputWithOptionsDriver = inputWithOptionsDriverFactory({
+        element,
+        eventTrigger
+    });
+    
+    const input = element.querySelector('input');
```
lbelinsk: It will be nicer to select the input using the data-hook that InputWithOptions places on it:


https://github.com/wix/wix-ui/blob/master/packages/wix-ui-core/src/components/InputWithOptions/InputWithOptions.tsx#L202

packages/wix-ui-core/src/components/AddressInput/AddressInput.tsx
```diff
@@ -79,6 +79,14 @@ export interface AddressInputProps {
     id?: string;
     /** aria-label - accessibility*/
     'aria-label'?: string;
+    /** Standard input onClick callback */
+    onClick?: () => void;
+    /** Standard input onDoubleClick callback */
+    onDoubleClick?: () => void;
+    /** Standard input onMouseEnter callback */
+    onMouseEnter?: () => void;
```
lbelinsk: Is there an actual need for this? If there is something that you can do with css it is better to use the `hover` pseudo selector, but if not than it is ok to add this callback.

giladartzi: Yes, it's needed in the editor

PR #796, opened 2018-10-16
Backoffice - <ButtonNext/> - with focusable.

mykas:
Ready for review.

This pr add focus-visible effect on button, iconButton, textButton and closeButton. This also fixes FocusableHOC hoisting issues with SFC components.


packages/wix-ui-core/src/components/button-next/button-next.tsx
```diff
@@ -14,7 +15,7 @@ export interface ButtonProps
 /**
  * ButtonNext
  */
-export class ButtonNext extends React.Component<ButtonProps> {
+export class ButtonNextRaw extends React.Component<ButtonProps> {
```
lbelinsk: why this naming?

packages/wix-ui-core/src/themes/backoffice/button/button-temp.st.css
```diff
@@ -0,0 +1,423 @@
+:import {
```
lbelinsk: This is only relevant for the focus-visible rule I guess, since we have the other pr

packages/wix-ui-core/src/components/button-next/button-next.tsx
mykas: Look at this: https://github.com/wix/wix-ui/pull/796/files#diff-4935ea9c14524abde8e424905ee9254cR70

mykas: I am not sure how we should export components in general. Should it be as `default` or `named` ?

lbelinsk: export should be named, it is better with typescript...
Still not sure why you name this that way

packages/wix-ui-core/src/themes/backoffice/button/button-temp.st.css
```diff
@@ -66,6 +66,10 @@ BackofficeTheme .button:disabled {
   border-color: value(D55);
```
lbelinsk: what is button-temp?

mykas: Button-temp is button. Temp is because we have issues with stc cli naming conflicts. As soon as I am gonna improve the stc command running  - we can rename it.

shlomitc: Please take care of it as discussed

mykas: Done

PR #801, opened 2018-10-18
Avatar Component Spec

JoMarton:


packages/wix-ui-core/src/components/avatar/README.md
```diff
@@ -0,0 +1,106 @@
...
+
+    render() {
+        return (
+            <div {...style('root')}>
+                <Avatar
+                    classnName={style.avatar}
+                    name="John Doe"
+                    imgProps={{
+                        srcset="elva-fairy-320w.jpg 320w, elva-fairy-800w.jpg 800w"
+                        sizes="(max-width: 320px) 280px, 800px"
```
shlomitc: aren't sizes defined as a css class?

packages/wix-ui-core/src/components/avatar/README.md
```diff
@@ -0,0 +1,106 @@
...
+                        src="elva-fairy-800w.jpg"
+                        alt="Elva dressed as a fairy"
```
shlomitc: I think this should be implemented internally.
Avatar will always display a contact, meaning it should be the name.

packages/wix-ui-core/src/components/avatar/README.md
```diff
@@ -0,0 +1,106 @@
...
+```
+
+
+## Style API
+
+#### Subcomponents (pseudo-elements)
+
+| selector | description  | type | children pseudo-states |
+| -- | -- | -- | -- |
+| ::container | Allows styling the background |  |  |
```
shlomitc: is this the root element?

shlomitc:
@wuwa - please review that all Avatar requirements of the backoffice are met in the suggested API and behavior

packages/wix-ui-core/src/components/avatar/README.md
JoMarton: I took this as reference:
https://material-ui.com/api/avatar/

packages/wix-ui-core/src/components/avatar/README.md
```diff
@@ -0,0 +1,106 @@
...
+
+
+## API
+
+#### Component Props
+
+| name | type | defaultValue | isRequired | description |
+| -- | -- | -- | -- | -- |
+| name | string  |  |  | The name of the avatar user. Initials will be generated from the name    |
+| imgProps | string  |  |  | the source url to load image from |
```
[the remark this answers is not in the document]
JoMarton: yeah should be `object`. `string` was a remenant of having just `src`. But I do think we should allow to provide all optional img attributes

packages/wix-ui-core/src/components/avatar/README.md
```diff
@@ -0,0 +1,106 @@
...
+| icon | JSX Element |  |  | an SVG icon component  |
+| tabIndex | number | 0 |  | the tabIndex value to put on the root  |
+| disabled | boolean | false |  | If `true`, the component will not be interactive |
```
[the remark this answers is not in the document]
JoMarton: correct. should be removed

packages/wix-ui-core/src/components/avatar/README.md
```diff
@@ -0,0 +1,106 @@
...
+
+
+## Elements
+
+Elements are "container" and content, which could be classified to either "text", "image" or "icon"
+
+
+## API
+
+#### Component Props
```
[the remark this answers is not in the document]
JoMarton: I thought that the fallback for the component should be the initials. If that is the case, when should placeholder be shown?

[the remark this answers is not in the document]
JoMarton: can you define the behaviors that this generic component have out of the box? click? focus? dropdown menu?

packages/wix-ui-core/src/components/avatar/README.md
```diff
@@ -0,0 +1,106 @@
...
+export class ComponentsDemo extends React.Component<{}, {}>{
+
+    render() {
+        return (
+            <div {...style('root')}>
+                <Avatar
+                    classnName={style.avatar}
+                    name="John Doe"
+                    imgProps={{
+                        srcset="elva-fairy-320w.jpg 320w, elva-fairy-800w.jpg 800w"
```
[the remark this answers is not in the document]
JoMarton: I took this as reference:
https://material-ui.com/api/avatar/

packages/wix-ui-core/src/components/avatar/README.md
JoMarton: These are sizes for responsive images. 
https://developer.mozilla.org/en-US/docs/Learn/HTML/Multimedia_and_embedding/Responsive_images

Could be an over-complication

packages/wix-ui-core/src/components/avatar/README.md
JoMarton: I'll add that `alt` should be added/overridden with the name. Probably also make `name` prop required?

JoMarton:
@shlomitc @wuwa 
Main talking points as I see it:
* What is the fallback behavior in case image isn't loaded? The zeplin says that initials is default, so when and how does placeholder comes in?

* Should the `name --> initials` function be implemented in the component or should it expect initials as input?

* Do we know of types of "behaviors" this component should have out of the box, like doing something on click, on focus and such? If not we need to make sure this is just a naive UI component that can be easily wrapped with behaviors

packages/wix-ui-core/src/components/avatar/README.md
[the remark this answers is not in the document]
lbelinsk: Default placeholder is problematic since it is dependent on the language. Initials sounds legit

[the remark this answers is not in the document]
shlomitc: In the backoffice design system, there is a default SVG that should be presented.
One way will be displaying it as part of the CSS, but there are some pros and cons for it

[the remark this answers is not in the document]
shlomitc: @wuwa - should the Avatar be clickable?

packages/wix-ui-core/src/components/avatar/README.md
```diff
@@ -0,0 +1,108 @@
...
+| icon | JSX Element |  |  | an SVG icon component  |
+| tabIndex | number | 0 |  | the tabIndex value to put on the root  |
+
+
+
+## General Behavior
+this component will display content based on the props provided. For example, if `imgSrc` is provided (and successfully loaded), it will display an image as content. If more than one type of content prop is provided, it will first attempt to load the image then icon and lastly initials.<br>
+If no fallback exists (or no content prop provided at all) Avatar will still show the container with background styling
+
+name conversion examples:
```
shlomitc: @wuwa - is this logic fine by you as well?

packages/wix-ui-core/src/components/avatar/README.md
```diff
@@ -0,0 +1,108 @@
...
+import { AvatarIcon } from 'my-icons/AvatarIcon';
+import style from './style.st.css'; // link to Style file - see examples of style files below
+
+export class ComponentsDemo extends React.Component<{}, {}>{
+
+    render() {
+        return (
+            <div {...style('root')}>
+                <Avatar
+                    classnName={style.avatar}
```
shlomitc: `className`

packages/wix-ui-core/src/components/avatar/README.md
```diff
@@ -0,0 +1,108 @@
...
+### React Code Example
+
+**Example 1:**
+
+```jsx
+//code example goes here
+import * as React from 'react';
+import { Avatar } from 'wix-ui-core/Avatar';
+import { AvatarIcon } from 'my-icons/AvatarIcon';
+import style from './style.st.css'; // link to Style file - see examples of style files below
```
shlomitc: I think this is a bit confusing.
If these are the styles of the component, it should be defined like this:

`import {avatar} from 'wix-ui-core/themes/backoffice;`

packages/wix-ui-core/src/components/avatar/README.md
```diff
@@ -0,0 +1,108 @@
...
+
+export class ComponentsDemo extends React.Component<{}, {}>{
+
+    render() {
+        return (
+            <div {...style('root')}>
```
shlomitc: We do not expect consumers to apply it directly

packages/wix-ui-core/src/components/avatar/README.md
```diff
@@ -0,0 +1,108 @@
...
+                    classnName={style.avatar}
+                    name="John Doe"
+                    imgProps={{
+                        srcset="elva-fairy-320w.jpg 320w, elva-fairy-800w.jpg 800w"
+                        sizes="(max-width: 320px) 280px, 800px"
+                        src="elva-fairy-800w.jpg"
+                        alt="Elva dressed as a fairy"
+                    }}
+                    icon={<AvatarIcon/>}
+                    name="John Mystery Doe"
```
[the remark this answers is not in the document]
JoMarton: correct :)

packages/wix-ui-core/src/components/avatar/README.md
JoMarton: OK I'll change it to something more along this line but we need to sync with @lbelinsk about themes location, there have been some f2f discussions about this part.

In any way I think we should create a wiki/readme for verticals on how to specifically consume and style core component apart from this file.

The format of the examples in spec files needs to be very un-opinionated due to OSS.

shlomitc: @mykas can you please add to the documentation the theming guidelines from the consumer prespective? thanks

packages/wix-ui-core/src/components/avatar/README.md
JoMarton: If you find this confusing i'll remove. Just wanted to give an example of rendering this comp in another one.

packages/wix-ui-core/src/components/avatar/README.md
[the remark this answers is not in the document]
shlomitc: click should come from  a wrapping component, meaning no need to support

packages/wix-ui-core/src/components/avatar/README.md
shlomitc: You can just add an example to the contribution guide
https://github.com/wix/wix-ui/blob/master/docs/contribution-guide.md#contributing-component-style-variants

shlomitc: (not in this PR)

packages/wix-ui-core/src/components/avatar/README.md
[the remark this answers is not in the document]
shlomitc: let's start with this and once implement, see that it fit all needs

packages/wix-ui-core/src/components/avatar/README.md
shlomitc: ok for now

PR #802, opened 2018-10-18
SelectionList - technical spec

nirdom:


packages/wix-ui-core/src/components/selection-list/spec.md
```diff
@@ -0,0 +1,248 @@
...
+            <SelectionList
+                value={this.state.value}
+                focusedValue={this.state.focusedValue}
+                selectionStartValue={this.state.selectionStartValue}
+                typeAheadValue={this.state.typeAheadValue}
+                typeAhead={true}
+                onChange={this.handleChange}
+            >
+                <SelectableItems
+                    dataSource={this.props.countries}
```
hepiyellow: Is `dataSource` an array of `dataItem`?

packages/wix-ui-core/src/components/selection-list/spec.md
```diff
@@ -0,0 +1,248 @@
...
+        focusedValue: null, 
+        selectionStartValue: null, 
+        typeAheadValue: null,
+        disabledValues: null
+    };
+
+    public render() {
+        return (
+            <SelectionList
+                value={this.state.value}
```
hepiyellow: I would do the following renaming:
- `value` ->  `selectedIds`
- `onChange` -> `onSelectedIdsChange
- `focusedValue` -> `focusedId`
etc...

It much clearer. And it says it is an `id` in the documentation.

packages/wix-ui-core/src/components/selection-list/spec.md
```diff
@@ -0,0 +1,248 @@
...
+| Name | Type | Default | Required | Description |
+| -- | -- | -- | -- | -- |
+| dataItem | any | none | yes | The data item that should be rendered |
+| isSelected | boolean | none | yes | Indicates of whether the data item is selected |
+| isFocused | boolean | none | yes | Indicates of whether the data item is focused |
+| isDisabled | boolean | none | yes | Indicates of whether the data item is disabled |
+| onRequestAction | (action: DomEvent \| SelectionListAction) => void | none | yes | A callback that can be called by the selectable item when it wants to trigger an action in the selection list. This callback receives a single parameter that can be one of the following: **1.** a DomEvent - the selection list will decide how to change its state based on the DomEvent data (e.g. if *CTRL* key is pressed, the selection list will add the item to the selection). **2.** One of the SelectionListAction enum values - the item will instruct the selection list how to change its state.  |
+
+#### SelectionListAction
+
```
hepiyellow: Should we add `selectAll` and `deselectAll`?

hepiyellow:
## Multiple <SelectableItems/>

### The issue

Could there be more than one `< SelectableItems/>` in one `<SelectionList/>`?
Even if the IDs would not clash between the 2 lists,
still the `selectionStartValue` feature would be ambivalent, since there is no order between the 2 lists (which list is first, and they don't know about each other).

### Suggestion

- Maybe we should move the `dataSource` to `<SelectionList/>`?
OR
- Maybe we should move the `selectionStartValue ` to `<SelectableItems/>`?

packages/wix-ui-core/src/components/selection-list/spec.md
```diff
@@ -0,0 +1,248 @@
...
+| Key | Description |
+| -- | -- |
+| Select | Select the item and clear the rest of the selection (similar to Click) |
```
hepiyellow: ## `selectionStartValue`
Does `Select` sets the `selectionStartValue` to the selected item?

hepiyellow:
@nirdom I think lots to discuss. I scheduled a meeting with you on Sunday. It will be more efficient that way.

packages/wix-ui-core/src/components/selection-list/spec.md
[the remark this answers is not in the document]
nirdom: In the suggested spec, when multiple=true, performing an action on an item (clicking/tapping/etc.) should select the item and clear the rest of the selection, rather than toggle the item selection (behavior that's similar to one in files commander).
When using SelectionList as a "toggle group", performing an action on an item should toggle its selection. This can be achieved by triggering `AddToSelection` / `RemoveFromSelection` actions based on the value of `isSelected`. I agree that for convenience a `ToggleSelection` action can be added and used instead of this logic.

packages/wix-ui-core/src/components/selection-list/spec.md
JoMarton: @hepiyellow  @nirdom 
Since this is an advanced replacement for `<select>` and `<option>` tags, lets try and keep naming as close to native as possible.
IMO `value`, `onChange` are ok. 

`focusedValue`, `disabledValues` I'm not sure are needed - maybe they should be provided by selectionList children, like <option disabled/>

don't currently understand what `typeAheadValue` does so can't comment

packages/wix-ui-core/src/components/selection-list/spec.md
```diff
@@ -0,0 +1,248 @@
...
+```
+
+## Components Contracts
+
+### SelectionList
+
+| Name | Type | Default | Required | Description |
+| -- | -- | -- | -- | -- |
+| value | string \| Array\<string\> \| Symbol \| Array\<Symbol\> | null | no | id/s of the selected item/s |
+| focusedValue | string | null | no | id of the focused item |
```
JoMarton: can `focusedValue` and `disabledValues` be inferred from the children's `isDisabled` and `isFocused`

packages/wix-ui-core/src/components/selection-list/spec.md
```diff
@@ -0,0 +1,248 @@
...
+| selectionStartValue | string | null | no | id of the item that is used for a anchor node in multiple selection when the *shift* key is pressed |
```
JoMarton: Didn't understand what this does

packages/wix-ui-core/src/components/selection-list/spec.md
```diff
@@ -0,0 +1,248 @@
...
+
+### SelectableItems
+
+| Name | Type | Default | Required | Description |
+| -- | -- | -- | -- | -- |
+| dataSource | any \| Array\<any\> | null | no | the selectable items data items |
+| idFunction | (dataItem: any) => string | none | yes | A function that extracts the ID from the data item |
+| typeAheadTextFunction | (dataItem: any) => string | none | yes | A function that extracts the type ahead text from the data item |
+| renderItem | React.SFC\<SelectableItemProps\> | none | yes | A SFC that used for rendering a data item |
+| tabIndex | number | none | no | The tab index that the SelectableItems component should put on its inner focused item (if such exists). This property will be auto set by the SelectionList component and its value would be the value of the SelectionList component's tabIndex property |
```
JoMarton: If tabIndex is auto set by `<SelectList>` why should this be in the API even?

packages/wix-ui-core/src/components/selection-list/spec.md
```diff
@@ -0,0 +1,248 @@
...
+| onChange | (event: ChangeEvent) => void | NOP | no | Triggered when the state of the list is changed. The state of the list consists of selected items, focused item, selection start item, disabled items, type ahead value. These values are provided inside the event structure.
+| multiple | boolean | false | no | Whether the selection list supports single or multiple selections. When true, adds aria-multiselectable='true' to the root element.
+| orientation | enum | Vertical | no | The orientation is used mostly for assistive technologies. Changing to Horizontal will change the behavior of keyboard navigation and add an aria-orientation attribute to the root with the 'horizontal' value |
+| typeAhead | boolean | true | no | Enables keyboard type-ahead |
+| children | any | null | no | See explanation about SelectionList components hierarchy |
+| disabled | boolean | false | no | Whether the SelectionList responds to events or not |
+| readonly | boolean | false | no | Gains tab focus but user cannot change value |
+| id | string | null | no | Unique identifier, relevant for a standalone component |
```
JoMarton: Should `id` be put on `<SelectList>` or one of its children?

packages/wix-ui-core/src/components/selection-list/spec.md
```diff
@@ -0,0 +1,248 @@
...
+| name | string | null | no | Specifies the name of the input element, relevant for a standalone component |
+| tabIndex | number | null | no | Tab order of the element, relevant for a standalone component |
+| aria-label | string | null | no | aria attribute, relevant for a standalone component |
+| aria-labelledby | string | null | no | aria attribute, relevant for a standalone component |
+| aria-describedby | string | null | no | aria attribute, relevant for a standalone component |
+
```
JoMarton: Lets add `size` prop to tell how many items can be chosen in `multiple` mode

packages/wix-ui-core/src/components/selection-list/spec.md
hepiyellow: @JoMarton Native `<select>` is meant for a **single selected option**. I don't see why we should keep the same naming if our semantics are different.

JoMarton: native select has multiple mode `<select multiple/>`

packages/wix-ui-core/src/components/selection-list/spec.md
```diff
@@ -0,0 +1,248 @@
...
+## Accessibility
+
+Reference [listbox](https://www.w3.org/TR/wai-aria-practices/#Listbox) in w3 draft.
+
+### Roles
+* Root Role - listbox (identifies the focusable element that has listbox behaviors and contains listbox options)
+* Children role - option (identifies a selectable element)
+
+### Aria Attributes
```
JoMarton: also include `aria-multiselectable`
or, just say we're using the aria roles and attributes that are defined in listbox spec

packages/wix-ui-core/src/components/selection-list/spec.md
hepiyellow: Thanks @JoMarton I didn't know that.
But it seems that when `multiple` attribute exists, still the change event (`onchange`), `event.target.value` is showing only the last item select (not a list of items).
You get the selected items only on `submit`.

So to conclude, I think its a "different animal".

hepiyellow:
@nadavov @nirdom 
May I suggest that we put this spec on hold,
and write a spec for a generic `<List/>` and `<ListItem/>` first?

### The `<List/>` may contain:

- Should support disabled items
- Should support item groups
- Should support orientation (horizontal, vertical)
- a11y - (keyboard navigation, aria)
- styling API

### Use cases for `<List/>`:

- Dump list  - just for presenting element (no navigation)
- ActionList (aka MenuList, a list that has navigation and item action)
  - click or select (mouse of keyboard) may trigger an action
  - support touch
  - Input-With-Options (aka Search, Autocomplete, etc...) - Navigation should not `focus` the list items, since the focus must stay on the `<input/>`


That is, a List that has keyboard navigation (if needed).

shlomitc:
was not continued

PR #812, opened 2018-10-22
Remove prop-types from wix-ui-core

Zombiefruit:
Changes approved by @shlomitc


argshook:
what's happening, why the removal?

Zombiefruit:
Prop-types and Typescript (the most recent/future React types) are essentially incompatible. In order to keep wix-ui-core up to date and not version locked, we need to get rid of prop-types. From my understanding, they're not used anymore so it's beneficial for that reason as well.

argshook:
there may be js consumers who after this PR will no longer receive proptypes. is this intended?

Zombiefruit:
It's not intended but it is a side-effect we're aware of. I'll be going over all the liabilities of wix-ui-core and making sure that they don't reference prop-types.

mykas:
lets merge?

hepiyellow:
@Zombiefruit 
1) Can you elaborate about the conflict between propTypes and Typescript's react types? couldn't it be solved with `any` type....or something.
2) propTypes is a very convenient dev tool. Are we abandoning the JS consumers?
3) In WSR (js project) we are still supporting propTypes, and we are proxying/ extending components from wix-ui-core. now we'll have to re-implement all deleted propTypes, right?

CC : @shlomitc

Zombiefruit:
@hepiyellow 

1. Unfortunately it doesn't look to be that easy. For example, one issue is that Typescript now wants you to define every possible prop type for a component (meaning that if you have a Button component, and you extended HTMLButtonAttributes, you now need to define a prop type for every extended attribute as well).

2. It looks like it's either Typescript or Prop-types, and although there are arguments for both, I think that Typescript is more worthwhile to us.

3. Yes, most likely.

PR #815, opened 2018-10-23
removed baseComponents with dropdownOption( old backward compatibility)

RashelleKosib:


hepiyellow:
@lbelinsk 
Is this POC relevant?
Can I make a change in wix-ui-core that will break it?
https://github.com/wix-private/wix-ui-poc

lbelinsk:
You can break the poc, it is not production code, but better to fix it after to point it to the correct import path

PR #828, opened 2018-10-25
<LiveCodeExample/> - add new component

yardnsm:
This PR adds a new component to `wix-storybook-utils`: `<LiveCodeExample/>`. It uses [`react-live`](https://github.com/FormidableLabs/react-live) for the most parts.

screen shot 2018-10-25 at 15 46 33

### Usage

```jsx
// You'll need to pass a `scope` to the component. Basically its
// just an object containing all the globals (in our case: our component).
import * as wsrScope from 'wix-style-react';

<LiveCodeExample
	scope={wsrScopr}
	title="Dropdown live example"
	initialCode={`
<Dropdown
	placeholder="Select dominant hand"
	options={[
		{id: 0, value: 'Left'},
		{id: 1, value: 'Right'},
		{id: 2, value: 'Ambidextrous'}
	]}
/>
	`}
/>
```

When using this component in a library, say WSR, you'd want to create a wrapper component that sets the scope for you. For example:

```jsx
// ./stories/components/LiveCodeExample
import LiveCodeExample from 'wix-storybook-utils/LiveCodeExample';
import * as scope from '../../';
export default props => (
	<LiveCodeExample scope={scope} {...props} />
);
```

#### Syntax highlighting

I noticed we don't have a syntax highlighting in out stories (maybe it's broken, or maybe its on purpose?). Since both `react-live` and the `<CodeBlock/>` uses `prism.js`, they can share the same highlighting theme. Currently the theme is scoped to the live editor only, but I think its a really hood idea that we'll have it in both of them.

#### RTL and dark background toggles

Do you think should add these toggles, like in the playground?

### DIFF
--- a/packages/wix-storybook-utils/src/LiveCodeExample/index.js
+++ b/packages/wix-storybook-utils/src/LiveCodeExample/index.js
@@ -13,10 +13,6 @@ import Code from 'wix-ui-icons-common/Code';
 import Document from 'wix-ui-icons-common/Document';
 import TextButton from '../TextButton';
 
-const LiveCodeExamplesRow = props => (
-  <div className={styles.examplesContainer} {...props}/>
-);
-
 export default class LiveCodeExample extends Component {
 
   static propTypes = {
@@ -29,8 +25,6 @@ export default class LiveCodeExample extends Component {
   static defaultProps = {
     compact: false
   };
-
-  static Row = LiveCodeExamplesRow;
 
   resetCode = () => {
     this.setState({
--- a/packages/wix-storybook-utils/src/LiveCodeExample/index.js
+++ b/packages/wix-storybook-utils/src/LiveCodeExample/index.js
@@ -1,8 +1,10 @@
 import React, {Component} from 'react';
 import PropTypes from 'prop-types';
 import {LiveProvider, LiveEditor, LiveError, LivePreview} from 'react-live';
+import classnames from 'classnames';
 import styles from './index.scss';
 
+import ToggleSwitch from 'wix-style-react/ToggleSwitch';
 import EmptyTrash from 'wix-ui-icons-common/EmptyTrash';
 import TextButton from '../TextButton';
 
@@ -26,15 +28,22 @@ export default class LiveCodeExample extends Component {
     });
   };
 
+  onToggleRtl = isRtl => this.setState({isRtl});
+  onToggleBackground = isDarkBackground => this.setState({isDarkBackground});
+
   constructor(props) {
     super(props);
 
     this.state = {
-      code: props.initialCode
+      code: props.initialCode,
+      isRtl: false,
+      isDarkBackground: false
     };
   }
 
   render() {
+    const {code, isRtl, isDarkBackground} = this.state;
+
     return (
       <div className={styles.wrapper}>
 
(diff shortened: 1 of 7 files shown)


yardnsm:
Updated styles:

screen shot 2018-10-25 at 16 47 09

yardnsm:
### Added compact mode

In compact mode, the code is stacked above the preview and collapsed by default. In addition, the RTL and the dark bg toggles are hidden.

You can use `<LiveCodeExample.Row/>` to hold multiple compact live examples.

screen shot 2018-10-28 at 16 29 13

yardnsm:
@argshook this new component uses `<ToggleSwitch/>` and `<Notification/>` from WSR, if there is an alternative please let me know or take this under consideration in your PR.

packages/wix-storybook-utils/src/LiveCodeExample/index.js
```diff
@@ -0,0 +1,102 @@
...
+
+        <LiveProvider code={code.trim()} scope={this.props.scope} mountStylesheet={false}>
+          <div className={styles.liveExampleWrapper}>
+
+            <div className={styles.editor}>
+              <LiveEditor className={styles.editorView} onChange={this.onCodeChange}/>
+            </div>
+
+            <div
+              className={classnames({
```
argshook: for those that are always `true`, just pass them
```
className={classnames(
  styles.preview,
  {
    rtl: isRtl,
    [styles.darkPreview]: isDarkBackground
  }
)}
```

packages/wix-storybook-utils/src/LiveCodeExample/index.scss
```diff
@@ -0,0 +1,184 @@
...
+  .preview {
+    box-sizing: border-box;
+    width: 50%;
+    position: relative;
+  }
+
+  .editor {
+    border-right: 1px solid #DFE5EB;
+
+    .editorView {
```
argshook: im pretty sure there's no need for nesting any classes, but you can ignore it :)

packages/wix-storybook-utils/stories/index.story.js
```diff
@@ -1,12 +1,16 @@
 import React from 'react';
 import Component from './component';
 import CodeShowcase from '../src/CodeShowcase/';
+import LiveCodeExample from '../src/LiveCodeExample';
+
+// Taking the whole wsr library for the code example test
+import * as wsrScope from 'wix-style-react';
```
argshook: yeeah, we can't have `wix-style-react` as dependency, we'll need to think of something for this

1. do a component registry?
1. prebuild wix-style-react bundle?
1. unpkg?
1. something else?

component registry is probably best option

packages/wix-storybook-utils/src/LiveCodeExample/index.js
```diff
@@ -0,0 +1,102 @@
+import React, {Component} from 'react';
+import PropTypes from 'prop-types';
+import {LiveProvider, LiveEditor, LiveError, LivePreview} from 'react-live';
+import classnames from 'classnames';
+import styles from './index.scss';
+
+import ToggleSwitch from 'wix-style-react/ToggleSwitch';
```
argshook: i will first merge https://github.com/wix/wix-ui/pull/830 then import will be `from '../ui/toggle-switch'`

packages/wix-storybook-utils/src/LiveCodeExample/index.scss
```diff
@@ -0,0 +1,184 @@
+@import '../mixins.scss';
```
argshook: maybe name it `checkerboard.scss` as there's only one mixin there

packages/wix-storybook-utils/src/LiveCodeExample/index.scss
```diff
@@ -0,0 +1,184 @@
...
+    .editorView {
+      padding: 20px;
+      min-height: 300px;
+      height: 100%;
+      overflow: auto;
+      outline: 0;
+      font-family: Consolas, Menlo, Monaco, "Andale Mono WT", "Andale Mono", "Lucida Console", "Lucida Sans Typewriter", "DejaVu Sans Mono", "Bitstream Vera Sans Mono", "Liberation Mono", "Nimbus Mono L", "Courier New", Courier, monospace;
+
+      // Theme extracted from
+      // https://github.com/PrismJS/prism-themes/blob/master/themes/prism-base16-ateliersulphurpool.light.css
```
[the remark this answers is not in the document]
yardnsm: Already tried that, and yes, I did some changes to the theme. Also, when doing this the styles will leak outside of the component. The `<CodeExample/>`s will also receive those styles.

Maybe this is a good thing? Currently the highlighting of `<CodeExample/>` is very minimal.

packages/wix-storybook-utils/src/LiveCodeExample/index.scss
yardnsm: You're right, but I feel that nesting these classes makes the intention more clear.

packages/wix-storybook-utils/src/TextButton/index.js
```diff
@@ -41,6 +45,16 @@ export default class TextButton extends Component {
         onMouseLeave={this.toggleHover}
         onClick={this.props.onClick}
       >
+        {this.props.prefixIcon ? (
+          <div
+            style={{
+              padding: '0 6px 0 0',
+              lineHeight: '0 !important'
```
argshook: oof how did i miss this `!important`, that's such a no no

packages/wix-storybook-utils/src/LiveCodeExample/index.scss
argshook: that's debatable, i'd say no use better names and what does nesting have to do with intention, why even replicate html hierarchy in css. but these don't matter, all are weak arguments.

what does matter, however, is selectors size, specificity and even bundle size. with nesting it is easy to accidentally create awkward selectors without noticing. here, compare the two:

2018-10-29-112230_screenshot

i zoomed out and some still don't fit screen. you can count up to 6 classnames for one thing
now less nested:

2018-10-29-112314_screenshot

up to 3 classnames for one thing, that's 50%

of course, minified would look better, but minifying doesn't solve specificity. you are more likely to need `!important` or similar techniques to overcome some issues (or add features) in future

so for me rule of thumb is the less selectors the better.

it doesn't really matter here for wix-storybook-utils but unfortunately this is sometimes done for widely used components too :(

PR #829, opened 2018-10-25
[WIP][autocomplete] #WCN-3330: add aria-activedescendant for VO

4alexandr:


shlomitc:
why do you need to know about hover change?

4alexandr:
For accessibility - we need to mark current(hovered) option for VoiceOver utils. 
To do it we should use aria-activedescendant attribute with value eq to ID of hovered option.

shlomitc:
Thanks for the PR. Please add missing tests

shlomitc:
It's been stale for a while now so I'm closing. please re-open if you get to it.

thanks

PR #831, opened 2018-10-28
TimePicker - Fixing 12/24h issue when useNativeInteractions is set to…

giladartzi:
… true


shlomitc:
what exactly was fixed here? how can I reproduce?

giladartzi:
There was an issue that when `useNativeInteraction` was set to `true`, and format was set to AM/PM, it still displayed the full 24h value (i.e 14:30 instead of 02:30 Pm)

PR #837, opened 2018-10-29
Avatar - spec review

hepiyellow:
`*` added in places that need dicussion

### DIFF
--- a/packages/wix-ui-core/src/components/avatar/README.md
+++ b/packages/wix-ui-core/src/components/avatar/README.md
@@ -1,6 +1,6 @@
 # Avatar
 
-Avatar is a type of element that visually represents a user, either as an image, icon or initials.
+Avatar is a type of element that visually represents a user. It renders one of the following content types: image, icon or text (initials).
 <br><br>
 ![image](./readme-assets/avatar-types.png)
 ## Elements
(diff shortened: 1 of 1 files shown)


packages/wix-ui-core/src/components/avatar/README.md
```diff
@@ -13,8 +13,8 @@ Elements are "container" and content, which could be classified to either "text"
 
 | name     | type                | defaultValue | isRequired | description                                                           |
 | -------- | ------------------- | ------------ | ---------- | --------------------------------------------------------------------- |
-| name     | string              |              | Yes        | The name of the avatar user. Initials will be generated from the name |
```
hepiyellow: - What about translations? Is it an issue? What about names with ‘-’
- What about order localization (Firs/Last or Last/First)

packages/wix-ui-core/src/components/avatar/README.md
```diff
@@ -13,8 +13,8 @@ Elements are "container" and content, which could be classified to either "text"
...
-| imgProps | Omit<HTMLImageAttributes, 'alt'> |              |            | the source url to load image from                                     |
+| name*     | string              |              | Yes        | The name of the avatar user. Initials will be generated from the name |
```
hepiyellow: - What about translations? Is it an issue? What about names with ‘-’
- What about order localization (Firs/Last or Last/First)

packages/wix-ui-core/src/components/avatar/README.md
```diff
@@ -13,10 +13,10 @@ Elements are "container" and content, which could be classified to either "text"
...
+| imgProps* | Omit<HTMLImageAttributes, 'alt'> |              |            | the source url to load image from                                     |
 | icon     | JSX Element         |              |            | an SVG icon component                                                 |
-| tabIndex | number              | 0            |            | the tabIndex value to put on the root                                 |
+| tabIndex* | number              | 0            |            | the tabIndex value to put on the root                                 |
```
hepiyellow: Maybe we should have the children prop act as an Icon or text

packages/wix-ui-core/src/components/avatar/README.md
```diff
@@ -29,9 +29,9 @@ name conversion examples:
 
 ## Technical Considerations
 
-The component will fallback to a different content prop in case the image provided didn't load. For this to happen an `onError` handler will be used on the `img` tag. If a user provided adn `onError` handler in `imgProps`, it will be called as well.<br>
+The component will fallback to a different content prop in case the image provided didn't load. For this to happen an `onError` handler will be used on the `img` tag. If a user provided an `onError` handler in `imgProps`, it will be called as well.<br>
 
-<br> The `alt` property is omitted from `imgProps` interface. `name` prop will be used as `alt` instead.<br>
+<br> The `alt` property is omitted from `imgProps` interface. Placeholder will be used as `alt` instead.<br>
```
hepiyellow: I think using Placeholder image as an alt image is more consistent, then using the name initials.

packages/wix-ui-core/src/components/avatar/README.md
```diff
@@ -5,35 +5,44 @@ Avatar is a type of element that visually represents a user, either as an image,
...
+| imgProps | Omit<HTMLImageAttributes,'alt'> |              |            | Image props, in particular image src url |
+| altContent | string \| JSXElement | | An alternate content to be shown in case the content is an image and it hasn't been loaded yet, or there was an error loading it.
+| srcSet | | | | TODO: We should consider how the consumer can provide a set of image sources for different sizes|
+| component | oneOfType([string, func]) | 'div' | | | The component used for the root node. Either a string to use a DOM element or a component. |
+| other* | | | | This is not a prop called `other`, it means that any other (e.g. `{...rest}`) props will be rendered onto the components root |
 
-## General Behavior
+## Name / title / aria-label
 
-this component will display content based on the props provided. For example, if `imgProps` is provided (and successfully loaded), it will display an image as content. If more than one type of content prop is provided, it will first attempt to load the image then icon and lastly initials.<br>
```
hepiyellow: - I think that fallback from image -> icon -> initials is a product decision.
- Also its another "magic" that need documentation.
- I think the consumer can decide by themselves the fallback logic.
- I think img is the only content that it makes sense to have a fallback , since it takes time to load and it could have an error. 

I am suggesting to have a 2 step fallback (`altContent`) and not a 3 step.

packages/wix-ui-core/src/components/avatar/README.md
```diff
@@ -93,13 +111,7 @@ export class ComponentsDemo extends React.Component<{}, {}>{
...
 }
-.avatar::text-container {
+.avatar::contentText {
   color: red;
 }
 ```
-
-## Accessibility & Keyboard Navigation
-
-The root will have `tabIndex = 0` by default meaning it will be focusable and part of the keyboard navigation flow.<br>
```
hepiyellow: I think the Avatar should not be focusable by default.
It can be wrapped with a `<a/>` or `<Link/>` component, and then it would be focusable, and would probably get an onClick handler as well.
It can also act as the target of a popover, so a `<Popover/>` wrapper would take care of tabIndex.

PR #840, opened 2018-10-29
WIP - `<Avatar/>` initial 

hepiyellow:
- minimal component
- stylesheet
- story
- minimal backoffice theme
- minimal backoffice story


packages/wix-ui-core/src/components/avatar/avatar.driver.ts
```diff
@@ -0,0 +1,23 @@
+import { UniDriver } from "unidriver";
+import {
+  BaseUniDriver,
+  baseUniDriverFactory
```
mykas: Why this `baseUniDriverFactory` is not used?

packages/wix-ui-core/src/components/avatar/avatar.driver.ts
```diff
@@ -0,0 +1,23 @@
...
+
+export interface AvatarDriver extends UniDriver {
+  /** returns text content driver */
+  getTextContentDriver: () => UniDriver;
+}
+
+export const avatarDriverFactory = (base: UniDriver): AvatarDriver => {
+  const byDataHook = dataHook => getByDataHook(base,dataHook);
+  return {
+    ...base,
```
mykas: I think it should be `...baseUniDriverFactory(base)`

packages/wix-ui-core/src/components/avatar/avatar.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+export interface AvatarProps
+  extends BaseProps {
+    /* The name of the avatar user. Initials will be generated from the name. And it will be used as default value for html `title` and `aria-label` attributes. */
+    name : string;
+}
+/**
+ * Avatar is a type of element that visually represents a user, either as an image, icon or initials.
+ */
+export const Avatar : React.SFC<AvatarProps> = props =>  {
```
mykas: You might want to wrap it already with `withFocusable`. But maybe in the next PR.

packages/wix-ui-core/src/components/avatar/avatar.driver.ts
hepiyellow: Good of you to ask.
The BaseUniDriver is hiding the UniDriver methods, why?
I want to use driver.text() dircetly. Why do I need a wrapper around it?

packages/wix-ui-core/src/components/avatar/avatar.driver.ts
hepiyellow: If we change `baseUniDriverFactory()` to extend base and not exclude it's methods then it makes sense.

hepiyellow: ok, I talked with Lior, he explained the idea. I'll fix it.

packages/wix-ui-core/src/components/avatar/avatar.tsx
hepiyellow: I think Avatar is not focusable. It ca be wrapped with a Link of Button.

packages/wix-ui-core/src/components/avatar/avatar.private.driver.ts
```diff
@@ -0,0 +1,36 @@
+import { UniDriver } from "unidriver";
```
shlomitc: why isn't there a public driver?

packages/wix-ui-core/src/components/avatar/avatar.spec.tsx
```diff
@@ -0,0 +1,52 @@
...
+describe("Avatar", () => {
+  const createDriver = new ReactDOMTestContainer()
+    .unmountAfterEachTest()
+    .createUniRenderer(avatarDriverFactory);
+
+  describe(`'name' prop`, () => {
+    it("should render generated initials as text content", async () => {
+      const driver = createDriver(<Avatar name='John Doe' />);
+      expect(await driver.getInitialsContent()).toBe('JD');
+      expect(await driver.isContentType('initials')).toBe(true);
```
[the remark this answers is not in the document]
hepiyellow: fixed

packages/wix-ui-core/src/components/avatar/avatar.st.css
```diff
@@ -0,0 +1,10 @@
+.root {
+  display: flex;
```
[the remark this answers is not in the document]
hepiyellow: Initially I thought any Avatar would be centred, but Actuallay the Backoffice placeholder is an icon which is not centered (aligned to bottom).
So I 'm moving this to default theme.

packages/wix-ui-core/src/components/avatar/avatar.private.driver.ts
hepiyellow: changed to public

packages/wix-ui-core/src/components/avatar/avatar.tsx
```diff
@@ -0,0 +1,69 @@
...
+}
+
+/**
+ * Avatar is a type of element that visually represents a user, either as an image, icon or initials.
+ * 
+ * <p>There are 3 props for corresponding content types: `initials`, `icon` and `imgProps`.
+ * If more than one of these props is supplied (with `name` prop giving default value to the `initials` prop),
+ * then the priority for display is : image -> icon -> initials.
+ */
+export const Avatar : React.SFC<AvatarProps> = props =>  {
```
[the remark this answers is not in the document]
hepiyellow: Talked to Jonatan and we agreed Avatar is not focusable. It could be wrapped with Button, Popover, whatever...

packages/wix-ui-core/src/components/avatar/avatar.meta.tsx
```diff
@@ -0,0 +1,12 @@
...
+import Registry from "@ui-autotools/registry";
+
+const avatarMetadata = Registry.getComponentMetadata(Avatar);
+
+avatarMetadata.addSim({
+  title: "default",
+  props: {
+    name: 'John Doe'
+  }
+});
```
JoMarton: I think there be additional simulations for icon and image modes

packages/wix-ui-core/src/components/avatar/avatar.private.driver.ts
JoMarton: @shlomitc what methods should this comp expose?

packages/wix-ui-core/src/themes/backoffice/avatar/avatar.st.css
```diff
@@ -0,0 +1,139 @@
...
+  background-image: value(defaultBackgroundImage);
+  color: value(D80);
+
+  height: 48px;
+  width: 48px;
+  border-radius: 30px; 
+}
+
+/* Colors */
+BackofficeTheme .avatar.colorBlue {
```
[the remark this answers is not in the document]
hepiyellow: Its true that it will work, using `BackofficeTheme .colorBlue`, But:
- The performance benefit in negligible
- Looking at a single rule, it makes no sense to have `BackofficeTheme .colorBlue`, what does it mean?

- And most important, doing so make it difficult and unexpected behaviour, when trying to apply defaults.

Say youlater want to add default color in `.avatar` class
```
.avatar {
  color: red;
} 
.colorBlue: {
  color: blue;
}
```

This means that now rule ordering decides the priority.
While if we have :
```
.avatar {
  color: red;
} 
.avatar.colorBlue: {
  color: blue;
}
```

Then the priority is set, and independent of the rule ordering.

packages/wix-ui-core/src/components/avatar/avatar.tsx
```diff
@@ -0,0 +1,76 @@
...
+        <div className={style.text} data-hook="content-text">
+          {textContent}
+        </div>
+      );
+    }
+
+    case 'icon': {
+      return React.cloneElement(props.icon, {
+        ['data-hook']: 'content-icon',
+        className:style.icon
```
JoMarton: Will this override `className` provided with the icon element? if so, is this on purpose? maybe classname the provided with this?

packages/wix-ui-core/src/components/avatar/avatar.tsx
```diff
@@ -0,0 +1,76 @@
...
+      });
+    }
+
+    case 'image': {
+      return (
+        <img 
+          {...props.imgProps}
+          data-hook="content-image"
+          className={style.image}
+        />
```
JoMarton: what about `alt`

packages/wix-ui-core/src/components/avatar/types.ts
```diff
@@ -0,0 +1 @@
+export type ContentType = 'text' | 'icon' | 'image';
```
JoMarton: I don't see the need for this to be an additional file

packages/wix-ui-core/src/components/avatar/avatar.tsx
```diff
@@ -0,0 +1,69 @@
...
+ * If more than one of these props is supplied (with `name` prop giving default value to the `initials` prop),
+ * then the priority for display is : image -> icon -> initials.
+ */
+export const Avatar : React.SFC<AvatarProps> = props =>  {
+  const { name, initials, icon, imgProps, ...rest } = props;
+  
+  const contentType: ContentType = 
+    imgProps ? 'image' :
+    icon ? 'icon' :
+    initials || name ? 'initials' :
```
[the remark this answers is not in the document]
hepiyellow: I think the trinaries are pretty clear.
And the `||` is also very clear.

packages/wix-ui-core/src/components/avatar/avatar.st.css
```diff
@@ -0,0 +1,7 @@
+.root {
```
JoMarton: I think it should have some basic styling

packages/wix-ui-core/src/components/avatar/avatar.meta.tsx
```diff
@@ -0,0 +1,12 @@
+import * as React from "react";
+import { Avatar } from "./avatar";
+import Registry from "@ui-autotools/registry";
+
+const avatarMetadata = Registry.getComponentMetadata(Avatar);
+
+avatarMetadata.addSim({
+  title: "default",
```
JoMarton: More descriptive title. It will be the title for auto generated tests so try and describe the scenario more specifically

packages/wix-ui-core/src/components/avatar/types.ts
hepiyellow: This type is used both in component, and both in driver.
Its better not to have those 2 dependent on each other.

packages/wix-ui-core/src/themes/backoffice/avatar/avatar.st.css
```diff
@@ -0,0 +1,141 @@
+:import {
+  -st-from: '../theme.st.css';
+  -st-default: BackofficeTheme;
+}
+
+:import {
+  -st-from: '../../../components/avatar/avatar.st.css';
```
JoMarton: @shlomitc @hepiyellow I'm not comfortable with direct consumption from file and not indexed export.

Currently we export all st.css files from root index.st.css but is this still what we want?

packages/wix-ui-core/src/components/avatar/avatar.tsx
hepiyellow: This the Image fallback is not yet implemented in this PR.
I'll add it later.

packages/wix-ui-core/src/components/avatar/avatar.tsx
hepiyellow: Changed to :
```
<div className={style.icon} data-hook="content-icon">
          {props.icon}
        </div>
```

packages/wix-ui-core/src/components/avatar/avatar.tsx
JoMarton: Not related to image fallback at all. `alt` is for screen readers

packages/wix-ui-core/src/components/avatar/avatar.driver.ts
```diff
@@ -0,0 +1,28 @@
...
+  getImageContent: ()=> UniDriver<HTMLImageElement>;
+}
+
+export const avatarDriverFactory = (base: UniDriver): AvatarDriver => {
+  const byDataHook = dataHook => base.$(`[data-hook="${dataHook}"]`);
+  const getContentType = (type: ContentType) => byDataHook(`content-${type}`);
+
+  return {
+    ...baseUniDriverFactory(base),
+    isContentType: (type: ContentType)=> getContentType(type).exists(),
```
JoMarton: I don't see the usefullness of `isContentType`. It is contained within more specific methods that help test behavior and not implementation (putting a `data-hook` on something)

packages/wix-ui-core/src/components/avatar/avatar.story.js
```diff
@@ -0,0 +1,18 @@
+import {Avatar} from '.';
+import {
+  avatar,
+} from "../../themes/default";
+
+export default {
```
JoMarton: why `.js`?

packages/wix-ui-core/src/components/avatar/avatar.tsx
```diff
@@ -0,0 +1,81 @@
+import * as React from "react";
+import {Omit} from 'type-zoo';
+
+import { BaseProps } from "../../types/BaseProps";
+import style from "./avatar.st.css";
+import {ContentType} from './types';
+
+export interface AvatarProps
+  extends BaseProps {
```
JoMarton: BaseProps extends `InheritedAttributes` which is:
```
interface InheritedAttributes {
    className?: string;
    [props: string]: any;
}
```

This means any prop is OK and no prop type validation at all. Is this fine???

packages/wix-ui-core/src/components/avatar/avatar.tsx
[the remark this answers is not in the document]
JoMarton: @hepiyellow I'm sorry maybe i'm confused but I don't remember agreeing that we remove `tabIndex`. It doesn't have to have focus/focus-visible styles, but if it isn't in tab navigation flow at all there's also no reason to put `aria-label` as well right?

packages/wix-ui-core/src/components/avatar/index.ts
```diff
@@ -0,0 +1 @@
+export { Avatar } from "./avatar";
```
JoMarton: If driver is public we need to export it as well right?

packages/wix-ui-core/src/themes/default/avatar/avatar.st.css
```diff
@@ -0,0 +1,24 @@
...
+:import {
+  -st-from: '../../../components/avatar/avatar.st.css';
+  -st-default: Avatar;
+}
+
+.avatar {
+  -st-extends: Avatar;
+
+  display: flex;
+  justify-content: center;
```
JoMarton: I think basic justification, aligning, and colors should be in core st.css

packages/wix-ui-core/src/components/avatar/avatar.st.css
hepiyellow: no need

packages/wix-ui-core/src/themes/backoffice/avatar/avatar.st.css
hepiyellow: @JoMarton 
We are considering taking the backoffice theme out of `wix-ui-core` and into `wix-style-react`.
For now let's keep it.
Since importing from the index, requires generating the index first, I am not sure how it fits our build process.

packages/wix-ui-core/src/components/avatar/avatar.private.driver.ts
hepiyellow: See public driver after change.

packages/wix-ui-core/src/components/avatar/avatar.tsx
hepiyellow: Thanks for the catch.

hepiyellow: @JoMarton 
FYI: https://github.com/wix/wix-ui/pull/691/files#r230583513

packages/wix-ui-core/src/themes/default/avatar/avatar.st.css
hepiyellow: Talking with @shlomitc 
I got the idea that in the core stylesheet there should be only rules which are NECESSARY for the component to function.

There icon and image are not necessarily centred. (See back-office avatar Placeholder)

Even the text is not necessarily centred.
I can imagine a design where the Avatar is no rounded, and the initials appear in a corner.

@JoMarton 
BTW: IN such case, maybe the consumer wants to render the user's image and add an overlay with his initials. How would we support that? (using `icon`? which is JSXElement slot?)

packages/wix-ui-core/src/components/avatar/index.ts
hepiyellow: Thats a good questions.
@shlomitc 
## Currently consumption methods

### Components JS 

consumed from the `root/comp` (`wix-ui-core/comp`) (using `import-path`)

### Component Stylesheets

Stylable index.st.css (which sits on root folder)

### Drivers

`import from 'wix-ui-core/dist/src/components/comp/comp.driver';`

We are now starting to use UniDriver, but still most of the components have both react and protractor (and might be pupeteer) drivers.

## Suggestion
I am not sure whats the status of tree-shaking, and Stylable tree-shaking.
So if we want to support all consumers (even those who are not using webpack-4 and don't have tree shaking working proparely),
then we should use `import-path` and put also `Comp.driver` and `Comp.protractor.driver` on the root.

JoMarton: We are already auto exporting drivers from `./drivers/vanilla` or `./drivers/protractor`.

packages/wix-ui-core/src/components/avatar/avatar.tsx
```diff
@@ -0,0 +1,82 @@
+import * as React from "react";
+import {Omit} from 'type-zoo';
+
+import { BaseProps } from "../../types/BaseProps";
+import style from "./avatar.st.css";
+import {ContentType} from './types';
+
+export interface AvatarProps extends BaseProps {
```
JoMarton: extending `BaseProps` negates proper type checking

packages/wix-ui-core/src/components/avatar/avatar.story.js
JoMarton: and why here and not in stories directory

shlomitc: both comments are right

packages/wix-ui-core/src/themes/default/avatar/avatar.st.css
shlomitc: I'm ok with having really zero styling in the core and have every theme dealing it on its own

packages/wix-ui-core/src/themes/backoffice/avatar/avatar.st.css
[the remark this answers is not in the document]
mykas: Good point. I'll change that in buttons too :)

packages/wix-ui-core/src/components/avatar/avatar.driver.ts
```diff
@@ -0,0 +1,28 @@
...
+    getTextContent: () => getContentType('text').text(),
+    getIconContent: () => getContentType('icon').$(':first-child'),
```
shlomitc: I don't think we should return an element. Better have common action needed.
in the icon case, it's really redundant. whoever passes the icon knows the component already.
you can do assertions related to the type (isTextDisplayed, isIconDisplayed) etc...

packages/wix-ui-core/src/themes/default/avatar/avatar.st.css
JoMarton: This means the component has NO looks whatsoever (at least nothing resembling avatar) out of the box. 

If that is ok with you then ok with me

packages/wix-ui-core/src/components/avatar/avatar.driver.ts
hepiyellow: @shlomitc 
This is a generic "Render-Slot" pattern we need to decide on.
@argshook @yardnsm @mykas 

- What if, in our case,  `isIconDisplayed() === true`, but the icon is not really rendered. (I don't mean that there's a bug in the driver). Shouldn't the consumer be able to test that?
- What if the component augments the given icon. In that case the consumer should be able to access it and assert on the rendered result.

### Suggested Guide-line
I think we should expose an element getter in the driver in the following cases:
- Pure Render slot (not augmented) 
- Render slot (augmented)
- xxxProps - Like in avatar's case where we have imgProps and we render an `` tag

## Avoid "complex" driver API
With all that said above, I understand the wish for a concise driver API,
but I don't think it should be on the account of consumer testing abilities.

There are several ways to expose everything which is needed, and still have a concise API.

### Nesting Raw driver
An example Avatar driver would be used like this:
```
driver.isContentType('icon')
driver.raw.getContentElement()
```

I used the word "raw" and not "private", since I believe that the consumer should be able to test anything they want. It is not only our privilege.

packages/wix-ui-core/src/components/avatar/avatar.tsx
hepiyellow: fixed

packages/wix-ui-core/src/themes/default/avatar/avatar.st.css
hepiyellow: Well, the "default" theme comes with wix-ui-core, so you can apply it "out-of-the-box".
And the wix-ui-core storybook (if I'm not mistaken) should shows the components with the default theme.
This of course need to be very clear for consumers, through documentation.

packages/wix-ui-core/src/components/avatar/avatar.story.js
hepiyellow: @shlomitc @JoMarton 

@mykas told me that Nadav asked for all component files to be under the component's folder.
And that's what he did with [button-next](https://github.com/wix/wix-ui/tree/master/packages/wix-ui-core/src/components/button-next)

I personally think that as long as we are using Yoshi, which supports only 1 build path,
then the storybook should be a separate package.
So let me know what is the final decision, and we'll update the [contribution guide](https://github.com/wix/wix-ui/blob/master/docs/contribution-guide.md), and fix it if needed.

packages/wix-ui-core/src/themes/backoffice/avatar/avatar.st.css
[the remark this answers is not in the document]
hepiyellow: I'll add it to the contribution guide-lines

packages/wix-ui-core/src/components/avatar/avatar.tsx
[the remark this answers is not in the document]
hepiyellow: @JoMarton 
`aria-label` is not only used on focusable elements.
It can be used as a label for an image, and [most screen readers will support it](https://www.powermapper.com/tests/screen-readers/labelling/img-aria-label/).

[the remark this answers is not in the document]
JoMarton: Yes but screen readers will read it when it is in focus

packages/wix-ui-core/src/components/avatar/avatar.driver.ts
JoMarton: Users can already test anything they want without using the driver. Just like the driver itself, the users can render the component and look for DOM hooks and attributes. Thats why I agree with @shlomitc  that returning an element isn't effective. 
More over, adding / removing attributes or nesting elements can break users tests that rely on getting raw elements.

The PUBLIC driver should assist the user in making assumptions based on the internal component DOM, and not make the user dependent on that DOM implementation.

shlomitc: agree with @JoMarton.
@hepiyellow - please make drivers simple and return primitive values

packages/wix-ui-core/src/components/avatar/avatar.tsx
hepiyellow: @JoMarton 
BTW: what about translating a11y strings?
Like if we want to have :
`` `title = Avatar of ${name}` ``
OR
`` `alt = Profile photo of ${name}` ``

We'd have to translate the "Avatar of" and "Profile photo of" parts.

hepiyellow:
BTW: what about translating a11y strings?
Like if we want to have :
`title = Avatar of ${name}`
OR
`alt = Profile photo of ${name}`

We'd have to translate the "Avatar of" and "Profile photo of" parts.

packages/wix-ui-core/src/components/avatar/avatar.driver.ts
hepiyellow: @JoMarton 
You would have to be more clear on that.

packages/wix-ui-core/src/components/avatar/avatar.driver.ts
hepiyellow: @JoMarton @shlomitc 
Jonathan wrote: 

> Just like the driver itself, the users can render the component and look for DOM hooks and attributes.

Well, that exactly makes the consumer's (users') test dependent on DOM structure. And thus, will break if we change the DOM structure, or we change a data-hook name. And that's what a driver selector (element getter) solves.

Also Jonathan wrote:

> More over, adding / removing attributes or nesting elements can break users tests that rely on getting raw elements.

Well, I don't understand exactly what you meant here. But I remind you, that I suggested to return elements ONLY in the following cases: [Pure Render slot, Render slot, xxxProps ]. In these cases the corresponding DOM elements are the domain of he consumer. (We might modify them, but in a way which is an API , a contract with the consumer, not an implementation detail).

hepiyellow: Well, good morning.
That was my original thought. But then I found myself wanting the Avatar to add it's own data-hook to the icon, and wanting to avoid that I added a wrapper to the icon, and a private getter,...

You are right (-:
Thanks for the concise answer.

Though at least, apart from waisting your time, we have a rule of thumb here to follow,
which is **don't add data-hook to render slots** (of any of the 3 types mentioned above).

P.S: We shell discuss (3) offline

packages/wix-ui-core/src/components/avatar/avatar.tsx
JoMarton: Since we'll fallback to either icon or text I hope we won't get to the point where `alt` is displayed as text.

What I want to avoid is that someone has an image of bruce lee and the alt of the image is "actor Bruce Lee in yellow track suit" when it should be the avatar `name`

JoMarton:
@hepiyellow  regarding `aria` and `title`
For this specific product if you look at the spec we don't put "Avatar of...." but only the `name`.

But the concern is indeed real and need to be discussed as guidelines

PR #843, opened 2018-10-30
[VOD-3967] Add United Video Player

mokhovyk:


packages/wix-ui-core/src/components/Video/Video.spec.tsx
```diff
@@ -1,87 +1,75 @@
...
       it('should set initial value', () => {
-        const driver = createDriver(<Video playing/>);
-        expect(driver.isAutoPlaying()).toBe(true);
+        const driver = createDriver(<Video src={playableVideoUrl}/>);
+        expect(driver.getPlayerDisplayName()).toBe('Playable');
+      });
+
+      it('should set array for src', () => {
+        const driver = createDriver(<Video src={[playableVideoUrl]}/>);
+        expect(driver.getPlayerDisplayName()).toBe('Playable');
```
kupriyanenko: can we check it real `src` that was set?

packages/wix-ui-core/src/components/Video/Video.spec.tsx
```diff
@@ -94,66 +82,95 @@ describe('Video', () => {
...
-        expect(driver.getRootDOMNode().style.height).toBe('100%');
+        const driver = createDriver(
+          <Video
+            config={{
+              playable: {poster: imageUrl, title: 'Awesome'}
+            }}
+            src={playableVideoUrl}
+          />
+        );
+        expect(driver.hasTitle()).toBe(true);
```
kupriyanenko: it's good to check and the content of the title.

packages/wix-ui-core/src/components/Video/Video.tsx
```diff
@@ -1,203 +1,86 @@
...
-};
-
-export class Video extends React.PureComponent<VideoProps, VideoState> {
+export class Video extends React.Component<IVideoProps, IVideoState> {
   static displayName = 'Video';
-
-  containerRef: HTMLDivElement;
-  player: any;
-
+  static getPlayerName = getPlayerName;
```
kupriyanenko: do you plan to use `getPlayerName` outside of this component?

packages/wix-ui-core/src/components/Video/players/playerHOC.tsx
```diff
@@ -0,0 +1,185 @@
...
+      const method = this.isPlayingNow ? 'pause' : 'play';
+      const event = this.isPlayingNow ? EVENTS.PAUSED : EVENTS.PLAYING;
+      const result = this._callPlayer(mapMethodsToPlayer[method]);
+
+      return (result instanceof Promise) ? result : new Promise(resolve => {
+        this.ref.eventEmitter.once(event, () => resolve());
+      });
+    }
+
+    public stop(): Promise<any> {
```
kupriyanenko: Better to use `Promise<void>` in cases like that (we have a couple of them).

packages/wix-ui-core/src/components/Video/players/DailyMotion.tsx
```diff
@@ -0,0 +1,151 @@
...
+
+  componentDidMount() {
+    const {
+      playing, muted, controls, showTitle, playerOptions,
+      onReady, onDuration, onError
+    } = this.props;
+    const src = this.props.src as string;
+    const [, id] = src.match(MATCH_URL);
+
+    getSDK(
```
kupriyanenko: we can think about using `async await` to make code pretty

kupriyanenko: or we can move this code to separate function.

packages/wix-ui-core/src/components/Video/Video.tsx
mokhovyk: yes, as a URL verifier in a settings panel

packages/wix-ui-core/src/components/Video/players/playerHOC.tsx
mokhovyk: done

packages/wix-ui-core/src/components/Video/players/DailyMotion.tsx
mokhovyk: moved resulted function to separate function.

packages/wix-ui-core/src/components/Video/Video.spec.tsx
mokhovyk: done

kupriyanenko:
@shlomitc can you please review (but not merge)

packages/wix-ui-core/src/components/Video/players/Twitch.tsx
```diff
@@ -0,0 +1,171 @@
...
+    this.progressTimeout = window.setTimeout(this.progress, PROGRESS_INTERVAL);
+  }
+
+  stopProgress() {
+    window.clearTimeout(this.progressTimeout);
+  }
+
+  render() {
+    return (
+      <React.Fragment>
```
shlomitc: why is this Fragment needed?. can't you just return the `div`?

packages/wix-ui-core/src/components/Video/players/Vimeo.tsx
```diff
@@ -0,0 +1,156 @@
...
+      this.currentTime = seconds;
+      onProgress(seconds);
+    });
+
+    this.player.on('error', onError);
+  }
+
+  render() {
+    return (
+      <React.Fragment>
```
shlomitc: same comment regarding the fragment

packages/wix-ui-core/src/components/Video/utils/getSDK.ts
```diff
@@ -0,0 +1,70 @@
...
+        }
+        onLoaded(window[name])
+      }
+    } else {
+      onLoaded(window[name]);
+    }
+  })
+};
+
+const requireSDK = (name, url, onLoaded, onError, resolveRequire) => {
```
shlomitc: I'm not sure I get this mechanism. how does it work? can't you use dynamic import for this case?

packages/wix-ui-core/src/components/Video/utils/getSDK.ts
```diff
@@ -0,0 +1,70 @@
...
+  // @ts-ignore
+  window.require([url], sdk => {
+    window[name] = resolveRequire(sdk);
+    onLoaded(window[name])
+  }, err => {
+    onError(err);
+  })
+};
+
+export function getSDK({
```
shlomitc: this should probably be tested as well

packages/wix-ui-core/package.json
```diff
@@ -50,10 +50,12 @@
   "dependencies": {
     "@stylable/dom-test-kit": "^0.1.0",
     "classnames": "^2.2.5",
+    "eventemitter3": "^3.1.0",
     "hoist-non-react-statics": "^3.1.0",
     "jsdom": "^11.12.0",
+    "load-script": "^1.0.0",
```
shlomitc: looks like an old and unmaintained library - are the any better alternatives

packages/wix-ui-core/src/components/Video/Video.tsx
```diff
@@ -1,203 +1,77 @@
...
-    const coverStyles = {
-      backgroundImage: poster ? `url(${poster})` : 'none'
-    };
+    const Player = playerComponents[playerName];
+    const playerProps = {...this.props, ...this.props.config[playerName]};
+    const {id, fillAllSpace, playerRef} = this.props;
     let {width, height} = this.props;
 
-    if (this.props.fillAllSpace) {
+    if (fillAllSpace) {
```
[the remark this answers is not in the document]
mokhovyk: because the fillAllSpace prop redefines width and height properties which can be set only as inline css, so I cannot use stylable state and css for fillAllSpace implementation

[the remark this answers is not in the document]
shlomitc: Why can't it be done?
In this file:
```tsx
{...styles('root', {fillAllSpace: this.props.filAllSpace}, this.props)}
```
And in your `.st.css`
```css
:root {
  -st-states: fillAllSpace;
}

.root:fillAllSpace {
  width: 100%;
  height: 100%
}
```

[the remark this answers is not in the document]
mokhovyk: because width and height are set as an inline style, in this case, !important rule should be used in css. I think current implementation is more clear.

packages/wix-ui-core/src/components/Video/types.ts
```diff
@@ -0,0 +1,212 @@
...
+  onPlay?: Function;
+  onPause?: Function;
+  onEnded?: Function;
+  onFirstPlay?: Function;
+  onFirstEnded?: Function;
+  onError?: Function;
+}
+
+export interface IDailyMotionConfig {
+  playerOptions?: {};
```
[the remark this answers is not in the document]
mokhovyk: the playerOptions are used to set additional player props which are specific to the player. I would not like to typify external properties that we do not control

packages/wix-ui-core/src/components/Video/Video.tsx
```diff
@@ -208,24 +82,11 @@ export class Video extends React.PureComponent<VideoProps, VideoState> {
         style={{width, height}}
         {...styles('root', {}, this.props)}
       >
-        <div 
-          ref={el => this.containerRef = el}
-          style={{width, height}} 
-          className={styles.playerContainer}
+        <Player
+          {...playerProps}
+          key={playerName}
```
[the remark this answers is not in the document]
mokhovyk: isn't needed. removed

packages/wix-ui-core/src/components/Video/Video.tsx
```diff
@@ -1,203 +1,77 @@
 import * as React from 'react';
+import noop = require('lodash/noop');
+import findKey  = require('lodash/findKey');
```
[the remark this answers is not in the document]
mokhovyk: reduced

packages/wix-ui-core/src/components/Video/Video.tsx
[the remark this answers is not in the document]
kupriyanenko: and `width` and `height` it's actually `props` that can be set in px and we need to apply them like inline CSS.

packages/wix-ui-core/src/components/Video/players/Twitch.tsx
mokhovyk: Fragment was removed

packages/wix-ui-core/src/components/Video/players/Vimeo.tsx
mokhovyk: Fragment was removed

packages/wix-ui-core/src/components/Video/players/playerHOC.tsx
```diff
@@ -0,0 +1,183 @@
...
+): React.ComponentType<any> {
+
+  return class extends React.Component<ICommonProps> {
+    static propTypes = Player.propTypes;
+    static displayName = Player.displayName;
+
+    ref: IPlayerRef;
+    isPlayingNow: boolean = false;
+    state: IState = {
+      playerKey: uniqueId('player-'),
```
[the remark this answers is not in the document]
mokhovyk: this key is not an identifier for the player, it is a helper to force-refresh a react child component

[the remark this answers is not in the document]
shlomitc: I know React `key`s are useful for repeating items in a list, not for a single one, so why do you think this is a force-refresh helper?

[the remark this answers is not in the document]
mokhovyk: there is no difference to use for one item or a list. the component is recreated with each change of key.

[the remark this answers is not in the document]
shlomitc: so... what's the point? the key is not going to change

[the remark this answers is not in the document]
mokhovyk: the playerKey may be changed in the _reload method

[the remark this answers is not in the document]
shlomitc: is it tested?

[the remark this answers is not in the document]
mokhovyk: not. I'll cover it

packages/wix-ui-core/package.json
mokhovyk: the alternative is https://github.com/muicss/loadjs. I'll use it instead of "load-script"

packages/wix-ui-core/src/components/Video/Video.tsx
```diff
@@ -1,203 +1,77 @@
...
-        if (typeof method === 'string') {
-          this.player[method](nextProps[propKey]);
-        } else {
-          method(this, this.player, nextProps[propKey]);
-        }
-      }
+  componentWillReceiveProps(nextProps: IVideoProps) {
+    if (this.props.src !== nextProps.src) {
+      this.setState({
+        playerName: getPlayerName(nextProps.src)
```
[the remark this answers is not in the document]
mokhovyk: added test for this case

packages/wix-ui-core/src/components/Video/Video.driver.private.ts
```diff
@@ -1,38 +1,36 @@
...
-    getWidth: () => containerNode.style.width,
-    getHeight: () => containerNode.style.height,
-    getLogoSrc: () => companyLogoNode.getAttribute('src'),
-    isAutoPlaying: () => player.getAutoPlay(),
-    isMuted: () => player.getMute(),
+    getWidth: () => rootDOMNode.style.width,
+    getHeight: () => rootDOMNode.style.height,
     setProp: (prop, value) => wrapper.setProps({[prop]: value}),
+    isMuted: () => player.isMuted(),
+    getPlayerDisplayName: () => player.constructor.displayName,
```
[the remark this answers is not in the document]
mokhovyk: Change logic of identification of a player to container's data attr

packages/wix-ui-core/stories/Video.story.tsx
```diff
@@ -6,20 +6,22 @@ export default {
   component: Video,
   componentPath: '../src/components/Video/Video.tsx',
 
-  componentProps: setState => ({
-    src: 'https://wixmp-01bd43eabd844aac9eab64f5.wixmp.com/videos/output/720p/Highest Peak.mp4',
-    poster: 'https://images-vod.wixmp.com/d0220cbc-4355-4bc0-8ebe-53e9ab8843ba/images/23a7996667c04ebc8bd7e9b6141e30cb~mv2/v1/fill/w_400,h_225,q_85,usm_0.66_1.00_0.01/file.jpg',
+  componentProps: {
```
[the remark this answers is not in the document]
mokhovyk: added exampleProps for src and config for each player

packages/wix-ui-core/src/components/Video/utils/getSDK.ts
mokhovyk: tested

packages/wix-ui-core/src/components/Video/COMPONENT_SPEC.md
```diff
@@ -0,0 +1,86 @@
...
+| width     | number | string                  |              |            | Set the width of the player                                            |
+| height    | number | string                  |              |            | Set the height of the player                                           |
+| fillAllSpace | string                        |              |            | Pass `true` to alow player fill all space of it container              |
+| playing   | boolean                          |              |            | Set to `true` or `false` to pause or play the video                    |
+| muted     | boolean                          |              |            | Mutes the player                                                       |
+| loop      | boolean                          |              |            | Loop video playback                                                    |
+| volume    | number                           |              |            | Start value of volume for audio, `0..100`                              |
```
shlomitc: no default value?

packages/wix-ui-core/src/components/Video/COMPONENT_SPEC.md
```diff
@@ -0,0 +1,86 @@
...
+| controls  | boolean                          | true         |            | Pass false to hide controls                                            |
+| showTitle | boolean                          | true         |            | Pass false to hide title                                               |
+| config    | string                           |              |            | Override options for the various players                               |
```
shlomitc: object?

packages/wix-ui-core/src/components/Video/Video.e2e.ts
```diff
@@ -0,0 +1,95 @@
+import {browser, $} from 'protractor';
+import {createStoryUrl, waitForVisibilityOf} from 'wix-ui-test-utils/protractor';
+import * as autoExampleDriver from 'wix-storybook-utils/AutoExampleDriver';
+import {videoTestkitFactory} from '../../testkit/protractor';
+import {storySettings} from './storySettings';
+
+describe('Video', () => {
```
shlomitc: tests are good but are they needed to be in e2e? I'm just afraid they will become flaky since making actual network requests could fail

packages/wix-ui-core/src/components/Video/Video.story.tsx
```diff
@@ -0,0 +1,64 @@
+import * as React from 'react';
+
+import {Video} from '.';
+import { storySettings } from './storySettings';
+
+export default {
+  category: storySettings.category,
+  storyName: storySettings.storyName,
+  component: Video,
+  componentPath: 'Video.tsx',
```
shlomitc: are you sure this is the right path?

packages/wix-ui-core/src/components/Video/utils/getSDK.ts
shlomitc: so can you please explain how come a browser has `window.require`?

packages/wix-ui-core/src/components/Video/utils/getSDK.spec.ts
```diff
@@ -0,0 +1,103 @@
+import * as eventually from 'wix-eventually';
+
+describe('Video/getSDK', () => {
```
shlomitc: `isRequireAllow` is not tested here

packages/wix-ui-core/src/components/Video/Video.story.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+  componentProps: {
```
shlomitc: `fillAllSpace` doesn't work for the native Playable player, can you please give a look?

packages/wix-ui-core/src/components/Video/COMPONENT_SPEC.md
mokhovyk: fixed and added default value as {}

packages/wix-ui-core/src/components/Video/Video.story.tsx
mokhovyk: sure. double checked

packages/wix-ui-core/src/components/Video/Video.story.tsx
mokhovyk: it doesn't work in the storybook because is required external wrapper with the given size parameters to fill all that space. 
e.g. set inline style  width: 600px; height: 600px; in a debugger for parent of data-hook="storybook-video"

packages/wix-ui-core/src/components/Video/utils/getSDK.spec.ts
mokhovyk: it is a specific case for one player so I don't test it. isRequireAllow helps to solve  https://github.com/vimeo/player.js/issues/79

packages/wix-ui-core/src/components/Video/Video.e2e.ts
mokhovyk: agree, it is duplicate of unit test logic. so I'll remove it. Also it will be e2e tests for this component within integration with santa-editor at https://github.com/wix-private/js-wixcode-integration-headless

packages/wix-ui-core/src/components/Video/COMPONENT_SPEC.md
mokhovyk: added 100 as default

packages/wix-ui-core/src/components/Video/utils/getSDK.ts
mokhovyk: It comes from an external script on a page. window.require is presented if browser has window.define and window.define.amd. this check the same as in the Vimeo Player script. https://github.com/vimeo/player.js/issues/79

packages/wix-ui-core/src/components/Video/utils/getSDK.spec.ts
mokhovyk: anyway I added test for this case

packages/wix-ui-core/src/components/video/utils/getSDK.spec.ts
```diff
@@ -0,0 +1,138 @@
...
+
+// Some tests are skipped because it either passes with jsdom and fails with browser (mocha-runner)
+describe('Video/getSDK', () => {
+  let getSDK;
+  let loadjs;
+
+  describe('success pass', () => {
+    beforeEach(() => {
+      if (typeof jest.mock === 'function') {
+        jest.mock('loadjs', () => jest.fn().mockImplementation((url, settings) => {
```
alisey: @shlomitc @mokhovyk mokhovyk 

`jest.mock` is not supported in the browser environment since it needs to read the module source code from the file system. I would suggest providing `loadjs` to `getSDK()` as an argument, with `loadjs` being the default. Then we could manually create a mock for `loadjs`. Basically the same pattern we have in [address input tests](https://github.com/wix/wix-ui/blob/master/packages/wix-ui-core/src/components/address-input/AddressInput.spec.tsx#L30).

mokhovyk: loadjs is mocked with dependency injection (added a mock function, I think it is more clear than use loadjs as an argument)

mokhovyk: @shlomitc @alisey that was the last question, so could you merge it now?

shlomitc: adding testing code to your production code is not a good idea and it keeps a global state between your tests.
however, it can be refactored later, so let's merge it first

PR #847, opened 2018-10-31
Spec for <SegmentedToggle/> Comp

JoMarton:


hepiyellow:
@JoMarton May I ask where is this component spec coming from? Which UX team designed it? Which company needs it?

JoMarton:
@hepiyellow 
I made this spec a while ago when we were trying to create as many product specs for `ui-core` components. As far as I know, it is part of the backoffice design system called multi-toggle. Since no one picked it up for dev it was in PR a long while and @shlomitc merged it in probably not to be lost.

Regarding building block approach - Product is usually here to describe the "building" and not the "building blocks". If the product is specifically for a building block (item without a functionality on its own) it at least needs to be aware of a few buildings it will be used in so it fits the needs of all users. 
If you see an already existing building block here then, by all means, use it.  

And just as a technical side note - there's a small difference between a "radio" option and a "list" option in terms of native browser behavior (a11y and keyboard) so just make sure it fits

hepiyellow:
@JoMarton 
Yeah, I know. Product drives development.
I just don't want us to forget the part where we pause for a minute before implementing a product spec, and do our own "Product" work, for building blocks.

About the difference between Radio option and List options, can you specify the differences?
- Keyboard navigation seems the same to mee. (Arrows control options, Tab steps out of component)
- a11y - Well the native element might be different, yeah. (`<input type="radio"/>` vs. `<div/>`)

JoMarton:
* in radio arrows select while in list arrows only move focused and space/enter selects
* in radio both up/down and left/right work
* in radio the selection is cyclical and list not


a11y screen readers i'm less able to give specifics so I could be wrong

PR #851, opened 2018-11-01
EnhancedUniDriver - with dataHook queries

hepiyellow:
Enhances UniDriver with dataHook query methods.


mykas:
Looks pretty great. Though I would suggest creating another PR for wix-ui-test-utils library and move the code there. As you see we have failing wix-ui-core because of that.

hepiyellow:
@mykas The EnhancedUniDriver is in wix-ui-test-utils

mykas:
Yes, but you modify and use it at the same time in this PR. If you separate wix-ui-test-utils changes with another PR - you could achieve green builds.

hepiyellow:
@shlomitc If you feel like lior then you can close this.

shlomitc:
Closing for now. let's revive it when more use cases will pop.

PR #852, opened 2018-11-01
Editing <Avatar/> spec

JoMarton:


packages/wix-ui-core/src/components/avatar/README.md
```diff
@@ -11,16 +11,29 @@ Elements are "container" and content, which could be classified to either "text"
...
-| -------- | ------------------- | ------------ | ---------- | --------------------------------------------------------------------- |
-| name     | string              |              | Yes        | The name of the avatar user. Initials will be generated from the name |
-| imgProps | Omit<HTMLImageAttributes, 'alt'> |              |            | the source url to load image from                                     |
-| icon     | JSX Element         |              |            | an SVG icon component                                                 |
-| tabIndex | number              | 0            |            | the tabIndex value to put on the root                                 |
+| name      | type                             | defaultValue | isRequired | description                                                            |
+|:----------|:---------------------------------|:-------------|:-----------|:-----------------------------------------------------------------------|
+| name      | string                           |              |     Yes    | The name of the avatar user. Initials will be generated from the name  |
```
hepiyellow: I don't think it should be required. To be decided....

JoMarton: lets have @shlomitc be the decider

shlomitc: The only scenario I can think about you don't have a name is for an unknown contact.

In that case - a placeholder is displayed, so no initials not name are required. correct me if I'm wrong

hepiyellow: I agree.
Jonathan was concerned what is rendered in case there is no name no text no icon and no image.
In which case it renders just an empty container with no content. and I am ok with that.

The WSR Avatar (which extends CoreAvatar) will pass a placeholder icon in such case.

PR #853, opened 2018-11-01
Stricted stylable import path

mykas:
This PR fixes issue when we can't create multiple file names with the same name. I have only narrow down the folder which we need to expose. Which is `components`. Although we have open question if we should expose theme st.css files too.

Previously we exposed `Focusable.st.css` file, but from `wix-ui-backoffice` I see that these are imported from `dist/src/hocs` anyway so we don't need to target those.

Tested dependencies:
 [x] wix-ui-backoffice
 [x] wix-ui-adi

### DIFF
--- a/packages/wix-ui-core/package.json
+++ b/packages/wix-ui-core/package.json
@@ -24,7 +24,7 @@
     "url": "git+https://github.com/wix/wix-ui.git"
   },
   "scripts": {
-    "build": "yoshi build && npm run generate-stylable-index && npm run transpile-mixins && npm run import-path && build-storybook",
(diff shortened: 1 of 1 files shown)


mykas:
Yes, maybe later. It requires me to create generator for it. As I would like to output the theme files to themes folder with paths to source `../dist/src/themes/**`.  There is something that fails to generate when i use it like this `stc --srcDir="../dist/src/themes/backoffice" --diagnostics --indexFile=./themes/backoffice/index.st.css`

mykas:
I will merge this for now and think about exposing themes later. Its more complex issue.

PR #857, opened 2018-11-01
wix-storybook-utils - Add componentWrapper config

hepiyellow:
@argshook do you agree with this?

A AutoExample Preview config to allow adding backoffice theme class to WSR components.

>Sorry for the formatting changes, my lint rules are probably not aligned.

A render function for the component (in the Preview). Typicaly this function can wrap the component in something usefull like a theme className if needed.
  Signature: ({component}) => JSXElement

  ```js
  export default {
    componentWrapper: ({component})=> <div className="theme">{component}</div>
  }
  ```

### DIFF
--- a/packages/wix-storybook-utils/src/AutoExample/index.test.js
+++ b/packages/wix-storybook-utils/src/AutoExample/index.test.js
@@ -102,8 +102,8 @@ describe('AutoExample', () => {
     it('should render wrapper when given', () => {
       const testkit = new Testkit(AutoExample);
       testkit.when.created({
-        componentWrapper: ({ componentWrapper }) => (
-          <div className="classname">{componentWrapper}</div>
+        componentWrapper: ({ component }) => (
+          <div className="classname">{component}</div>
         ),
       });
       expect(testkit.get.exists('[data-hook*="wrapper"]')).toBeTruthy();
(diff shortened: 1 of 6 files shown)


mykas:
Not sure if this is the way to go.I've talked with Shlomi about showing the story of theme and decided that we should go for examples instead of API in themes. Core component should show autodocs, but not the theme story.

mykas:
Oh wait, this is for wsr?

argshook:
wix-storybook-utils is now yoshi 3, so your lint changes should fail build, please update.

Its not a bad idea as it can be useful not only for themes, but id call it simply `preview`

packages/wix-storybook-utils/src/AutoExample/index.test.js
```diff
@@ -93,7 +94,24 @@ describe('AutoExample', () => {
...
+
+  describe('componentWrapper', () => {
+    it('should render wrapper when given', () => {
+      const testkit = new Testkit(AutoExample);
+      testkit.when.created({
+        componentWrapper: ({ component }) => (
+          <div className="classname">{component}</div>
+        ),
+      });
+      expect(testkit.get.exists('[data-hook*="wrapper"]')).toBeTruthy();
```
hepiyellow: I think you should also assert that the located element is a div with the given className.
Otherwise, the code may simply render you data-hook without the given  wrapper (div+className), and the test would still pass.

hepiyellow:
not needed

PR #861, opened 2018-11-05
Backoffice stylesheet- TextButton

mykas:
This pr adds weight stylesheet and renames `secondary` class to `onHover` as it is more descriptive. There is no `secondary` text button.


milkyfruit:
@mykas What is shown in storybook looks well.
I couldn't test variation `small` with `prefix` or `suffix` icon.

mykas:
Thanks!

PR #863, opened 2018-11-06
DOCS: Add Styling best-practices

hepiyellow:


docs/styling-best-practices.md
```diff
@@ -0,0 +1,33 @@
...
+} 
+.colorRed  {
+  color: red;
+} 
+```
+The css priority logic would take the default color rule (blue) over the red.
+
+In other cases, rule ordering would decide on the priority.
+
+This makes it hard to add feaults, or refactor the stylesheet.
```
JoMarton: "feaults" typo?

hepiyellow: why of course. you don't know what "feaults" are?
I'll tell you.
Its....
It is very much like "defaults".

docs/styling-best-practices.md
```diff
@@ -0,0 +1,65 @@
+# Styling Best-Practices
+
+## Avoid lifting classes
+Some people would write a component's stylsheet like this:
+```css
+.button {
```
alisey: Is `button` a root of the component or does it `-st-extend` something?

Do you want to add this section based on something you saw in a PR? A component CSS should not be written like this, and I think we have a more fundamental problem here than wrong specificity.

docs/styling-best-practices.md
```diff
@@ -0,0 +1,65 @@
...
+  background-color: grey;
+  color: white;
+}
+
+.button .prefix {
```
alisey: Where does `.prefix` class come from? Using descendant selector may accidentally affect descendant nodes that don't belong to the component.

hepiyellow:
@shlomitc I will meet with Alexey and discuss this next week

PR #870, opened 2018-11-07
`<Avatar/>` - add Backoffice theme

hepiyellow:
## Known Issues
- `font-weight` - should it be `normal` or `500`


hepiyellow:
Replaced by #887

PR #877, opened 2018-11-08
Feat/auto testkit

jakutis:


argshook:
PR is ready to merge, but as it has impact on many projects, i'll merge it on Monday to avoid any weekend fires

PR #886, opened 2018-11-11
Contribution guide - meta files

JoMarton:


docs/contribution-guide.md
```diff
@@ -185,55 +184,63 @@ By providing a single method which causes a chain of actions, rather than separa
...
-
-
-```ts
+// you can even override the component's state to simulate interactions (like focus for example)
+metaData.addSim('checked and focused', {
+    props: {
+      name: 'myCheckbox',
+      checked: true
+    },
+    state: {
```
shlomitc: is this a css only override?

JoMarton: nope. can override any state (for instance `open: true` for dropdown). Checkbox is very simple so I don't think it will have a state that's not related directly to style. 

You think it needs another example (like for dropdown)?

shlomitc: If it’s not only css, meaning it’s just another prop. So why do you need this section apart from the props?

JoMarton: Maybe the example I brought is bad. Consider Slider - https://github.com/wix/wix-ui/blob/master/packages/wix-ui-core/src/components/Slider/Slider.tsx#L77

Could be you want to present in showcase and take a snapshot of the slider while in dragging state. This is the way to do it

shlomitc: this is better. should probably explain this actually changes the internal react state of the component

JoMarton: cool. will do

JoMarton: @shlomitc added a sentence, is that enough?

shlomitc: Not sure it's clear enough...
1. Should explain the this will change the **internal** react component state
2. The example is talking about focus, which you'd expect to be native css. I prefer something that explains something about functionality. for example, a dropdown in an open state

JoMarton: Ok added the dropdown example

PR #894, opened 2018-11-13
`<Avatar/>` - add image fallback

hepiyellow:
This PR Includes:
- image fallback
- Change style API: 
  - add `::content` instead (text / icon / image)
  - add `contentType` state


packages/wix-ui-core/src/components/avatar/storySettings.ts
```diff
@@ -0,0 +1,7 @@
+import { StorySettings, Category } from '../../../stories/utils';
```
shlomitc: Just from curiosity, wouldn't it be better to have a single storySettings file for all the stories instead of one per story?

packages/wix-ui-core/stories/utils.ts
```diff
@@ -0,0 +1,13 @@
+export enum Category {
+  COMPNENTS = 'Components'
+}
+
+/**
+ * Hold story settings for AutoStory.
+ */
+export interface StorySettings {
```
shlomitc: if we go with one file, there is no need to split the definitions from it

packages/wix-ui-core/src/components/avatar/storySettings.ts
hepiyellow: It might.
As long as this setting s file include only simple objects and strings. it is possible.
IN some cases storySettings may contain also data-hooks of examples, and even more data describing the examples.
I added to my suggested infra backlog.

PR #901, opened 2018-11-17
`<ButtonNext/>` - Theme - change API to be prop object

hepiyellow:
instead of :
```js
<ButtonNext className={button('primary', 'secondary')}/>
```

We'll have:

```js
<ButtonNext className={button({skin: 'primary', secondary: true})}/>
```

I did it only got ButtonNext.
If it look good, we can do it for all the rest.

### DIFF
--- a/packages/wix-ui-core/src/themes/backoffice/index.ts
+++ b/packages/wix-ui-core/src/themes/backoffice/index.ts
@@ -1,7 +1,6 @@
 import classNames from 'classnames';
 
 import Avatar from './avatar/avatar.st.css';
-import ButtonNext from './button/button.st.css';
 import IconButton from './icon-button/icon-button.st.css';
 import TextButton from './text-button/text-button.st.css';
 import CloseButton from './close-button/close-button.st.css';
@@ -14,8 +13,9 @@ const getClassNames = (values, stylesheet, rootcls) => {
 
 export const avatar = (...values) =>
   getClassNames(values, Avatar, 'avatar');
-export const buttonNext = (...values) =>
-  getClassNames(values, ButtonNext, 'button');
+
+export {button as Button} from './button/button-styles';
+
 export const iconButton = (...values) =>
   getClassNames(values, IconButton, 'iconButton');
 export const textButton = (...values) =>
--- a/packages/wix-ui-core/src/themes/backoffice/index.ts
+++ b/packages/wix-ui-core/src/themes/backoffice/index.ts
@@ -14,7 +14,7 @@ const getClassNames = (values, stylesheet, rootcls) => {
 export const avatar = (...values) =>
   getClassNames(values, Avatar, 'avatar');
 
-export {button as Button} from './button/button-styles';
+export {button} from './button/button-styles';
 
 export const iconButton = (...values) =>
   getClassNames(values, IconButton, 'iconButton');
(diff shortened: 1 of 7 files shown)


packages/wix-ui-core/src/themes/backoffice/index.ts
```diff
@@ -1,21 +1,22 @@
...
 
 import Avatar from './avatar/avatar.st.css';
-import ButtonNext from './button/button.st.css';
 import IconButton from './icon-button/icon-button.st.css';
 import TextButton from './text-button/text-button.st.css';
 import CloseButton from './close-button/close-button.st.css';
 import BackofficeTheme from './theme.st.css';
 
+/* New generation prop-like API */
+export {button, ButtonStyleProps, ButtonSize, ButtonSkin} from './button/button-classes';
```
mykas: finally, we can have types! Looks pretty great.

mykas: why do we need to export these static variables?

packages/wix-ui-core/stories/backoffice/button-next/showcase/button/button-affixes.tsx
```diff
@@ -5,19 +5,19 @@ import CodeShowcase from 'wix-storybook-utils/CodeShowcase';
...
 
 const example = `import * as React from "react";
-import { ButtonNext, buttonNext } from "wix-ui-core/button-next";
+import { ButtonNext, Button } from "wix-ui-core/button-next";
 import Add from "wix-ui-icons-common/Add";
 
 export default () => (
   <React.Fragment>
-    <ButtonNext className={buttonNext()} prefixIcon={<Add />}>
+    <ButtonNext className={Button()} prefixIcon={<Add />}>
```
mykas: I think we have forgot about the idea Shahar wanted us to do. Basically to apply default theme when scope is applied. So maybe here there shouldn't be any className.

packages/wix-ui-core/stories/backoffice/button-next/showcase/button/button-primary.tsx
```diff
@@ -13,12 +13,12 @@ import { buttonNext } from "wix-ui-core/themes/backoffice";
 
 export default () => (
   <React.Fragment>
-    <ButtonNext className={buttonNext()}>default</ButtonNext>
-    <ButtonNext className={buttonNext("destructive")}>destructive</ButtonNext>
-    <ButtonNext className={buttonNext("premium")}>premium</ButtonNext>
-    <ButtonNext className={buttonNext("dark")}>dark</ButtonNext>
-    <ButtonNext className={buttonNext("light")}>light</ButtonNext>
-    <ButtonNext className={buttonNext("transparent")}>transparent</ButtonNext>
+    <ButtonNext className={button()}>Standard (default)</ButtonNext>
```
mykas: Here too I think it should be without defining anything.

packages/wix-ui-core/src/themes/backoffice/index.ts
shlomitc: maybe just `export * from './button/button-classes'`?

packages/wix-ui-core/src/themes/backoffice/button/button-classes.spec.ts
```diff
@@ -0,0 +1,46 @@
...
+  function parseClasses(classes: string) {
+    return classes.split(' ').map(c => c.split('--')[1]);
+  } 
+
+  function testProp<T extends string>(propName: string, values: T[], defaultValue?: T) {
+    values.forEach((value)=>{
+      const classes = button({[propName]: value});
+      if (value === defaultValue) {
+        //expect classes NOT to contain any value
+        const pClasses = parseClasses(classes);
```
shlomitc: maybe we should use here the `@stylable/dom-test-kit` instead of parsing manually

Issue #762, opened 2018-10-02
wix-ui-core: `<Popover/>` - componentDidUpdate does not update portal node


hepiyellow:
I think the title is quite descriptive.

I think we need to call `applyStylesToPortaledNode` in `componentDidUpdate`

### Example
See FloatingHelper's story (`wix-ui-backoffice`), with the new `appearance` property.
https://wix-wix-ui-backoffice-pr-218.surge.sh/?selectedKind=Components&selectedStory=FloatingHelper&full=0&addons=0&stories=1&panelRight=0

When switching to `light` appearance, the `data-appearance` attribute on the portal node is still `dark`.

yardnsm:
Resolved in https://github.com/wix/wix-ui/pull/902

PR #902, opened 2018-11-18
<Popover/> - update styles and re-apply them in componentDidUpdate

yardnsm:
Currently, the `<Popover/>` styles were calculated only once (in `initAppendToNode`). This would cause a problem when the styles changed, as it simply wouldn't update.

Relevant issues:
  - https://github.com/wix/wix-ui/issues/762
  - https://github.com/wix-private/wsr-issues/issues/323

This PR resolve both of them. The styles (`stylesObj`) will be re-calculated in the `componentDidUpdate` lifecycle method. In addition, there were multiple calls to `this.applyStylesToPortaledNode`. These calls were removed and instead a single call is made in `componentDidUpdate`.

This PR was tested with `wix-ui-backoffice`'s `<FloatingHelper/>` component (https://github.com/wix-private/wsr-issues/issues/323).

### DIFF
--- a/packages/wix-ui-core/src/components/Popover/Popover.tsx
+++ b/packages/wix-ui-core/src/components/Popover/Popover.tsx
@@ -236,10 +236,6 @@ export class Popover extends React.Component<PopoverProps, PopoverState> {
         height:0
       });
       this.appendToNode.appendChild(this.portalNode);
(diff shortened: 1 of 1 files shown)


hepiyellow:
@yardnsm 
I don't see any new tests.

PR #910, opened 2018-11-20
Backoffice - textbutton, iconbutton, closebutton - remove backoffice scope

mykas:
This PR removes backoffice scope for textbutton, iconbutton and closebutton stylesheets. The reason for this is because these components are variation components of a core Button and it will not have any default values when applied with Backoffice scope wrapper. Additional classes must be passed instead.

### DIFF
--- a/packages/wix-ui-core/src/themes/backoffice/text-button/text-button.st.css
+++ b/packages/wix-ui-core/src/themes/backoffice/text-button/text-button.st.css
@@ -1,9 +1,4 @@
 :import {
-  -st-from: "../theme.st.css";
-  -st-default: BackofficeTheme;
-}
-
-:import {
   -st-from: '../../../components/button-next/button-next.st.css';
   -st-default: Button;
   -st-named: prefix, suffix, content;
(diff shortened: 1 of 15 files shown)


packages/wix-ui-core/src/themes/backoffice/text-button/text-button.st.css
```diff
@@ -287,81 +282,81 @@ BackofficeTheme .textButton.always.light:disabled {
...
-BackofficeTheme .textButton.always.dark:active:not(:disabled) {
+.textButton.always.dark:active:not(:disabled) {
   background-color: transparent;
   border-color: transparent;
   color:value(D10);
   text-decoration: underline;
 }
 
-BackofficeTheme .textButton.always.dark:disabled {
+.textButton.always.dark:disabled {
```
hepiyellow: @mykas don't you want to prefix variation classes with the semantic meaning of them?
`always` -> `underlineAlways`
`small` -> `sizeSmall`
etc...

PR #916, opened 2018-11-20
`<Avatar/> - Add nameToInitials improvements, title, aria-labal

hepiyellow:
- improve nameToInitials logic
- Add title & aria-label props

## name To initials
backoffice design changed so that they want to limit to 2 letters.
I added that as a prop `initialsLimit` that can be set to either 2 or 3.
Since I still think that 3 letter initials could be useful for other design systems,
and we can save them the bother of writing and testing  their own `nameToInitials` function.

### DIFF
--- a/packages/wix-ui-core/src/components/avatar/avatar.driver.ts
+++ b/packages/wix-ui-core/src/components/avatar/avatar.driver.ts
@@ -10,7 +10,12 @@ export interface AvatarDriver extends BaseUniDriver{
   getContentType: () => Promise<ContentType>;
   /** Get the text content (generated initials) */
   getTextContent: ()=> Promise<string>;
+  /** Wether the image wass loaded */
   isImageLoaded: () => Promise<boolean>;
+  /** Get the native HTML title attribute */
+  getTitle: () =>  Promise<string>;
+  /** Get the native HTML aria-label attribute */
+  getAriaLabel: () =>  Promise<string>;
 }
 
 export const avatarDriverFactory = (base: UniDriver): AvatarDriver => {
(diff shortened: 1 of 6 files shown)


packages/wix-ui-core/src/components/avatar/avatar.spec.tsx
```diff
@@ -1,9 +1,10 @@
 import * as React from 'react';
 import {StylableDOMUtil} from '@stylable/dom-test-kit';
 import * as eventually from 'wix-eventually';
-import { reactUniDriver, UniDriver } from 'unidriver';
+import { reactUniDriver } from 'unidriver';
```
mykas: Why do we use this here? don't we have dom-container for this?

packages/wix-ui-core/src/components/avatar/avatar.driver.ts
```diff
@@ -20,6 +25,8 @@ export const avatarDriverFactory = (base: UniDriver): AvatarDriver => {
     ...baseUniDriverFactory(base),
     getContentType,
     getTextContent: ()=> base.$('[data-hook="text-container"]').text(),
-    isImageLoaded: async () => ((await base.attr('data-img-loaded')) === 'true')
+    isImageLoaded: async () => ((await base.attr('data-img-loaded')) === 'true'),
+    getTitle: ()=> base.attr('title'),
```
JoMarton: I don't think the PUBLIC driver should support `getTitle` or `getAriaLabel`.

hepiyellow: why not. Consumer use translations for the title aria-label and they want to verify that the translation is correct.

packages/wix-ui-core/src/components/avatar/avatar.spec.tsx
hepiyellow: `dom-test-container` has a method `renderUniDriver` which both renders and creates the driver.
The avatar public driver does not expose a UniDriver for the root element (as most of us think it shouldn't).
So If I need a UniDriver for the root element I need to create it myself.

PR #918, opened 2018-11-21
 kebab case for folder names

RashelleKosib:
a change in the format of folder names naming. will cause breaking changes in liabilities. **MAJOR**


RashelleKosib:
@shlomitc  pls review :) tnx

PR #921, opened 2018-11-21
remove costly lodash operators

noakorkus:


packages/wix-ui-core/src/components/Dropdown/Dropdown.tsx
```diff
@@ -76,11 +74,20 @@ export class DropdownComponent extends React.PureComponent<DropdownProps & Injec
...
 
   componentWillReceiveProps(props: DropdownProps) {
-    if (!isEqual(props.initialSelectedIds, this.props.initialSelectedIds)) {
+    if (!this.areSelectedIdsEqual(props.initialSelectedIds, this.props.initialSelectedIds)) {
       this.initializeSelectedOptions(props);
     }
   }
 
+  areSelectedIdsEqual = (selectedIds1, selectedIds2) => {
+    if ((selectedIds1 === undefined && selectedIds2 === undefined) || (selectedIds1 === null && selectedIds2 === null)) {
```
shlomitc: maybe just do `if(!selectedIds1 && !selectedIds2) return true`

packages/wix-ui-core/src/components/Input/Input.tsx
```diff
@@ -55,6 +53,8 @@ export class Input extends React.Component<InputProps, InputState> {
       style: inlineStyle
     } = this.props;
 
+    let {error: errorProp, style: styleProp, ...allOtherProps}: any = this.props;
```
shlomitc: `const` instead of `let`.
and `any` should probably be replaced with the right type

packages/wix-ui-core/src/components/TimePicker/TimePicker.tsx
```diff
@@ -380,14 +379,12 @@ export class TimePicker extends React.PureComponent<TimePickerProps, TimePickerS
       ...rest
     } = this.props;
 
+    let {onChange: restOnChange, step: restStep, value: restValue, ...allOther}: any = rest;
```
shlomitc: same comment as above

packages/wix-ui-core/src/components/AddressInput/AddressInput.tsx
```diff
@@ -206,7 +203,7 @@ export class AddressInput extends React.PureComponent<AddressInputProps, Address
         const {lang, filterTypes} = this.props;
         const results = await this.client.autocomplete(this._getKey(), lang, createAutocompleteRequest(input, this.props));
         const filteredResults = filterAddressesByType(results, filterTypes);
-        const options = map(filteredResults, this._createOptionFromAddress);
+        const options = (filteredResults || []).map(this._createOptionFromAddress);
```
shlomitc: move the `|| []` to the `filterAddressByType` result

packages/wix-ui-core/src/utils/intersection.tsx
```diff
@@ -0,0 +1,5 @@
+export function intersection(array1, array2) {
+  const set1 = new Set(Array.isArray(array1) ? array1 : []);
```
shlomitc: @giladsegal - is the `Set` function transpiled by `typescript` to ES5 or a polyfill is needed?

PR #923, opened 2018-11-22
Rashellek/refactor directories to kebab case

RashelleKosib:
changed all component directories names from pascal-case to kebab-case, in favour of windows case sensitivity issues.


RashelleKosib:
> could this be done gradually and not all at once? 261 files changed in one PR, it's scary

@argshook  I understand, but- at the same time - at least 80% of the files are triggered as changes because their directory changed, so there is no code to check.

argshook:
> I understand, but- at the same time - at least 80% of the files are triggered as changes because their directory changed, so there is no code to check.

elaborate please. are you not able to take one component and convert it to kebab case? why does it have to happen for all code base at once?

doing such a change throughout whole library is dangerous. it's very possible there are other projects that rely on file path, maybe they import some stylable file or some driver or whatever.

is it urgent? if yes, why?

JoMarton:
To begin with I don't think this PR is ready yet, The entry point files have changed so we need to deal with that first.

> elaborate please. are you not able to take one component and convert it to kebab case? why does it have to happen for all code base at once?

> doing such a change throughout whole library is dangerous. it's very possible there are other projects that rely on file path, maybe they import some stylable file or some driver or whatever.


Of course, it can happen at different PRs, but raises the question: how many files changed are ok? also, I assume you would want to wait between merges to make sure nothing breaks - this makes the whole process so much longer.

We took A LOT of time beforehand, PRing `ui-core` liabilities redirecting requires from `ui-core/dist/src/...` to a common entry point. So we currently shouldn't have known liabilities relying on the file path. That being said, if somebody is relying on internal file path it will break and fixing is easy.


> is it urgent? if yes, why?

Not urgent but (when PR is ready) should be done, cause why not?

argshook:
@JoMarton 

> Of course, it can happen at different PRs, but raises the question: how many files changed are ok?

one component per PR

> this makes the whole process so much longer.

so what? are we suddenly in a rush for kebab case?

> Not urgent but (when PR is ready) should be done, cause why not?

the `cause why not` is what concerns me. Sure we can do it but not as mega bomb PR all at once. You are the one who introduce requirements, we are the one who maintain and answer questions. I would prefer to answer questions about one component at a time and not whole lib about changes that i may have missed, because there were so many.

JoMarton:
@argshook @shlomitc 
It was decided to change the directory and filenames to kebab-case. No rush for kebab but currently we have the two patterns in the library and we shouldn't have people making new components copying the old pattern. 

A library should have an entry point for each type of "utility" (component, drivers, testKits) so we mask the internal structure and we can change it as we like (as long as the entry point is named the same and still correct)
@RashelleKosib has done previous work both on `ui-core` and liabilities to formalize missing entry points.
This makes such PR low risk. If A liability does break because it reaches into `dist` then it will be fixed.

Now logistically - We don't have merge permissions to `ui-core`. That makes every merge take an additional +1 day at least from experience. Separating this PR to ~30 PRs is just something we won't do.

Let us know how to continue

argshook:
@shlomitc may the decision be yours, i'm not gonna fight

RashelleKosib:
@shlomitc I also think bumping the version make sense. @JoMarton are there any consequences in ui-core by bumping the version you can think of?

JoMarton:
@shlomitc 

I'm ok with major version, but if so - let's discuss maybe we can change / rethink additional stuff we've been working around

shlomitc:
@JoMarton @RashelleKosib I see that the output of import-path is forcing **PascalCase**, but didn’t we want to make it kebab-case for importing as well?

I understand more work will be done later, but I think subject this should be resolved in this PR, before we merge.

PR #924, opened 2018-11-25
<Popover/> - migrate to react-popper 1.3.0

yardnsm:
This PR migrates `react-popper` to its latest version `1.3.0`, as it is a "stable" version and the `0.x` versions are not maintained anymore.

Verified on `wix-ui-backoffice`; tooltip looks and functions as expected.

Note that this PR also updates `@types/react` to use `^16.4.2` instead of `~16.4.2`. This change can cause some TypeScript type errors when.


packages/wix-ui-core/src/components/Popover/Popover.tsx
```diff
@@ -136,33 +140,43 @@ export class Popover extends React.Component<PopoverProps, PopoverState> {
...
-        }
+        {({ ref, style: popperStyles, placement: popperPlacement, arrowProps, scheduleUpdate }) => {
+          this.scheduleUpdate = scheduleUpdate;
+
+          return (
+            <div
+              ref={ref}
+              data-hook="popover-content"
+              style={popperStyles}
+              data-placement={popperPlacement || placement}
```
hepiyellow: This is for styling only, so it could/should be changed into a Stylable state.
(although it is a breaking change, If some other component which uses Popover has css rules with selectors using the `data-placement` attribute)

packages/wix-ui-core/src/components/Popover/Popover.tsx
```diff
@@ -256,34 +285,46 @@ export class Popover extends React.Component<PopoverProps, PopoverState> {
       // Apply the styles to the portal
       this.applyStylesToPortaledNode();
     }
+
+    // Update popper's position
+    if (this.scheduleUpdate) {
+      this.scheduleUpdate();
```
yardnsm: I've added a test case for this, but it was removed in https://github.com/wix/wix-ui/pull/924/commits/29adecdc2e8bdd780c220c49882928cc2fde040f as I had troubles mocking `popper.js`. I would appreciate if someone would suggest an alternative. @hepiyellow @shlomitc @argshook

packages/wix-ui-core/src/utils/index.ts
```diff
@@ -18,8 +18,11 @@ export const buildChildrenObject = <T>(children: React.ReactNode, childrenObject
 
 export interface ElementProps { children: any; }
 export const createComponentThatRendersItsChildren = (displayName: string) => {
-  const Element: React.SFC<ElementProps> = ({children}) => children;
+  const Element: React.SFC<ElementProps> = ({children}) => 
+    typeof children === 'string' ? React.createElement('div', {}, children) : children;
```
hepiyellow: Why do this?
You said something about an array.

yardnsm: Maybe `children` can be wrapped in an `Array`, so for examples:

```suggestion
    typeof children === 'string' ? [children] : children;
```

packages/wix-ui-core/src/components/Popover/Popover.tsx
hepiyellow: Please try to assert on something which is a DOM side-effect of calling popper `scheduleUpdate`.
That way the test is independent of the specific usage of the popper library.

yardnsm: If we'd assert something in the DOM as a result of `scheduleUpdate`, it'll be the opposite of "independent of the specific usage of the popper library". IMHO testing that `scheduleUpdate` was called regardless of its implementation is enough.

packages/wix-ui-core/src/components/Popover/Popover.tsx
yardnsm: I'll add a comment about this, I think we could live with it for now (we also don't want to break anyone).

packages/wix-ui-core/src/utils/index.ts
yardnsm: Didn't work :/

PR #929, opened 2018-11-28
Backoffice - theming - change scoping to new stylable feature @st-scope.

mykas:
It get rids of warnings on build.

### DIFF
--- a/packages/wix-ui-core/src/themes/backoffice/button/button.st.css
+++ b/packages/wix-ui-core/src/themes/backoffice/button/button.st.css
@@ -26,7 +26,7 @@
 }
 
 
-@st-scope {
+@st-scope BackofficeTheme {
 
   Button {
     box-sizing: border-box;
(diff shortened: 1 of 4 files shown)


mykas:
@hepiyellow can you check from your side that everything works on avatar? we don't have screenshot testing so it is a bit hard to check everything.

packages/wix-ui-core/src/themes/backoffice/button/button.st.css
```diff
@@ -25,465 +25,466 @@
...
-}
-
-BackofficeTheme .light.secondary:hover:not(:disabled) {
-  background-color: value(B50);
-  color: value(B10);
-  border-color: value(B50);
-}
+@st-scope BackofficeTheme {
+
+  Button {
```
hepiyellow: I think you still need to define a `.button` class,
Since if you add `Button` rules, 
then TextButton and others must override them.

mykas:
https://github.com/wix/wix-ui/pull/930

PR #997, opened 2018-12-03
Image component product spec

JoMarton:


packages/wix-ui-core/src/components/image/image.md
```diff
@@ -0,0 +1,193 @@
...
+
+## Description
+
+The **Image** component represents an image on the DOM.
+
+## API
+
+**Props**
+
+**Image** accepts all native `<img />` attributes, with several additional features listed below.
```
JoMarton: This approach of accepting all native attributes has both benefits and problems to it. 
We need to consider cases such as user provides `width` and `height` props to the component (since they are supported https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img) and how does it play along with stylable...

packages/wix-ui-core/src/components/image/image.md
```diff
@@ -0,0 +1,193 @@
...
+
+
+| name        | type       | default | required | description       |
+| ----------- | ---------- | ------- | -------- | ----------------- |
+| src | string | - | ✖ |  Contains the path to the image you want to embed. |
```
JoMarton: If we include all native attributes, `src` is included in them along with `srcset`

packages/wix-ui-core/src/components/image/image.md
```diff
@@ -0,0 +1,193 @@
...
+| alt | string | - | ✖ | Contains a textual description of the image. |
```
JoMarton: same as for `src`

packages/wix-ui-core/src/components/image/image.md
```diff
@@ -0,0 +1,193 @@
...
+| title | string | - | ✖ | Specifies extra information about an element, shown as a tooltip text. |
```
JoMarton: Do you mean the native `title` attribute to put on an `` ?

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img
This has a warning about using `title` in such way, and also warns about using it in combination with `alt`.

What is the exact use case for this prop? Is it to show a tooltip?

packages/wix-ui-core/src/components/image/image.md
```diff
@@ -0,0 +1,193 @@
...
+| onLoad | (event: ImageEvent) => void; | noop | ✖ | An event handler triggered by the state's status. |
+| onError | (event: ImageEvent) => void; | noop | ✖ | An event handler setting an Error state. |
+| resizeMode | 'fill' \| 'cover' \| 'contain' | 'fill' | ✖ | Defines how the Image responds to the height and width of its content box. |
+| defaultImage | string | - | ✖ | URL to load when src is not provided. |
+| errorImage | string | - | ✖ | URL to load if src (or defaultImage) loading result in an error. |
+| loader | 'icon' \| 'dots' \| 'spinner' | - | ✖ | Element being displayed during the loading time. |
```
JoMarton: I don't understand the type here. can you elaborate on what happens when the user chooses each?

packages/wix-ui-core/src/components/image/image.md
```diff
@@ -0,0 +1,193 @@
...
+
+
+
+### React Code Example
+TODO: change the example by for the new components (new stylable version)
```
JoMarton: Please fix the example. The format has to change. It can be simplified so its easier to understand

packages/wix-ui-core/src/components/image/image.md
```diff
@@ -0,0 +1,193 @@
...
+            this.setState({resizeMode: e.target.value as ImageDemoState['resizeMode']});
+        }
+}
+
+```
+*Example of rendering the Image component, passing it specific props*
+
+## Style 
+
+The style states are derived from ImageStatus which can be one of the following options: { Loaded, Loading, Error }.
```
JoMarton: I'm missing the `Selectors (pseudo-elements)`

packages/wix-ui-core/src/components/image/image.md
JoMarton: Also no other mention of it in the rest of the doc

packages/wix-ui-core/src/components/image/image.md
```diff
@@ -0,0 +1,193 @@
...
+.myImage:loaded { border: 3px solid gold }
+
+.myImage:loading { border: 3px solid grey }
+
+.myImage:error { border: 3px solid red }
+
+```
+
+## Accessibility
+Our components is accessible according to [WAI-ARIA 1.0](https://www.w3.org/TR/wai-aria/) standards.
```
JoMarton: No need to keep this generic text here

packages/wix-ui-core/src/components/image/image.md
```diff
@@ -0,0 +1,193 @@
...
+#### Resize Modes
+
+`<Image>` allows specifying a `resizeMode` prop with these possible values: `fill`, `contain`, and `cover`. The behavior of `resizeMode` is the same as that of the CSS [`object-fit`](https://developer.mozilla.org/en-US/docs/Web/CSS/object-fit) prop, except that `resizeMode` supports older browsers that don't support `object-fit`.
+
+> Note: Just as in `object-fit`, `resizeMode` only affects images with both width and height axes provided (e.g. `width: 200px; height: 200px;`).
+
+
+#### Avoiding user-agent "broken image" outline
+
+When an `<img />` has no `src` prop, or it fails to load the specified image, some web browsers show a "broken image" placeholder or an outline around the element. These typically do not conform to the page design causing the page to appear broken.
```
JoMarton: note the existence of `srcset` attribute

packages/wix-ui-core/src/components/image/image.md
```diff
@@ -0,0 +1,193 @@
...
+
+The `<Image />` component allows supplying `defaultImage` to replace `src` if missing. If the source fails loading `errorImage` will be displayed. If `errorImage` is not supplied the `<Image/>` component will render an empty pixel.
+
+```
+src -> defaultImage -> errorImage -> one empty pixel
```
JoMarton: This is misleading. `src` failure doesn't result in `defaultImage` loading.

packages/wix-ui-core/src/components/image/image.md
```diff
@@ -0,0 +1,193 @@
...
+
+✔ - source provided and is loaded successfully
+
+✖ - source provided but loading results in error
+
+(empty) - source was not provided
+
+## Public Test Driver
+
+Alongside the component we usually provide our users with a test driver to simplify testing their app's behavior to interaction with the component. See [ImageDriver](./image-driver.ts).
```
JoMarton: No need to keep this text here

packages/wix-ui-core/src/components/image/image.md
```diff
@@ -0,0 +1,193 @@
...
+
+### Test Driver API
+> Note: ImageDriver API is an API for all intents and purposes, meaning changing this API may result in breaking changes for the component users. To avoid this scenario exclude unneeded methods from the public driver (helper methods for the component's tests can still be implemented in a "private" driver that extends the public one)
```
JoMarton: No need to keep this text here

packages/wix-ui-core/src/components/image/image.md
```diff
@@ -0,0 +1,193 @@
...
+
+| name   | type                    | description |
+|:-------|:------------------------|:------------|
+| nativeElement | () => HTMLImageElement | return the native image element. |
```
JoMarton: We don't return native elements from the driver. This causes tests to be dependant on the DOM structure or attributes

packages/wix-ui-core/src/components/image/image.md
```diff
@@ -0,0 +1,193 @@
...
+Alongside the component we usually provide our users with a test driver to simplify testing their app's behavior to interaction with the component. See [ImageDriver](./image-driver.ts).
+
+### Test Driver API
+> Note: ImageDriver API is an API for all intents and purposes, meaning changing this API may result in breaking changes for the component users. To avoid this scenario exclude unneeded methods from the public driver (helper methods for the component's tests can still be implemented in a "private" driver that extends the public one)
+
+| name   | type                    | description |
+|:-------|:------------------------|:------------|
+| nativeElement | () => HTMLImageElement | return the native image element. |  
+| source | () => string \| null | return the value of 'src' attribute of the element. | 
+| hasStylableState | (stateName: string): boolean | return `true` when component has stylable state. |
```
JoMarton: This is too open-ended. I would much prefer something like 
```
getLoadStatus: () => 'loading' | 'loaded' | 'error'
```

packages/wix-ui-core/src/components/image/image.md
RashelleKosib: good point

packages/wix-ui-core/src/components/image/image.md
RashelleKosib: i am removing it from the props table, but i am going to write usage recommendations and talk about the title attribute as well.

packages/wix-ui-core/src/components/image/image.md
RashelleKosib: loader is removed for the ui-core image component.

packages/wix-ui-core/src/components/image/image.md
RashelleKosib: you are talking about an explanation and an example? im adding a reference to what it is , please let me know if it's what you mean

packages/wix-ui-core/src/components/image/image.md
RashelleKosib: I have a general confusion and doubt about how product wise - I can write here about drivers (?)

packages/wix-ui-core/src/components/image/image.md
```diff
@@ -0,0 +1,221 @@
...
+
+
+#### Edge case handling
+
+Source loading and resulting states breakdown table:
+
+| src | defaultImage | errorImage | source to be displayed | component status |
+| --- | ------------ | ---------- | ---------------------- | ---------------- |
+| ✔ | |  | src | Loaded |
+|  |✔ |  | defaultImage | Loaded |
```
hepiyellow: If `defaultImage` is displayed only when `src` is not provided, then why not just make `src` required?

hepiyellow:
## Placeholder Component (Loader)

I see there is a `loading` state for styling.

But what if we want to show a `<Loader/>` while the image is loading?
(Or any other component, some text, etc...)

A `placeholder` prop could be of use here.

> BTW: This is what Avatar does.

packages/wix-ui-core/src/components/image/image.md
```diff
@@ -0,0 +1,221 @@
...
+| name        | type       | default | required | description       |
+| ----------- | ---------- | ------- | -------- | ----------------- |
+| onLoad | (event: ImageEvent) => void;| noop | ✖ | An event handler triggered by the state's status. |
+| onError | (event: ImageEvent) => void; | noop | ✖ | An event handler setting an Error state. |
+| resizeMode | 'fill' \| 'cover' \| 'contain' | 'fill' | ✖ | Defines how the Image responds to the height and width of its content box. |
+| defaultImage | string | ✖  | ✖ | URL to load when src is not provided. |
+| errorImage | string | ✖  | ✖ | URL to load if src (or defaultImage) loading result in an error. |
+| crop | Crop | ✖  | ✖ | Crops an image from the (x,y) pixel coordinates to be of size (width X height) pixels, scaled down times (scaleFactor).  |
+| filter | Filter | ✖  | ✖ | Used to transform the image using different graphical techniques.  |
+
```
hepiyellow: Missing `src` prop. Or need to state that all HTMLImageAttributes are valid.

packages/wix-ui-core/src/components/image/image.md
```diff
@@ -0,0 +1,221 @@
...
+**Image** accepts all native `<img />` attributes (which you can read about [here](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img)), with several additional features listed below.
+
+
+| name        | type       | default | required | description       |
+| ----------- | ---------- | ------- | -------- | ----------------- |
+| onLoad | (event: ImageEvent) => void;| noop | ✖ | An event handler triggered by the state's status. |
+| onError | (event: ImageEvent) => void; | noop | ✖ | An event handler setting an Error state. |
+| resizeMode | 'fill' \| 'cover' \| 'contain' | 'fill' | ✖ | Defines how the Image responds to the height and width of its content box. |
+| defaultImage | string | ✖  | ✖ | URL to load when src is not provided. |
```
hepiyellow: I think we should replace `defaultImage` with a `placeholder` prop, that:
- Can get any `node` (like `<Loader/>` )
- Will be displayed while the image is loading.

packages/wix-ui-core/src/components/image/image.md
RashelleKosib: it is stated above ..
**Image** accepts all native `` attributes (which you can read about [here](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img)), with several additional features listed below.

packages/wix-ui-core/src/components/image/image.md
RashelleKosib: Im removing defaultImage, tnx!

packages/wix-ui-core/src/components/image/image.md
RashelleKosib: I'm changing the type to a node one, but I'll think of a better name than a placeholder :)

packages/wix-ui-core/src/components/image/image.md
```diff
@@ -0,0 +1,189 @@
...
+
+| src | errorImage | source to be displayed | component status |
+| --- | ---------- | ---------------------- | ---------------- |
+| ✔  |  | src | Loaded |
+|   |  | one empty pixel | Loaded |
+|   | ✔ | one empty pixel | Loaded |
+| ✔  |  | src | Loaded |
+| ✖  | ✔ | errorImage | Error |
+|   | ✔ | errorImage | Error |
+| ✖  | ✔ | errorImage | Error |
```
JoMarton: This one is the same as two lines up

packages/wix-ui-core/src/components/image/image.md
```diff
@@ -0,0 +1,189 @@
...
+Source loading and resulting states breakdown table:
+
+| src | errorImage | source to be displayed | component status |
+| --- | ---------- | ---------------------- | ---------------- |
+| ✔  |  | src | Loaded |
+|   |  | one empty pixel | Loaded |
+|   | ✔ | one empty pixel | Loaded |
+| ✔  |  | src | Loaded |
+| ✖  | ✔ | errorImage | Error |
+|   | ✔ | errorImage | Error |
```
JoMarton: This one contradicts 3 rows up

packages/wix-ui-core/src/components/image/image.md
```diff
@@ -0,0 +1,189 @@
...
+#### Edge case handling
+
+Source loading and resulting states breakdown table:
+
+| src | errorImage | source to be displayed | component status |
+| --- | ---------- | ---------------------- | ---------------- |
+| ✔  |  | src | Loaded |
+|   |  | one empty pixel | Loaded |
+|   | ✔ | one empty pixel | Loaded |
+| ✔  |  | src | Loaded |
```
JoMarton: same as 3 rows up

packages/wix-ui-core/src/components/image/image.md
```diff
@@ -0,0 +1,189 @@
...
+| ✖  | ✔ | errorImage | Error |
+|   | ✔ | errorImage | Error |
+| ✖  | ✔ | errorImage | Error |
+| ✖  |  | one empty pixel | Error |
+|   |  | one empty pixel | Error |
```
JoMarton: contradicts the second line

packages/wix-ui-core/src/components/image/image.md
```diff
@@ -0,0 +1,189 @@
...
+**Image** accepts all native `<img />` attributes (which you can read about [here](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img)), with several additional features listed below.
+
+
+| name        | type       | default | required | description       |
+| ----------- | ---------- | ------- | -------- | ----------------- |
+| onLoad | (event: ImageEvent) => void;| noop | ✖ | An event handler triggered by the state's status. |
+| onError | (event: ImageEvent) => void; | noop | ✖ | An event handler setting an Error state. |
+| resizeMode | 'fill' \| 'cover' \| 'contain' | 'fill' | ✖ | Defines how the Image responds to the height and width of its content box. |
+| errorImage | string | ✖  | ✖ | URL to load if src loading result in an error. |
+| placeholder | ReactNode | ✖  | ✖ | Placeholder to display while image is being loaded. |
```
JoMarton: Maybe add to the code example?

packages/wix-ui-core/src/components/image/image.md
```diff
@@ -0,0 +1,189 @@
...
+    -st-states: loaded, loading, error;
+    display: inline-block;
+}
+```
+
+#### States 
+
+| state | description|
+| ----- | -----------|
+| :loading | Styles the Image when image is loading
```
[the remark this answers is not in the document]
RashelleKosib: Sounds great

PR #1007, opened 2018-12-04
GoogleMapsClient - Don't fail test when getting OVER_QUERY_LIMIT error

hepiyellow:
Well, it's either we eliminate querying 3rd party services in our automated tests,
or we need to ignore the OVER_QUERY_LIMIT error.

That is...  assuming that in production we have an real google account with very high limit ?
 
Because now the full error description we are getting from Google is:
> You have exceeded your daily request quota for this API. If you did not set a custom daily request quota, verify your project has an active billing account: http://g.co/dev/maps-no-account  For more information on usage limits and the Google Maps JavaScript API services please see: https://developers.google.com/maps/documentation/javascript/usage

Which means we have no account.
Is that the same in production?


hepiyellow:
@giladartzi suggests that we skip these tests, or use an api mock.

I found this one (which is more popular than the others I've found):
https://www.npmjs.com/package/jest-googleapi-mock

@shlomitc What do you think?

shlomitc:
Closing until becomes relevant again

PR #1008, opened 2018-12-04
<LiveCodeExample/> - allow supplying non-inline code

yardnsm:
This PR adds the `inlineCode` prop to the `<LiveCodeExample/>` to allow supplying non-inline code. Check out the example in the sample storybook and [`react-live` docs](https://github.com/FormidableLabs/react-live#liveprovider-).


packages/wix-storybook-utils/src/LiveCodeExample/index.js
```diff
@@ -19,13 +19,15 @@ export default class LiveCodeExample extends Component {
     scope: PropTypes.object,
     compact: PropTypes.bool,
     previewRow: PropTypes.bool,
-    previewProps: PropTypes.object
+    previewProps: PropTypes.object,
+    inlineCode: PropTypes.bool
```
argshook: maybe let's call it `manualRender` or something? `inlineCode` to me doesn't tell anything. `noInline` in react-live is also not really descriptive imo

yardnsm: Here are some options I just thought about:

- `manualRender`
- `evalImmediately`
- `mountImmediately`
- `mount`

wdyt?

argshook: i don't like immediately, cause i don't know how to spell it :)

maybe `autoRender` which is true by default? if false, prop name kinda implies you need to render by hand then

PR #1010, opened 2018-12-04
AddressInput - Adding the horizontalPadding prop

giladartzi:


shlomitc:
so - the decision is to have an inline `style` object prop that will be applied to the root. If needs to customize internal parts, have another style prop that is dedicated to the internal element (the dropdown option for example)

giladartzi:
@shlomitc Done

PR #1012, opened 2018-12-05
Multiple code examples fix

egidijusj:


packages/wix-storybook-utils/src/AutoTestkit/examples/puppeteer.ts
```diff
@@ -0,0 +1,21 @@
+export default ({
+  capitalizedTestkitFactoryName,
+  componentLC,
+  pathToTestkit,
+  testkitFactoryName,
+}) => {
+  const puppeteerTestkitFactoryName = `puppeteer${capitalizedTestkitFactoryName}`;
+  return `
+import puppeteer from 'puppeteer';
```
argshook: you can find puppeteer example here https://github.com/wix/wix-style-react/blob/master/src/Input/README.TESTKIT.md#L114-L131

egidijusj: @argshook puppeteer example updated. Keep in mind that now it is not consistent with `enzyme` example:
- generic variable names vs verbose ones
- mocha API vs Jest API
- show-casing actual method `exists()` (which is prevalent in most drivers)  vs imaginary one `getText()`

argshook: > generic variable names vs verbose ones
can you show example of what you mean?

> mocha API vs Jest API

i think we should opt to jest API but even then i wouldn't say it matters much, cause we want to
outline how to get testkit working, not how to assert things

> show-casing actual method exists() (which is prevalent in most drivers) vs imaginary one getText()

there are some white ravens, of course, but generally we can assume that all drivers have `exists`
method, cause they should be extending `baseDriver` implementation which has `exists`

PR #1019, opened 2018-12-10
<Popover/> - added `onClickOutside` prop

yardnsm:
This PR adds an `onClickOutside` prop to the `<Popover/>` component. It is needed for https://github.com/wix/wix-style-react/pull/2509.


packages/wix-ui-core/stories/Popover.story.tsx
```diff
@@ -65,6 +65,7 @@ export default {
 
   exampleProps: {
     children,
+    onClickOutside: () => 'onClickOutside called!',
```
hepiyellow: This doesn't "glow" in story when clicking outside.
Maybe it should be in `componentProps`?

packages/wix-ui-core/src/components/popover/Popover.tsx
```diff
@@ -43,6 +44,8 @@ export interface PopoverProps {
   shown: boolean;
   /** onClick on the component */
   onClick?: React.MouseEventHandler<HTMLDivElement>;
+  /** Provides callback to invoke when outside of popover is clicked */
```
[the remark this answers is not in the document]
yardnsm: When attached to body - yes. This is intended, so you can have non-interactive popovers (see the last examples [here](https://wix-wix-style-react-pr-2509.surge.sh/?selectedKind=Components&selectedStory=Popover&full=0&addons=0&stories=1&panelRight=0), take a look at the mouse events example and not in the click events since the latter depends on this PR :D).

packages/wix-ui-core/stories/Popover.story.tsx
yardnsm: It should be in `exampleProps`: https://github.com/wix/wix-ui/blob/master/packages/wix-storybook-utils/docs/usage.md#using-functions

Maybe a bug? @argshook

hepiyellow:
The gpu thing looks good, just missing a test

PR #1021, opened 2018-12-11
Add playback-bi-module to Playable Video player

mokhovyk:


shlomitc:
BI events are business logic, meaning they should not be part of a reusable component in an open-source project.

it should sit inside your component wrapper repo, AKA wix-ui-santa

PR #1030, opened 2018-12-13
<ButtonNext/> - feature - render button as any other passed html tag

mykas:
Ready.

### DIFF
--- a/packages/wix-ui-core/src/components/button-next/button-next.spec.tsx
+++ b/packages/wix-ui-core/src/components/button-next/button-next.spec.tsx
@@ -1,4 +1,5 @@
 import * as React from 'react';
+import * as eventually from 'wix-eventually';
 import { ReactDOMTestContainer } from '../../../test/dom-test-container';
 import { ButtonNext } from './';
 import { buttonNextPrivateDriverFactory } from './button-next.driver.private';
@@ -7,6 +8,8 @@ describe('ButtonNext', () => {
   const createDriver = new ReactDOMTestContainer()
     .unmountAfterEachTest()
     .createUniRenderer(buttonNextPrivateDriverFactory);
+
+  const renderedCmponent = new ReactDOMTestContainer().unmountAfterEachTest();
 
   describe(`'onClick' prop`, () => {
     it('should be called on click', async () => {
@@ -46,6 +49,34 @@ describe('ButtonNext', () => {
       const driver = createDriver(<ButtonNext prefixIcon={prefix} />);
       expect(await driver.prefixExists()).toBeTruthy();
       expect(await driver.suffixExists()).toBeFalsy();
+    });
+  });
+
+  describe(`'as' prop`, () => {
+    const Test = props => <span {...props} />;
+    it('should render by default as html button', async () => {
+      renderedCmponent.renderSync(<ButtonNext />);
+
+      await eventually(() => {
+        const htmlTag = renderedCmponent.componentNode.tagName;
+        expect(htmlTag).toBe('BUTTON');
+      });
+    });
+    it('should render custom html tag', async () => {
+      renderedCmponent.renderSync(<ButtonNext as="a" />);
+
+      await eventually(() => {
+        const htmlTag = renderedCmponent.componentNode.tagName;
+        expect(htmlTag).toBe('A');
+      });
+    });
+    it('should render custom react component', async () => {
+      renderedCmponent.renderSync(<ButtonNext as={Test} />);
+
+      await eventually(() => {
+        const htmlTag = renderedCmponent.componentNode.tagName;
+        expect(htmlTag).toBe('SPAN');
+      });
     });
   });
 });
--- a/packages/wix-ui-core/src/components/button-next/button-next.spec.tsx
+++ b/packages/wix-ui-core/src/components/button-next/button-next.spec.tsx
@@ -9,7 +9,7 @@ describe('ButtonNext', () => {
     .unmountAfterEachTest()
     .createUniRenderer(buttonNextPrivateDriverFactory);
 
-  const renderedCmponent = new ReactDOMTestContainer().unmountAfterEachTest();
+  const testContainer = new ReactDOMTestContainer().unmountAfterEachTest();
 
   describe(`'onClick' prop`, () => {
     it('should be called on click', async () => {
@@ -55,27 +55,77 @@ describe('ButtonNext', () => {
   describe(`'as' prop`, () => {
     const Test = props => <span {...props} />;
     it('should render by default as html button', async () => {
-      renderedCmponent.renderSync(<ButtonNext />);
+      testContainer.renderSync(<ButtonNext />);
 
       await eventually(() => {
-        const htmlTag = renderedCmponent.componentNode.tagName;
+        const htmlTag = testContainer.componentNode.tagName;
         expect(htmlTag).toBe('BUTTON');
       });
     });
     it('should render custom html tag', async () => {
-      renderedCmponent.renderSync(<ButtonNext as="a" />);
+      testContainer.renderSync(<ButtonNext as="a" />);
 
       await eventually(() => {
-        const htmlTag = renderedCmponent.componentNode.tagName;
+        const htmlTag = testContainer.componentNode.tagName;
         expect(htmlTag).toBe('A');
       });
     });
     it('should render custom react component', async () => {
-      renderedCmponent.renderSync(<ButtonNext as={Test} />);
+      testContainer.renderSync(<ButtonNext as={Test} />);
 
       await eventually(() => {
-        const htmlTag = renderedCmponent.componentNode.tagName;
+        const htmlTag = testContainer.componentNode.tagName;
         expect(htmlTag).toBe('SPAN');
+      });
+    });
+  });
+
+  describe(`'disabled' prop`, () => {
+    it('when given should render component with tabIndex -1', async () => {
+      testContainer.renderSync(<ButtonNext disabled />);
+
+      await eventually(() => {
+        const htmlTag = testContainer.componentNode.getAttribute('tabindex');
+        expect(htmlTag).toBe('-1');
+      });
+    });
+
+    it('when given should render aria-disabled as true', async () => {
+      testContainer.renderSync(<ButtonNext disabled />);
+
+      await eventually(() => {
+        const htmlTag = testContainer.componentNode.getAttribute(
+          'aria-disabled'
+        );
+        expect(htmlTag).toBe('true');
+      });
+    });
+  });
+
+  describe(`'type' prop`, () => {
+    it('should render value `button` when props `as` and `type` are undefined', async () => {
+      testContainer.renderSync(<ButtonNext />);
+
+      await eventually(() => {
+        const htmlTag = testContainer.componentNode.getAttribute('type');
+        expect(htmlTag).toBe('button');
+      });
+    });
+    it('should render value `submit` when prop `type="submit"` and `as` is undefined', async () => {
+      testContainer.renderSync(<ButtonNext type="submit" />);
+
+      await eventually(() => {
+        const htmlTag = testContainer.componentNode.getAttribute('type');
+        expect(htmlTag).toBe('submit');
+      });
+    });
+
+    it('should render prop `type` undefined if as prop is `a`', async () => {
+      testContainer.renderSync(<ButtonNext as="a" />);
+
+      await eventually(() => {
+        const htmlTag = testContainer.componentNode.getAttribute('type');
+        expect(!!htmlTag).toBe(false);
       });
     });
   });
--- a/packages/wix-ui-core/src/components/button-next/button-next.st.css
+++ b/packages/wix-ui-core/src/components/button-next/button-next.st.css
@@ -4,6 +4,7 @@
 }
 
 .root {
+  -st-states: disabled;
   -st-extends: Focusable;
   display: inline-flex;
   align-items: center;
--- a/packages/wix-ui-core/src/components/button-next/button-next.tsx
+++ b/packages/wix-ui-core/src/components/button-next/button-next.tsx
@@ -2,16 +2,27 @@ import * as React from 'react';
 import { withFocusable } from '../../hocs/Focusable/FocusableHOC';
 import style from './button-next.st.css';
 
-export interface ButtonProps extends React.ButtonHTMLAttributes<any> {
+export interface BaseButtonProps {
+  as?: any;
   /** accepts prefix icon */
   prefixIcon?: React.ReactElement<any>;
   /** accepts suffix icon  */
   suffixIcon?: React.ReactElement<any>;
   /** callback need to be applied for onFocus event */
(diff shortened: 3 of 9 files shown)


packages/wix-ui-core/src/components/button-next/button-next.tsx
```diff
@@ -25,30 +36,48 @@ const _addAffix = (Affix, classname) =>
...
+      >
+        {_addAffix(prefixIcon, 'prefix')}
+        <span className={style.content}>{children}</span>
+        {_addAffix(suffixIcon, 'suffix')}
+      </Component>
+    );
+  };
+
+  // Handle rendering anchor instead of button
+  const anchorRestProps = rest as AnchorButtonProps;
```
shlomitc: Why do we need a specific handling to `a`?
Isn't it just as any other element and the rest props can be `any`?
I also not sure I like that `button` defines also the anchor types

mykas: Well, we can say that oficially with types we support only button and anchor. Which means typescript would show autocomplete for anchor and button attributes. Any other custom component will not have this typechecking.

mykas: But we can also leave this to react to check invalid attributes on native html tag

mykas: Ok, so another issue is that wix-ui-core does not have proptypes defined anymore. Which means that my mentioned `feature` to support types for anchor (runntime and build time) would not work in wix-style-react anyway.

packages/wix-ui-core/src/components/button-next/button-next.st.css
```diff
@@ -4,6 +4,7 @@
 }
 
 .root {
+  -st-states: disabled;
```
hepiyellow: Add type boolean

packages/wix-ui-core/src/components/button-next/button-next.spec.tsx
```diff
@@ -8,6 +9,8 @@ describe('ButtonNext', () => {
     .unmountAfterEachTest()
     .createUniRenderer(buttonNextPrivateDriverFactory);
 
+  const renderedCmponent = new ReactDOMTestContainer().unmountAfterEachTest();
```
hepiyellow: TestContainer is a more suitable name than renderedComponent

packages/wix-ui-core/src/components/button-next/button-next.tsx
```diff
@@ -25,30 +28,33 @@ const _addAffix = (Affix, classname) =>
...
     focusableOnBlur,
+    disabled,
     ...rest
   } = props;
+  const Component = as;
   return (
-    <button
+    <Component
       {...rest}
       onFocus={focusableOnFocus}
```
hepiyellow: The provided component may not accept `onFocus` and `onBlur` props (if its a custom react component). 
But the only thing we can do us document it. 

BTW: the `rest` could include an `onFocus`, so we should chain event handlers

packages/wix-ui-core/src/components/button-next/button-next.tsx
```diff
@@ -25,30 +28,33 @@ const _addAffix = (Affix, classname) =>
...
       onBlur={focusableOnBlur}
-      onClick={onClick}
-      {...style('root', {}, props)}
+      type={as === 'button' ? 'button' : rest.type}
```
hepiyellow: Why limit `<button/>` to be of `button` type? If consumer wants a submit type?

The only thing we need to do here is NOT to pass `type` if the `as` results in an element that doesn't accept `type` attribute (not a `button`).

NOTE: See note above, we don't always know what the root DOM element tagName will be. (only after `componentDidMount`)

packages/wix-ui-core/src/components/button-next/button-next.tsx
```diff
@@ -25,30 +28,33 @@ const _addAffix = (Affix, classname) =>
...
 const ButtonNextComponent: React.SFC<ButtonProps> = props => {
   const {
+    as,
     suffixIcon,
     prefixIcon,
     children,
-    onClick,
     focusableOnFocus,
     focusableOnBlur,
+    disabled,
```
hepiyellow: - The `disabled` props should also be spread onto toe root element (e.g. `<button disabled/>`) if it can accept it (e.g. `<a/>` does not accept `disabled` attribute).
- If the root DOM element can NOT accept `disabled`, the we should add `aria-disabled=true`.

NOTE: If the `as` component is a custom react component, then we don't know what will be the resulting root DOM element. We can do this after component has mounted (with `ReactDOM.findDOMNode(this)` )

packages/wix-ui-core/src/components/button-next/button-next.tsx
```diff
@@ -25,30 +28,33 @@ const _addAffix = (Affix, classname) =>
 
 const ButtonNextComponent: React.SFC<ButtonProps> = props => {
   const {
+    as,
```
hepiyellow: I prefer the prop name `component` rather than `as`. (Like material-ui uses)
But do as you wish.

packages/wix-ui-core/src/components/button-next/button-next.tsx
```diff
@@ -2,15 +2,18 @@ import * as React from 'react';
 import { withFocusable } from '../../hocs/Focusable/FocusableHOC';
 import style from './button-next.st.css';
 
-export interface ButtonProps extends React.ButtonHTMLAttributes<any> {
+export interface ButtonProps
+  extends React.ButtonHTMLAttributes<HTMLButtonElement> {
```
hepiyellow: How do you solve the typescript issue ? We need to all AnchorHTMLAttributes. 
I think shlomi’s idea to have a ‘comp’ and ‘compProps’ props would solve that.

packages/wix-ui-core/src/components/button-next/button-next.tsx
mykas: Thanks, thats one of the issues I will solve with disabled. Right now I have only made styles work across other html tags by using stylable states

packages/wix-ui-core/src/components/button-next/button-next.tsx
mykas: Prop as is more widely used in augmented pattern (yes, there is even a pattern name for this). When I participated in one of our lithuanian fed guild weeks and talked about this idea - some of them recognized this pattern when I said 'as prop'.

packages/wix-ui-core/src/components/button-next/button-next.tsx
mykas: well it is true that we need to document this part so that user would pass ...rest in his custom component.

packages/wix-ui-core/src/components/button-next/button-next.tsx
mykas: Resolved this to by default to types for native button attributes, but if user will render as something else - it will fallback to `any`

shlomitc:
this issue can be resolved once merged
https://github.com/wix/wix-style-react/issues/1636

packages/wix-ui-core/src/components/button-next/button-next.tsx
```diff
@@ -2,21 +2,28 @@ import * as React from 'react';
 import { withFocusable } from '../../hocs/Focusable/FocusableHOC';
 import style from './button-next.st.css';
 
-export interface ButtonProps extends React.ButtonHTMLAttributes<any> {
+export interface ButtonProps {
+  /** an element type to render as (string or function).  */
```
hepiyellow: `string, function (SFC), or class constructor`

packages/wix-ui-core/src/components/button-next/button-next.tsx
```diff
@@ -25,30 +32,36 @@ const _addAffix = (Affix, classname) =>
...
   return (
-    <button
+    <Component
       {...rest}
       onFocus={focusableOnFocus}
       onBlur={focusableOnBlur}
-      onClick={onClick}
-      {...style('root', {}, props)}
+      disabled={disabled}
+      type={as === 'button' ? restProps.type || 'button' : undefined}
```
hepiyellow: ### `as === 'button`
The `as` may be an SFC (or class constructor) of a Custom component that renders a `<button/>`.
In that case the `type` value will be `undefined`, which is not what the consumer wants.

packages/wix-ui-core/src/components/button-next/button-next.tsx
```diff
@@ -2,21 +2,28 @@ import * as React from 'react';
 import { withFocusable } from '../../hocs/Focusable/FocusableHOC';
 import style from './button-next.st.css';
 
-export interface ButtonProps extends React.ButtonHTMLAttributes<any> {
+export interface ButtonProps {
+  /** an element type to render as (string or function).  */
+  as?: any;
```
hepiyellow: It would be nice to have a precise type here:
I think : `as?: string | React.ComponentType<any>`

packages/wix-ui-core/src/components/button-next/button-next.tsx
```diff
@@ -25,30 +32,36 @@ const _addAffix = (Affix, classname) =>
...
     suffixIcon,
     prefixIcon,
     children,
-    onClick,
     focusableOnFocus,
     focusableOnBlur,
+    disabled,
     ...rest
   } = props;
+  const Component = as;
```
hepiyellow: I think you can do instead of this line:
```js
const {
  as: Component
} = props
```

packages/wix-ui-core/src/components/button-next/button-next.tsx
```diff
@@ -25,30 +32,36 @@ const _addAffix = (Affix, classname) =>
...
+  const restProps = as === 'button'? rest as NativeButtonProps: rest as any;
   return (
-    <button
+    <Component
       {...rest}
       onFocus={focusableOnFocus}
       onBlur={focusableOnBlur}
-      onClick={onClick}
-      {...style('root', {}, props)}
+      disabled={disabled}
```
hepiyellow: ### `as='a'` & `disabled=true`

In this case, the`disabled` is not a valid attribute of an `<a/>`.
We will get a React warning for that (i think). So in this case:
- [ ] We need to NOT pass it at all to the root element.
- [ ] We still want to allow passing `disabled` prop, to make a "LinkButton" look disabled. (Stylable state)
- [ ] Also if disabled, we need to NOT pass the `href` prop to the `<a/>` so the link would be "in-active".

packages/wix-ui-core/src/components/button-next/button-next.tsx
```diff
@@ -2,21 +2,28 @@ import * as React from 'react';
...
 }
 
+export type NativeButtonProps = React.ButtonHTMLAttributes<HTMLButtonElement> & ButtonProps;
+
+
 const _addAffix = (Affix, classname) =>
   Affix &&
   React.cloneElement(Affix, {
-    className: style[classname]
+    className: style[classname],
```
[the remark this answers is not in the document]
mykas: I think in this case we control it and ignore user defined classnames. Many cases I see when user add their own classnames to icons and it looks bad in production.

packages/wix-ui-core/src/components/button-next/button-next.tsx
```diff
@@ -25,30 +32,36 @@ const _addAffix = (Affix, classname) =>
 
 const ButtonNextComponent: React.SFC<ButtonProps> = props => {
   const {
+    as,
     suffixIcon,
```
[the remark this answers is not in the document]
mykas: Nothing `icon`-ish? What do you mean?

packages/wix-ui-core/src/components/button-next/button-next.tsx
```diff
@@ -2,21 +2,28 @@ import * as React from 'react';
...
+export interface ButtonProps {
+  /** an element type to render as (string or function).  */
+  as?: any;
   /** accepts prefix icon */
   prefixIcon?: React.ReactElement<any>;
   /** accepts suffix icon  */
   suffixIcon?: React.ReactElement<any>;
   /** callback need to be applied for onFocus event */
-  focusableOnFocus?: React.FocusEventHandler<HTMLButtonElement>,
+  focusableOnFocus?: React.FocusEventHandler<HTMLButtonElement>;
```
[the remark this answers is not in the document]
mykas: Do you mean that I should just extend interface from FocusableHOC?

packages/wix-ui-core/src/components/button-next/button-next.tsx
```diff
@@ -25,30 +32,36 @@ const _addAffix = (Affix, classname) =>
...
     prefixIcon,
     children,
-    onClick,
     focusableOnFocus,
     focusableOnBlur,
+    disabled,
     ...rest
   } = props;
+  const Component = as;
+  const restProps = as === 'button'? rest as NativeButtonProps: rest as any;
```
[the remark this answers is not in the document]
mykas: This helps me to deconstruct `rest` in any way I like. For example `href`. If I would just leave NativeButtonProps - it will complain that the component is `button` based.

packages/wix-ui-core/src/components/button-next/button-next.tsx
mykas: Doesn't work. Complains that react custom element is not the right type.

packages/wix-ui-core/src/components/button-next/button-next.tsx
mykas: good trick

packages/wix-ui-core/src/components/button-next/button-next.tsx
```diff
@@ -2,21 +2,28 @@ import * as React from 'react';
...
-  focusableOnFocus?: React.FocusEventHandler<HTMLButtonElement>,
+  focusableOnFocus?: React.FocusEventHandler<HTMLButtonElement>;
   /** callback need to be applied for onBlur event */
-  focusableOnBlur?: React.FocusEventHandler<HTMLButtonElement>,
+  focusableOnBlur?: React.FocusEventHandler<HTMLButtonElement>;
+  /** apply disabled styles */
+  disabled?: boolean;
 }
 
+export type NativeButtonProps = React.ButtonHTMLAttributes<HTMLButtonElement> & ButtonProps;
```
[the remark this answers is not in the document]
mykas: Yes, that could be done in this way too.

packages/wix-ui-core/src/components/button-next/button-next.tsx
mykas: Regarding the last point I think that its gonna be messy to control this in source code. Maybe it is actually easier just to make this `disabled` state available for users instead?

Look at this 
```js
const anchorDisabled = (isAnchor && !restProps.href) || disabled ? disabled : undefined;
```

packages/wix-ui-core/src/components/button-next/button-next.tsx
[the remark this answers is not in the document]
mykas: Or maybe this could break their designed icon with classnames...

PR #1044, opened 2018-12-26
<Popover/> - add support for `flip` and update `moveBy` properly

yardnsm:
This PR adds a `flip` prop in order to control Popper.JS's `flip` behaviour (more info can be found [here](https://popper.js.org/popper-documentation.html#modifiers..flip)).

In addition, the `moveBy` behaviour has been changed so when its enabled, the `flip` behaviour will be disabled in order to support passing negative values to the Content element will overlap the Target element.

Also, it fixes the misleading usage at per https://github.com/wix/wix-style-react/issues/2636.

----

**Update:** a new `fixed` prop was added. When set, it'll disable the `preventOverflow` modifier so the popover will stay at the same position at all times. Check the examples for more info.

### DIFF
--- a/packages/wix-ui-core/src/components/popover/Popover.tsx
+++ b/packages/wix-ui-core/src/components/popover/Popover.tsx
@@ -59,6 +59,11 @@ export interface PopoverProps {
    * when it starts to overlap the target element (`<Popover.Element/>`).
    */
   flip?: boolean;
+  /**
+   * Whether to enable the fixed behaviour. This behaviour used to keep the `<Popover/>` at it's
+   * original placement even when it's being positioned outside the boundary.
+   */
+  fixed?: boolean;
   /** Moves popover relative to the parent */
   moveBy?: { x: number, y: number };
   /** Hide Delay */
@@ -127,7 +132,7 @@ const calculateOffset = ({moveBy, placement}): string => {
   return `${moveBy ? moveBy.x : 0}px, ${moveBy ? moveBy.y : 0}px`;
 };
 
-const createModifiers = ({moveBy, appendTo, shouldAnimate, flip, placement}) => {
+const createModifiers = ({moveBy, appendTo, shouldAnimate, flip, fixed, placement}) => {
   const modifiers: PopperJS.Modifiers = {
     offset: {
       offset: calculateOffset({moveBy, placement}),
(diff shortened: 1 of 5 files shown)


packages/wix-ui-core/src/components/popover/Popover.tsx
```diff
@@ -54,6 +54,16 @@ export interface PopoverProps {
   onKeyDown?: React.KeyboardEventHandler<HTMLDivElement>;
   /** Show show arrow from the content */
   showArrow?: boolean;
+  /**
+   * Whether to enable the flip behaviour. This behaviour used to flip the `<Popover/>`'s placement
```
hepiyellow:
```suggestion
   * Whether to enable the flip behaviour. This behaviour is used to flip the `<Popover/>`'s placement
```

packages/wix-ui-core/src/components/popover/Popover.tsx
```diff
@@ -54,6 +54,16 @@ export interface PopoverProps {
...
+   * when it starts to overlap the target element (`<Popover.Element/>`).
+   */
+  flip?: boolean;
+  /**
+   * Whether to enable the fixed behaviour. This behaviour used to keep the `<Popover/>` at it's
```
hepiyellow:
```suggestion
   * Whether to enable the fixed behaviour. This behaviour is used to keep the `<Popover/>` at it's
```

yardnsm:
@hepiyellow there isn't a way to test that so much, as this logic is specific to PopperJS.

https://github.com/wix/wix-style-react/pull/2676 does have some e2e tests for it.

packages/wix-ui-core/src/components/popover/Popover.tsx
```diff
@@ -106,13 +116,38 @@ const getArrowShift = (shift: number | undefined, direction: string) => {
...
       gpuAcceleration: !shouldAnimate
+    },
+    flip: {
+      enabled: typeof moveBy === 'undefined' ? flip : !moveBy
+    },
+    preventOverflow: {
+      enabled: !fixed,
+    },
+    hide: {
+      enabled: !fixed
```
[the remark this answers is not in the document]
yardnsm: We'll want to use `hide` anyway in `preventOverflow` is `true`.

PR #1051, opened 2018-12-31
OptionFactory.createHighlighted - Supporting spaces in search term

giladartzi:
This change allows us to highlight values, even if the search term contains spaces.

Having an option with the value of "New York" -
Before this change, if the user has typed "ne", the result would be "**Ne**w York" (which is perfectly fine), but if the user typed "Ne Y", the string will not be highlighted at all.
After this fix, for the "Ne Y" input, the result will be "**Ne**w **Y**ork"

### DIFF
--- a/packages/wix-ui-core/src/components/address-input/AddressInput.spec.tsx
+++ b/packages/wix-ui-core/src/components/address-input/AddressInput.spec.tsx
@@ -799,36 +799,6 @@ describe('AddressInput', () => {
         });
     });
 
-    describe('Highlight', () => {
-        it('should highlight text in dropdown', async () => {
-            init({value: 'ne'});
-            GoogleMapsClientStub.setAddresses([helper.ADDRESS_1, helper.ADDRESS_2]);
-            driver.click();
-            driver.setValue('ne');
-            driver.keyDown('e');
-            await waitForCond(() => driver.isContentElementExists());
-            const highlightedStrings = driver.optionAt(0).getHighlightedStrings();
-            expect(highlightedStrings).toEqual({
-                highlighted: ['Ne'],
-                nonHighlighted: ['1 East Broadway, ', 'w York, NY, USA']
-            });
-        });
-
-        it('should highlight text in dropdown, even if it is not continuous', async () => {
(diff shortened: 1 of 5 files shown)


packages/wix-ui-core/src/components/address-input/AddressInput.spec.tsx
```diff
@@ -799,6 +799,36 @@ describe('AddressInput', () => {
...
+            driver.keyDown('e');
+            await waitForCond(() => driver.isContentElementExists());
+            const highlightedStrings = driver.optionAt(0).getHighlightedStrings();
+            expect(highlightedStrings).toEqual({
+                highlighted: ['Ne'],
+                nonHighlighted: ['1 East Broadway, ', 'w York, NY, USA']
+            });
+        });
+
+        it('should highlight text in dropdown, even if it is not continuous', async () => {
```
shlomitc: this functionality should be tested in `DropdownOption.spec.tsx`

packages/wix-ui-core/src/components/dropdown-option/DropdownOption.driver.ts
```diff
@@ -13,5 +13,14 @@ export const dropdownOptionDriverFactory = ({element, eventTrigger}) => {
     isDisabled: () => element && domUtils.hasStyleState(element, 'disabled'),
     getText: () => element && element.textContent,
     getElement: () => element,
+    getHighlightedStrings: () => {
```
shlomitc: please remove the `nonHighlighted`, return only the highlighted part

packages/wix-ui-core/src/components/address-input/AddressInput.spec.tsx
giladartzi: Agreed. Moved to `DropdownOption.spec.tsx`

packages/wix-ui-core/src/components/dropdown-option/DropdownOption.driver.ts
giladartzi: Done

PR #1054, opened 2019-01-02
wix-ui-core: Fix focusableHOC types

hepiyellow:


hepiyellow:
This gets a little complicated with static method on Component classes.
Need to consult someone.

PR #1063, opened 2019-01-10
wix-ui-test-utils: upgrade to unidriver@2

ophirg:


shlomitc:
@ophirg - can you explain what required a major change update?
Will this affect any of the existing usage?

ophirg:
@shlomitc The motivation for this upgrade was a new function, introduced in puppeteer 1.6.0 called [_isIntersectingViewport_](https://github.com/GoogleChrome/puppeteer/blob/v1.6.0/docs/api.md#elementhandleisintersectingviewport) which allows verifying whether components can be actually seen by users

This new functionality helps in testing the [_scrollIntoView_](https://github.com/wix/wix-style-react/blob/1e5b2f8717f3bfdb2531ecb0b6f341d6f51c4599/src/DataTable/DataTable.protractor.driver.js#L13) function, a one we were required to develop, as it appeared in the [_wix-style-react/DataTable_](https://github.com/wix/wix-style-react/blob/master/src/DataTable/DataTable.protractor.driver.js) test kit.

After consulting with @itaibenda we realized that the upgrade should not affect current consumers, as it is a new functionality which is used by a new [_Unidriver_](https://github.com/wix-incubator/unidriver/blob/master/src/lib/puppeteer/index.ts) function, called [_isDisplayed_](https://github.com/wix-incubator/unidriver/blob/192c6dd9bf3930d86a988e68a254ffb01473c954/src/lib/puppeteer/index.ts#L108-L111).

Let me know if you think otherwise.

itaibenda:
@shlomitc the consumers set their puppeteer version (if any) so they shouldn't be affected. If they use the new functionality they'll have to upgrade to a version later than 1.6.0

shlomitc:
great, thanks guys

PR #1068, opened 2019-01-14
InputWithOptions - onManualInput will be triggered upon blur

giladartzi:
@shlomitc what do you think about these test cases? the change in the actual code is pretty simple

### DIFF
--- a/packages/wix-ui-core/src/components/input-with-options/InputWithOptions.spec.tsx
+++ b/packages/wix-ui-core/src/components/input-with-options/InputWithOptions.spec.tsx
@@ -15,11 +15,11 @@ describe('InputWithOptions', () => {
     .createLegacyRenderer(inputWithOptionsDriverFactory);
 
   const options = generateOptions();
-  const createInputWithOptions = (props = {}) => (
+  const createInputWithOptions = (props = {}, inputProps = {}) => (
     <InputWithOptions 
       {...Object.assign({
         options: [],
-        inputProps: {}
+        inputProps
       }, props)}
     />
   );
@@ -33,16 +33,6 @@ describe('InputWithOptions', () => {
   it('should display content element', () => {
     const driver = createDriver(createInputWithOptions({options, forceContentElementVisibility: true}));
     expect(driver.isContentElementExists()).toBeTruthy();
-  });
-
-  it('should trigger onManualInput', () => {
-    const onManualInput = jest.fn();
-
-    const driver = createDriver(createInputWithOptions({options, onManualInput}));
-    driver.keyDown('a');
-    driver.keyDown('Enter');
-
-    expect(onManualInput).toHaveBeenCalled();
   });
 
   it('should trigger not onSelect by default, if item is selected', () => {
@@ -67,40 +57,6 @@ describe('InputWithOptions', () => {
     driver.optionAt(0).click();
 
     expect(onSelect).toHaveBeenCalledTimes(2);
-  });
-
-  it('should trigger onManualInput with the actual value', () => {
-    const onManualInput = jest.fn();
-    let inputValue = 'a';
-
-    const driver = createDriver(createInputWithOptions({
-      options,
-      onManualInput,
-      inputProps: {
-        value: inputValue
-      }
-    }));
-
-    driver.click();
-    driver.keyDown('Enter');
-    expect(onManualInput).toHaveBeenCalledWith('a');
-  });
-
-  it('should trigger onManualInput with the actual value even if option list is empty', () => {
-    const onManualInput = jest.fn();
-    let inputValue = 'a';
-
-    const driver = createDriver(createInputWithOptions({
-      options: [],
-      onManualInput,
-      inputProps: {
-        value: inputValue
-      }
-    }));
-
-    driver.click();
-    driver.keyDown('Enter');
-    expect(onManualInput).toHaveBeenCalledWith('a');
   });
 
   it('should not show options element upon blur', async () => {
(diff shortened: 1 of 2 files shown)


packages/wix-ui-core/src/components/input-with-options/InputWithOptions.spec.tsx
```diff
@@ -174,4 +130,87 @@ describe('InputWithOptions', () => {
...
+
+      // should call onManualInput once blurring a changed value
+      driver.setValue('a');
+      driver.keyDown('a');
+      driver.blur();
+      expect(onManualInput).toHaveBeenCalledWith('a');
+      expect(onManualInput).toHaveBeenCalledTimes(1);
+    });
+
+    it('should trigger onManualInput with the actual value even if option list is empty', () => {
```
hepiyellow: I think there should be no connection between the options and this behavior, so this test is redundant.

packages/wix-ui-core/src/components/input-with-options/InputWithOptions.spec.tsx
```diff
@@ -174,4 +130,87 @@ describe('InputWithOptions', () => {
...
+      // should not call onManualInput if onSelect has been triggered
+      driver.click();
+      driver.optionAt(0).click();
+      expect(onSelect).toHaveBeenCalledWith(options[0]);
+      driver.blur();
+      expect(onManualInput).not.toHaveBeenCalled();
+
+      // should call onManualInput once blurring a changed value
+      driver.setValue('a');
+      driver.keyDown('a');
```
hepiyellow: ### Improvement for Input.driver

Maybe the `setValue` method should also fire a `keyDown` event for each character (-:

packages/wix-ui-core/src/components/input-with-options/InputWithOptions.spec.tsx
```diff
@@ -174,4 +130,87 @@ describe('InputWithOptions', () => {
       expect(driver.optionAt(1).getText()).toBe('c');
     });
   })
+
+  describe('Focus and blur events', () => {
+    it('Should integrate focus event with inner input', () => {
```
hepiyellow: Cryptic message...
`should call onFocus when input is focused`

packages/wix-ui-core/src/components/input-with-options/InputWithOptions.spec.tsx
```diff
@@ -174,4 +130,87 @@ describe('InputWithOptions', () => {
...
+      const onFocus = jest.fn();
+      const driver = createDriver(createInputWithOptions({options}, { onFocus }));
+      driver.focus();
+
+      expect(onFocus).toHaveBeenCalledTimes(1);
+    });
+
+    it('Should integrate blur event with inner input', () => {
```
hepiyellow:
```suggestion
    it('Should call onBlur when input is blurred', () => {
```

packages/wix-ui-core/src/components/input-with-options/InputWithOptions.spec.tsx
```diff
@@ -174,4 +130,87 @@ describe('InputWithOptions', () => {
...
+    it('should trigger onManualInput with the actual value on Tab key', () => {
+      const driver = getDriver();
+
+      driver.click();
+      driver.keyDown('Tab');
+      expect(onManualInput).toHaveBeenCalledWith('a');
+      expect(onManualInput).toHaveBeenCalledTimes(1);
+    });
+
+    it('should trigger onManualInput with the actual value on blur', () => {
```
hepiyellow:
```suggestion
    it('should trigger onManualInput when input is blurred given that the user had editted the input', () => {
```

packages/wix-ui-core/src/components/input-with-options/InputWithOptions.spec.tsx
```diff
@@ -174,4 +130,87 @@ describe('InputWithOptions', () => {
...
+      expect(onManualInput).not.toHaveBeenCalled();
+
+      // should call onManualInput once blurring a changed value
+      driver.setValue('a');
+      driver.keyDown('a');
+      driver.blur();
+      expect(onManualInput).toHaveBeenCalledWith('a');
+      expect(onManualInput).toHaveBeenCalledTimes(1);
+    });
+
```
hepiyellow: Missing test case:
` should NOT trigger onManualInput when input is blurred given that the input value was NOT edited by the user`

packages/wix-ui-core/src/components/input-with-options/InputWithOptions.tsx
```diff
@@ -122,6 +122,14 @@ export class InputWithOptions extends React.PureComponent<InputWithOptionsProps>
     onFocus && onFocus(event);
   }
 
+  _onBlur = (event: React.FocusEvent<HTMLInputElement>) => {
+    if (this.isEditing) {
```
hepiyellow: I think that we might be able to remove the `isEditing` state, by checking:
```js
   if (this.inputRef.value !== this.inputProps.value)
```

> We would need to add this `inputRef`

The consumer does NOT provide an `onChange` callback for the input's value.
Instead, the consumer is being updated about the input's value, only when a "submit" is performed.

The `isEditing` comes to tell us that the native input's value is Different (dirty) from what the consumer is aware of (so we need to inform the consumer by calling `onManuuallyInput`).

packages/wix-ui-core/src/components/input-with-options/InputWithOptions.tsx
```diff
@@ -122,6 +122,14 @@ export class InputWithOptions extends React.PureComponent<InputWithOptionsProps>
...
+      this._onSelect(null);
```
hepiyellow: I'd prefer if we would call a new `_submit` method, which will reset the editing state and call `onManuallInput`,
instead of calling `_onSelect` which will call `onManualInput`.

PR #1072, opened 2019-01-15
fix(eslint): tests

mastertheblaster:


mastertheblaster:
The original change was reverted, so this fix is not needed anymore.

PR #1087, opened 2019-01-17
Storybook - update to stable version 4.1.6

mykas:


shlomitc:
@mykas - I'm closing, please re-open when you get to have the solution here and got time to fix

PR #1095, opened 2019-01-23
Image component implementation

RashelleKosib:
New component, based on the `` native element.
Includes features for resizing( object-fit behaviour) supported in explorer as well, errorImage and fallback image behaviour. 
In the future will also include placeholder abilities. 

please refer to the image/README doc for more information.


packages/wix-ui-core/src/components/image/image.meta.ts
```diff
@@ -0,0 +1,103 @@
...
+imageMetadata
+  .addSim({
+    title: 'renders provided src',
+    props: {
+      src: SRC,
+      alt: 'This is an image of 2 flamingos',
+    }
+});
+
+imageMetadata
```
JoMarton: this is irrelevant since statically the result of rendering is similar to the above simulation

packages/wix-ui-core/src/components/image/image.tsx
```diff
@@ -0,0 +1,94 @@
...
+  private handleOnLoad: React.EventHandler<React.SyntheticEvent<HTMLImageElement>> = e => {
+    this.setState({
+      status: this.state.status === 'error' ? ImageStatus.error : ImageStatus.loaded
+    });
+
+    this.props.onLoad && this.props.onLoad(e);
+  };
+
+  private handleOnError: React.EventHandler<React.SyntheticEvent<HTMLImageElement>> = e => {
+
```
JoMarton: remove empty line

packages/wix-ui-core/src/components/image/image.tsx
```diff
@@ -0,0 +1,94 @@
...
+  private resized = () =>
+    this.props.resizeMode === 'contain' || this.props.resizeMode === 'cover'
+
+  state = {
+    src: this.setSrc(),
+    status: ImageStatus.loading
+  };
+  
+  render() {
+    const { errorImage, resizeMode, srcSet, ...props} = this.props;
```
JoMarton: you should take `nativeProps` out and spread it on its own on native img.
Also - please rename `...props` to something like `...additionalProps`.

packages/wix-ui-core/src/components/image/image.tsx
```diff
@@ -0,0 +1,94 @@
...
+  resizeMode?: 'fill' | 'contain' | 'cover';
+  onError?: (event: React.SyntheticEvent<HTMLImageElement>) => void;
+  onLoad?: (event: React.SyntheticEvent<HTMLImageElement>) => void;
+};
+export interface ImageState {
+    src?: string;
+    status: ImageStatus;
+}
+ export class Image extends React.PureComponent<ImageProps, ImageState> {
+  private setSrc = () :string =>
```
JoMarton: this method **returns** `src` and doesn't **set** it. rename to something like `getSrc`

packages/wix-ui-core/src/components/image/image.tsx
```diff
@@ -0,0 +1,94 @@
...
+};
+export interface ImageState {
+    src?: string;
+    status: ImageStatus;
+}
+ export class Image extends React.PureComponent<ImageProps, ImageState> {
+  private setSrc = () :string => 
+  !!this.props.src ? this.props.src : this.srcSetExists() 
+
+  private srcSetExists = () :string =>
```
JoMarton: a method name as `somethingExists` will presumably return boolean. this is not the case so rename it

packages/wix-ui-core/src/components/image/image.tsx
```diff
@@ -0,0 +1,94 @@
...
+    !!this.props.srcSet ? this.errorImageExists() : FALLBACK_IMAGE
+
+  private errorImageExists = () :string =>
```
JoMarton: same as remark above

packages/wix-ui-core/src/components/image/image.tsx
```diff
@@ -0,0 +1,94 @@
...
+  !!this.props.errorImage ? this.props.errorImage : FALLBACK_IMAGE
+
+  private setErrorImage = () =>
```
JoMarton: same remark for method name. add return type

packages/wix-ui-core/src/components/image/image.tsx
```diff
@@ -0,0 +1,94 @@
...
+  !!this.props.src ? this.props.src : this.srcSetExists() 
+
+  private srcSetExists = () :string =>
+    !!this.props.srcSet ? this.errorImageExists() : FALLBACK_IMAGE
+
+  private errorImageExists = () :string =>
+  !!this.props.errorImage ? this.props.errorImage : FALLBACK_IMAGE
+
+  private setErrorImage = () => 
+    this.state.src === this.props.errorImage ? FALLBACK_IMAGE : this.errorImageExists()
```
JoMarton: this condition is derivative. the proper condition is to check for **error state**

packages/wix-ui-core/src/components/image/image.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+  private resized = () =>
```
JoMarton: no return type. 
you probably want to rename it to `isResized` if it returns boolean

packages/wix-ui-core/src/components/image/image.tsx
```diff
@@ -0,0 +1,94 @@
...
+      /> 
+    );
+  };
+
+  private handleOnLoad: React.EventHandler<React.SyntheticEvent<HTMLImageElement>> = e => {
+    this.setState({
+      status: this.state.status === 'error' ? ImageStatus.error : ImageStatus.loaded
+    });
+
+    this.props.onLoad && this.props.onLoad(e);
```
JoMarton: should `onLoad` be called when in error state?

packages/wix-ui-core/src/components/image/image.tsx
```diff
@@ -0,0 +1,94 @@
...
+  };
+
+  private handleOnError: React.EventHandler<React.SyntheticEvent<HTMLImageElement>> = e => {
+    
+    this.setState({
+        status: ImageStatus.error,
+        src: this.setErrorImage() 
+    });
+    this.props.onError && this.props.onError(e);
```
JoMarton: lets say src is broken (error callback will be called) and now errorImage is broken. Should `onError` callback be called twice?

packages/wix-ui-core/src/components/image/test-fixtures.ts
```diff
@@ -0,0 +1,3 @@
...Bo/IwAeV6OTk5M8H/gpAfDyFU0kElwn3t7eDuvEfe04yALQDR4VSSQSfCe+cOGCK4yiU/b+/XtbGC0Wi2YcRvYtDdpJ8wJRLBYXjRtGMdWo1+u2PHT+/Pl7lFJnmduhExMTPBrd+PDhQ08ACm+r+hANHj16xE1WQW6Zt+b69evOpFiyNgwmJiZciQxnomfPnuVSCU3TYt1uV09IuVxuUVVVJ90uPXny5EgiFWphp32Lolgwa+94PB4VBMHFu3QqEYvF+OGJ0vLHjx97ZC6bzdo0hX4QY6zgx5eCaAM8Bw0xR3+oV7BAxrlz526hP+SUh3Cvk654PM6l091ut0enUfSg4HcIaYiiWBoUhEHSENttHAeFO4oVc694PM6l091uN64DiMViXk2skizLPTOZnp7mmRtAlV+/fh2qWzEzMwOnxa06CVrlzZs3vRIWFuJB91dlWf5e0GARL00TQjo7Ozu2oiaRSPA6F+haVA8ODqqm33iZD+w5EomsGBHF2b6pbW9v9zoOkBGNRrklJRoNsix/Lynx8Kiu8aqyu7vbuxEcwMiIvAYtHK1mlIANbGBcEG67YFRvXgMRNIZL1gsYHx/nFvVWyt4DMDY25tlWgeC9vT1bxBkfH+/XOw3iw+aa1d3dXVsDN+h5bJXT2NiYb2Pr8+fPNjs/derUTYPi+jVm/YCgolv58uWLrV14+vRpvTvn1dja29tzN7awiyRJc6Iowil5vUk9AztBwExPnDhR1jQNWTIoEHS2q9++fYMT24ofHN6SkZ3gZVVVC4qi8FuLWG3cqmdzF7UCB4S+USQSQcLTp5WcS8BAb00URVRt3IglSRJykmdzF21Pp7a47fWTJ0/6TmWQARVFOdJunSRJXEZgUUH169ev/dvr5geSJPVzUthvWVGUoegEbt3IB74DDkVRuFMa3xHT6Oho3xETcgXaMoqiQPWBRkzwG0mSlo1xra/fgELs7++HHzGZmohEIjwa6xVZ9CGfIAgdJ/kDOdM0TR/yObOvlzAjOQ4+5DMFj46OYmrzQ8esGC7u7+8PP2a13A5C7A8ZdKuqipAcqEXTd07sVO/IyMiSqqoYjx7pXw1geqIoVg4PD0MFhdAALIAwU1g5ij97GH9hCMVmzXMMA8CqHIRC/NUGScz37zbGFB80Ab9Qt81z9v8BlelFJzK8InwAAAAASUVORK5CYII='
```
JoMarton: maybe find something less intricate

packages/wix-ui-core/src/components/image/image.driver.ts
```diff
@@ -0,0 +1,29 @@
...
+    getSrc: () => base.attr('src'),
+    getAlt: () => base.attr('alt'),
+    getImageStyleState: async styleState => {
+      const el = (await base.getNative()) as Element
+      const domUtils = new StylableDOMUtil(style, el);
+      return domUtils.getStyleState(el, styleState)
+    },
+    getSrcSet: () => base.attr('srcSet'),
+  };
+  };
```
JoMarton: indentation

packages/wix-ui-core/src/components/image/image.tsx
shlomitc: why having a `nativeProps` prop is better than just spreading the rest?

packages/wix-ui-core/src/components/image/image.tsx
RashelleKosib: I believe it should..  maybe what causing it to break twice (in a real case) can be something with the server they dispatch the url's. I would want to know.. besides, the user gets the event and chooses what to do with it.

packages/wix-ui-core/src/components/image/image.tsx
```diff
@@ -0,0 +1,96 @@
...
+  src?: string;
+  srcSet?: string;
+  alt?: string;
+  errorImage?: string;
+  resizeMode?: 'fill' | 'contain' | 'cover';
+  onError?: (event: React.SyntheticEvent<HTMLImageElement>) => void;
+  onLoad?: (event: React.SyntheticEvent<HTMLImageElement>) => void;
+};
+export interface ImageState {
+    src?: string;
```
JoMarton: indentation in this file is all over the place. please find a way to fix it

packages/wix-ui-core/src/components/image/image.tsx
```diff
@@ -0,0 +1,96 @@
...
+  private getSrcSet = () :string =>
+    !!this.props.srcSet ? this.getErrorImage() : FALLBACK_IMAGE
+
+  private getErrorImage = () :string =>
+  !!this.props.errorImage ? this.props.errorImage : FALLBACK_IMAGE
+
+  private getErrorSrc = () :string => 
+    this.state.src === this.props.errorImage ? FALLBACK_IMAGE : this.getErrorImage()
+
+  private isErrorSrc = () :boolean =>
```
JoMarton: why is this called `isErrorSrc` if it checks the status in the state

packages/wix-ui-core/src/components/image/image.tsx
```diff
@@ -0,0 +1,96 @@
...
+        src={this.state.src} 
+        srcSet={this.isErrorSrc( )? null : srcSet}
+        onLoad={this.handleOnLoad}
+        onError={this.handleOnError}
+      /> 
+    );
+  };
+
+  private handleOnLoad: React.EventHandler<React.SyntheticEvent<HTMLImageElement>> = e => {
+    this.setState({
```
JoMarton: why not just wrap everything here with if condition?
```
if (this.isErrorState()) {
  this.setState(...)
  this.callback()
}

File: packages/wix-ui-core/src/components/image/image.tsx [hunk 17]
```diff
@@ -0,0 +1,96 @@
...
+    });
+    this.isErrorSrc() ? null : this.props.onLoad && this.props.onLoad(e);
+  };
+
+  private handleOnError: React.EventHandler<React.SyntheticEvent<HTMLImageElement>> = e => {
+    this.setState({
+        status: ImageStatus.error,
+        src: this.getErrorSrc() 
+    });
+    this.isErrorSrc() ? null : this.props.onError && this.props.onError(e);
```
JoMarton: I don't understand the condition here `this.isErrorSrc()`.  You **know** that you're in error state since you are in `handleOnError` method plus you set the state just one command above this (although not synchronous).

If what you mean to not do something if you're **already** in error state, then the solution is the same as `handleOnLoad`- Just wrap in if condition

packages/wix-ui-core/src/components/image/image.driver.ts
```diff
@@ -0,0 +1,29 @@
...
+  getImageStyleState: (styleState: string | ImageStatus) => Promise<string>;
+  getSrcSet: () => Promise<string>;
+};
+
+export const imageDriverFactory = (base: UniDriver): ImageDriver => {
+  return {
+    ...baseUniDriverFactory(base),
+    getSrc: () => base.attr('src'),
+    getAlt: () => base.attr('alt'),
+    getImageStyleState: async styleState => {
```
shlomitc: I prefer not exposing this as the public driver.
instead, if this comes to query the loading state - expose an `isLoaded` method.
same for `getResizeMehtod`.

packages/wix-ui-core/stories/image/style.st.css
```diff
@@ -0,0 +1,20 @@
+:import {
```
shlomitc: wdyt about having this in the default themes folder?
https://github.com/wix/wix-ui/tree/master/packages/wix-ui-core/src/themes/default

packages/wix-ui-core/stories/image/image.story.tsx
```diff
@@ -0,0 +1,30 @@
...
+  component: Image,
+  componentPath: '../../src/components/image',
+
+  componentProps: {
+    ...style('root'),
+    src: 'https://www.gettyimages.com/gi-resources/images/CreativeLandingPage/HP_Sept_24_2018/CR3_GettyImages-159018836.jpg',
+    errorImage: 'https://cdn.pixabay.com/photo/2016/04/24/13/24/error-1349562__340.png',
+  }
+};
+export class ImageStory extends React.Component<{}> {
```
shlomitc: I don't see any use for this

packages/wix-ui-core/src/components/image/image.tsx
```diff
@@ -0,0 +1,100 @@
...
+            onLoad={this.handleOnLoad}
+            onError={this.handleOnError}
+          />
+        </div>
+      );
+    };
+
+    return (
+      <img
+        {...style('root', { resizeMode, loadState: this.state.status }, this.props)}
```
shlomitc: There is a duplication of code between the resize mode with the wrapper and this part.
Also - the duplication of `style('root')`. I prefer calling it only once and in the root of the returned node.

The minimal change will be refactoring to a common object that spreads on the img obj.

packages/wix-ui-core/src/components/image/image.browser.spec.tsx
```diff
@@ -0,0 +1,163 @@
...
+
+        it('displays image with given src', async () => {
+            const onLoadSpy = jest.fn()
+            const imageDriver = createDriver(<Image src={SRC} onLoad={onLoadSpy} />);
+
+            expect(await imageDriver.getImageStyleState('loadState')).toEqual('loading');
+            await eventually(async () => {
+                expect(onLoadSpy).toHaveBeenCalled();
+                expect(await imageDriver.getImageStyleState('loadState')).toEqual('loaded');
+                expect(await imageDriver.getSrc()).toEqual(SRC);
```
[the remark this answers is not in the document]
RashelleKosib: it was decided to include the state expectations within the behaviour tests, and because it all happens asynchronously -> I wanted to make sure both line 33,34 return the expected results together. If I take the  expect(await imageDriver.getSrc()).toEqual(SRC) outside of the eventually It can fail before the state succeed to change.

packages/wix-ui-core/stories/image/style.st.css
RashelleKosib: could you just explain a bit why? I am not familiar with the reason of creating stories inside the themes folder.

shlomitc: the current approach says each theme sits in its own folder.
this is the default one - so better have it there

PR #1096, opened 2019-01-23
`<ButtonNext/>` - Fix `driver.isButtonDisabled`, fix tests, add missing tests.

hepiyellow:
### isButtonDisabled
- It was broken, it never returned false. `base.attr('disabled')` always returns an empty string. See Unidriver [issue](https://github.com/wix-incubator/unidriver/issues/22)
- We should check the Stylable `disabled` state (not the native `disabled` attribute) since when `href` is provided then we don't put the `disabled` attribute.

### Tests
- isButtonDisabled was not tested at all.

### StylableUnidriverUtil

I added a partial implementation of StylableDOMUtil for use with Unidriver, trying to keep the same api.

### DIFF
--- a/packages/wix-ui-core/src/components/button-next/button-next.spec.tsx
+++ b/packages/wix-ui-core/src/components/button-next/button-next.spec.tsx
@@ -77,8 +77,26 @@ describe('ButtonNext', () => {
     });
   });
 
-  describe(`'disabled' prop`, () => {
-    it('when given should render component with tabIndex -1', async () => {
+  describe(`Disabled`, () => {
+    describe('isButtonDisabled', () => {
+      it('should NOT be disabled by default', async() => {
+        const driver = createDriver(<ButtonNext />);
+        expect(await driver.isButtonDisabled()).toBeFalsy();
+      });
+      
+      it('should be disabled when disabled is passed', async() => {
+        const driver = createDriver(<ButtonNext disabled />);
+        
+        expect(await driver.isButtonDisabled()).toBeTruthy();
+      });
+      
+      it('should be disabled when href is provided', async() => {
+        const driver = createDriver(<ButtonNext as="a" disabled href="wix" />);
+        expect(await driver.isButtonDisabled()).toBeTruthy();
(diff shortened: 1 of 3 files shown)


packages/wix-ui-core/src/components/button-next/button-next.spec.tsx
```diff
@@ -98,13 +116,34 @@ describe('ButtonNext', () => {
...
         const htmlTag = testContainer.componentNode.getAttribute('href');
         expect(!!htmlTag).toBe(false);
       });
     });
+
+    describe('disabled attribute', () => {
+      it(`should have 'disabled' attribute when disabled`, async () => {
+        testContainer.renderSync(<ButtonNext disabled />);
+        
+        await eventually(() => {
```
shlomitc: Why eventually is needed here?

packages/wix-ui-core/test/StylableUnidriverUtil.ts
```diff
@@ -0,0 +1,22 @@
...
+
+/**
+ * This is an implementation of StylableDOMUtil for Unidriver.
+ * 
+ * Work-In-Progress: Not all methods are implemented yet !
+ */
+export class StylableUnidriverUtil {
+  constructor(private style: RuntimeStylesheet) { }
+
+  public async hasStyleState(element: UniDriver, stateName: string, param: StateValue = true): Promise<boolean> {
```
shlomitc: The name element is confusing. It’s actually base (the unidriver).

And this means this won’t work, you need the native element and not the unidriver

I think you will need an adapter per environment and not just a single unidriver class

hepiyellow: 1. First of all I don't know where the name `base` came from. The UniDriver is representing an element. I don't like the name base, IT is confusing. (base of what?)
2. It will work. This `StylableUnidriverUtil` is a UniDriver. the `element` argument is already with an adapter. All adapters have the same interface, which is `UniDriver`.

PR #1100, opened 2019-01-24
wix-ui-core: (Tests Cleanup):  Use render async in tests that use Unidriver

hepiyellow:
Minor switch to using `ReactDomTestContainer.render` (which is async) instead of `ReactDomTestContainer.renderSync`.

This makes the tests forward-compatible with future version of react where the render may be async (According to React docs)


hepiyellow:
Waiting for weird build errors to get resolved.

PR #1114, opened 2019-01-29
wix-storybook-utils: add remounting capability to playground

shlomitc:
**What's in here**
This new feature will allow remounting the component on demand.

**Motivation**:
In cases where a prop has an effect on the component only mounted, this is a way to force a clean mounting with the given props.
For example, [Captcha](https://wix-wix-ui-core.surge.sh/?selectedKind=Components&selectedStory=Captcha&full=0&addons=0&stories=1&panelRight=0) component sets the theme only when loaded and script is loaded to page.

**How to play with it**
I create a new prop `valueSetOnMounting` for the test component which is copied to the state on initialization. Changing it's value will not have an effect until remounting
image

### DIFF
--- a/packages/wix-storybook-utils/src/AutoExample/components/index.js
+++ b/packages/wix-storybook-utils/src/AutoExample/components/index.js
@@ -5,6 +5,7 @@ import classnames from 'classnames';
 import { Cell } from '../../ui/Layout';
 import UIInput from '../../ui/input';
 import ToggleSwitch from '../../ui/toggle-switch';
+import TextButton from '../../TextButton';
 import Heading from '../../ui/heading';
 
 import ComponentSource from '../../ComponentSource';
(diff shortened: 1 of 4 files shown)


packages/wix-storybook-utils/src/AutoExample/components/index.js
```diff
@@ -69,7 +77,7 @@ const Toggle = ({ value, onChange }) => (
 
 Toggle.propTypes = {
   value: PropTypes.bool,
-  onChange: PropTypes.func,
+  onChange: PropTypes.func
```
argshook: @shlomitc overall looks good, just want to fyi about these commas that went missing, cause `yoshi lint` now gives error for them. file is `.js` but project has `tslint.json` and so `yoshi lint` runs `tslint`. i had to setup my editor to consider `.js` files as typescript

eventually i'll move it all to typescript so it wont be an issue, just want to fyi, that this might happen for projects "in transition period"

PR #1132, opened 2019-02-06
Add additional tests for case when no WSR was found.

yavorsky:
This PR adds additional tests + fixtures that checks no-full-wsr rule works when no wix-style-react was installed in the package.

For now, it's trying to find wsr version and according to the version, set `no-full-wsr-lib` rule. We have fixtures for cases when WSR is lower or higher 5.9.0, but have no case when no wix-style-react was installed at all.

In case of WSR is not present, we should use no-wsr-rule-lib rule, like before. It's something like fallback when we can't find any wsr lib.


packages/tslint-plugin-wix-style-react/tests/enableNoFullWsrLibRuleNoWsrDep.spec.ts
```diff
@@ -0,0 +1,101 @@
...
+  it('should fix property access require statement', () => {
+    const src = `const Button2 = require('wix-style-react').Button;`;
+    const output = `import Button2 from 'wix-style-react/Button';`;
+
+    const result = helper({ src, rule, rulesDirectory });
+    expect(result.errorCount).toBe(1);
+    expect(getFixedResult({ src, rule, rulesDirectory })).toEqual(output);
+  });
+
+  it.only('should not fix requiring all the lib', () => {
```
argshook:
```suggestion
  it('should not fix requiring all the lib', () => {
```

Issue #1139, opened 2019-02-15
wix-storybook-utils (LiveCodeExample) - playground breaks under some conditions


shlomitc:
You start with something simple
![image](https://user-images.githubusercontent.com/6093192/52865336-149f4f00-3145-11e9-8a94-c9e989ef1f6d.png)

Then you start typing and changing prop like `style` and you get this error:
![image](https://user-images.githubusercontent.com/6093192/52865409-3d274900-3145-11e9-9d33-ba4ebd0d6ae9.png)

The playground breaks and is not responsive anymore.

1. try solving this issue
2. make `reset` actually reset the component, (rerender by passing a new `key`?)

argshook:
fixed in #1152

PR #1152, opened 2019-02-18
<LiveCodeExample/> - rerendering LivePreview component on code change

nitayneeman:
This PR solves that issue which's described here: https://github.com/wix/wix-ui/issues/1139

Notice the solution is a workaround, probably until WSR will be upgraded to React v16.


mykas:
https://github.com/wix/wix-ui/issues/1139

PR #1160, opened 2019-02-19
<Popover/> - resolve Interactivity issues when not appended to parent

mykas:
This PR makes sure that onClickOutside wrapper ignores its own Popover content and does not fire onClickOutside callback

### DIFF
--- a/packages/wix-ui-core/src/components/popover/Popover.spec.tsx
+++ b/packages/wix-ui-core/src/components/popover/Popover.spec.tsx
@@ -1,10 +1,10 @@
 import * as React from 'react';
-import {Simulate} from 'react-dom/test-utils';
-import {queryHook} from 'wix-ui-test-utils/dom';
-import {Popover, PopoverProps} from './';
-import {createModifiers} from './modifiers';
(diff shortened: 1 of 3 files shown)


packages/wix-ui-core/src/components/popover/Popover.spec.tsx
```diff
@@ -80,42 +90,34 @@ describe('Popover', () => {
...
+          const onClickOutside = jest.fn();
 
-        driver.clickOutsideOnContent();
-        expect(onClickOutside).toBeCalled();
+          const driver = createDriver(
+            popoverWithProps({
+              placement: 'bottom',
+              shown: true,
+              onClickOutside,
+              appendTo: 'parent',
```
shlomitc: should be `appendToValue`

packages/wix-ui-core/src/components/popover/Popover.spec.tsx
```diff
@@ -80,42 +90,34 @@ describe('Popover', () => {
...
+            }),
+          );
+
+          driver.clickOutsideOnContent();
```
shlomitc: rename to `clickOnContent`

PR #1162, opened 2019-02-19
DO NOT MERGE wix-ui-core created new component <WixMedia />

liorMar:
WixMedia Component uses the new Image Component to render Media Platform Images
by converting the object to a url using image-client-api


packages/wix-ui-core/src/components/wix-media/WixMedia.driver.private.ts
```diff
@@ -0,0 +1,28 @@
+import { UniDriver } from 'unidriver';
+import {
+  BaseUniDriver,
+  baseUniDriverFactory,
+} from 'wix-ui-test-utils/base-driver';
+import { StylableDOMUtil } from '@stylable/dom-test-kit';
+import style from '../image/image.st.css';
+
+export interface WixMediaDriver extends BaseUniDriver {
```
JoMarton: It seems like you can just reuse the `Image` comp driver completely instead of writing your own.

packages/wix-ui-core/src/components/wix-media/WixMedia.browser.spec.tsx
```diff
@@ -0,0 +1,173 @@
+import * as React from 'react';
+import { ReactDOMTestContainer } from '../../../test/dom-test-container';
+import { wixMediaDriverFactory } from './WixMedia.driver.private';
+import { MediaPlatformItem, WixMedia } from './WixMedia';
+import * as imageClientSDK from '@wix/image-client-api/dist/imageClientSDK';
+import * as eventually from 'wix-eventually';
+import { BROKEN_SRC, ERROR_IMAGE_SRC, SRC } from '../image/test-fixtures';
+import { FALLBACK_IMAGE } from '../image/image';
```
JoMarton: FALLBACK_IMAGE should have been exported from `image/index.ts`. could you fix it and consume from there?

packages/wix-ui-core/src/components/wix-media/WixMedia.tsx
```diff
@@ -0,0 +1,83 @@
...
+  height: number;
+  uri: string;
+  alt?: string;
+}
+
+export interface WixMediaProps {
+  mediaPlatformItem?: MediaPlatformItem;
+  width?: number;
+  height?: number;
+  errorMediaPlatformItem?: MediaPlatformItem;
```
JoMarton: The readme says you have `errorImage: string`.  the story has that prop as well...

packages/wix-ui-core/src/components/wix-media/WixMedia.driver.private.ts
liorMar: is this looks OK?
`
export interface WixMediaDriver extends BaseUniDriver {
  getSrc(): Promise<string | null>;
  getAlt(): Promise<string>;
  getLoadStatus(): Promise<string>;
}

export const wixMediaDriverFactory = (base: UniDriver): WixMediaDriver => {
  return imageDriverFactory(base);
};`
i don't want to export all the ImageDriver Functions

packages/wix-ui-core/src/components/wix-media/WixMedia.browser.spec.tsx
liorMar: sure

packages/wix-ui-core/src/components/wix-media/WixMedia.driver.private.ts
JoMarton: ah I see. If you mean this driver to be "public" (be exported from `/drivers/vanilla`) and be used by other component tests, you should remove the `.private` from the file name. 

Also, we didn't add a public driver for Image comp which is a mistake on our part. You should have been able to just use that driver.  Simplest course of action is to just keep this file as you wrote it and refactor when we add the `image.driver.ts`

JoMarton:
@liorMar @shlomitc 
Maybe this component should be named `wix-media-image`?

packages/wix-ui-core/package.json
```diff
@@ -48,6 +48,7 @@
   },
   "dependencies": {
     "@stylable/dom-test-kit": "^0.1.0",
+    "@wix/image-client-api": "^1.1942.0",
```
shlomitc: This is a private package in an open source project, so this must be removed or refactored into an open source lib

packages/wix-ui-core/src/components/wix-media/README.md
```diff
@@ -0,0 +1,58 @@
...
+## API
+
+**Props**
+
+**WixMedia** accepts 
+
+| name        | type       | default | required | description       |
+| ----------- | ---------- | ------- | -------- | ----------------- |
+| mediaPlatformItem | MediaPlatformItem | ✖ | ✔ | media platform item to be used as the source for the media |
+| width | number | mediaPlatformItem.width | ✖ | the width of the image |
```
shlomitc: Isn’t it redundant if it’s defined in the mediaPlatformItem?

packages/wix-ui-core/src/components/wix-media/README.md
```diff
@@ -0,0 +1,58 @@
...
+
+#### `MediaPlatformItem`
+
+| Property         |                 Type                  | default | Required | Description                              |
+| ---------------- | :-----------------------------------: | :------: | :------: | ---------------------------------------- |
+| width | number |   ✖   | Yes | width of the source image. |
+| height | number |   ✖   | Yes | height of the source image. |
+| uri | string |   ✖   | Yes | relative url of the image |
+| mediaType | 'picture' &#124; 'video' | ✖ |   Yes   | type of the the media item |
+| alt | string |   ✖   | No | alternative text for the image used ny screen readers |
```
shlomitc: Maybe this can be spread as the rest of the image props instead of uniquely adding it to the item?

packages/wix-ui-core/src/components/wix-media/WixMedia.driver.private.ts
```diff
@@ -0,0 +1,28 @@
...
+    const el = (await base.getNative()) as Element;
+    const domUtils = new StylableDOMUtil(style, el);
+    return domUtils.getStyleState(el, styleState);
+  };
+
+  return {
+    ...baseUniDriverFactory(base),
+    getSrc: () => base.attr('src'),
+    getAlt: () => base.attr('alt'),
+    getLoadStatus: () => getStyleState('loadState'),
```
shlomitc: Maybe it’s better to create “isLoaded” and “isError” instead

packages/wix-ui-core/src/components/wix-media/WixMedia.tsx
```diff
@@ -0,0 +1,83 @@
...
+  onLoad?(event: React.SyntheticEvent<HTMLImageElement>): void;
+}
+
+export interface WixMediaState {
+  src: string;
+  errorImageSrc: string;
+  alternativeText: string;
+}
+
+export class WixMedia extends React.PureComponent<
```
shlomitc: Add displayName to the component

packages/wix-ui-core/src/components/wix-media/README.md
liorMar: i did not explained it well but i was going for that the downloaded image will be cropped to the width/height props using MediaPlatform capabilities and not the browser. instead of downloading big image that later will be cropped to download already small one.

packages/wix-ui-core/src/components/wix-media/README.md
liorMar: see bottom comment

packages/wix-ui-core/src/components/wix-media/WixMedia.driver.private.ts
liorMar: good point. maybe isLoading as well

liorMar:
@JoMarton @shlomitc the reason for naming it wix-media and accepting an object instead of spreading it is for future accepting different types of medias
(video, vector art ...) i added the mediaType attribute to MediaPlatformItem but did not used it. i'm aiming to use the Video Component to. but don't want to limit the Interface to only images

shlomitc:
a video might have other concerns, fallbacks and more... I would try to focus on the current image wrapper and maybe later create a wrapper that handle more than the image media type

packages/wix-ui-core/src/components/wix-media/WixMedia.driver.private.ts
JoMarton: update: Image now has a public driver you should be able to use: https://github.com/wix/wix-ui/blob/master/packages/wix-ui-core/src/components/image/image.driver.ts

shlomitc:
@liorMar is this PR still active?

PR #1164, opened 2019-02-20
wix-ui-core migrating `<Popover/>` testkit to use unidriver

argshook:
part of unidrivification effort

popover.spec.tsx runs test suite twice:
1. old driver (Popover.driver.ts, sync)
1. unidriver (Popover.uni.driver.ts, async)

migration included proxying old methods to unidriver, because unidriver is unable to:

1. access `document` (popover requires, as it can be appended to window)
1. trigger arbitrary events (limited amount of events is possible with unidriver.  popover requires `mouseLeave`, `mouseDown`, `clickOutside` etc.)

---

for event triggering we suggest generic method (`trigger`) and not dedicated method
for each event (`click`).

---

Popover.driver.ts should be refactored, as the methods there are too
specific:

1. `mouseEnter` (sould be `openPopover`, or smthng)
1. `mouseLeave` (should be `closePopover`, or smthng)
1. `clickOutside`


packages/wix-ui-core/src/components/popover/Popover.uni.driver.ts
```diff
@@ -0,0 +1,35 @@
...
+
+    /** Returns `true` whether the target element (`<Popover.Element/>`) exists */
+    isTargetElementExists: async () => byHook('popover-element').exists(),
+
+    /** Returns `true` whether the content element (`<Popover.Content/>`) exists */
+    isContentElementExists: async () => byHook('popover-content').exists(),
+
+    mouseEnter: () => byHook('popover-element').hover(),
+
+    mouseLeave: async () => Simulate.mouseLeave(await base.getNative()),
```
nitayneeman: Would be better to use ReactBase here once it's implemented

PR #1177, opened 2019-02-28
wix-ui-core: Backoffice Theme: `<Button/>`: Limit button transition to relevant attributes

hepiyellow:


packages/wix-ui-core/src/themes/backoffice/button/button.st.css
```diff
@@ -41,7 +45,7 @@
   -st-mixin: sizeMedium, weightNormal;
 
   outline: none;
-  transition: all 100ms linear;
+  transition: background-color value(hoverTransitionParams), color value(hoverTransitionParams), border-color value(hoverTransitionParams);
```
argshook: @hepiyellow 
use explicit transition rules to avoid `:vars` trickery:

```css
transition-duration: 100ms;
transition-timing-function: linear;
transition-property: background-color, color, border-color;
```

hepiyellow: cool

PR #1180, opened 2019-02-28
Move wix-ui-test-utils to devDependencies

nirnatan:
Having wix-ui-test-utils as dependency creates a lot of warnings in our build.
Can we move it?

### DIFF
--- a/packages/wix-ui-core/package.json
+++ b/packages/wix-ui-core/package.json
@@ -64,8 +64,7 @@
     "tslib": "^1.9.3",
     "type-zoo": "3.1.1",
     "unidriver": "^2.0.1",
-    "wix-eventually": "^2.2.0",
-    "wix-ui-test-utils": "^1.0.0"
+    "wix-eventually": "^2.2.0"
   },
   "devDependencies": {
     "@storybook/addon-options": "4.0.0-alpha.14",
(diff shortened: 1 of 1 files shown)


EtaiG:
@shlomitc we get TONS of warnings, due to missing peer dependencies etc. which all stem from the fact that we are installing this 'dev dependency' when installing our dependencies.

To reproduce, just create a new project and add this as a dependency, run yarn
Can you split the testKit from this repo?
If not, can we split this repo into 2 (i.e. wix-ui-core and wix-ui-testkit, both will be required by wix-ui, and then we will only be dependent on wix-ui-core)?

hepiyellow:
I agree with the split

hepiyellow:
We can ha a ‘wix-ui-driver-utils’

shlomitc:
image
From what I see here, all warnings are coming from [`unidriver`](https://github.com/wix-incubator/unidriver).

Our components are exposed together with their `testkit` drivers, which are using `unidriver`, so this is a dependency of the library and not a dev-dependency.

An easy fix will be asking to remove the `peerDependencies` from `unidriver`

shlomitc:
closing the PR since this won't be merged, but the discussion can continue

PR #1184, opened 2019-03-11
Configure ES6 imports/exports with `dist/es` dir.

yavorsky:
This PR is a part of the work made to implement dynamic imports and tree shaking work.
I'll also open PRs for wix-style-react and wix-ui-backoffice.

Here we want to:
- Configure ES6 imports for `dist/es` directory. This feature is implemented for yoshi and we need to add `module` and `yoshi.experimentalTSTarget` fields for `package.json` + update yoshi to 3.x
- Add logic to lazy-load Tooltip (~290KB) using `withEllipsedTooltip`.
- Create `index.es.st.css` where we are re-exporting stylable files, but for `es` location. Then, if we want to import something for ES directory using WSR, we want to have correct component paths. So we need to import `index.es.st.css` for wix-ui-backoffice (will create separate PR for it).


mykas:
Hey @yavorsky , why do we need to reexport  `index.es.st.css` files? Maybe we can actually at last export css files with stylable loader instead of injecting those to bundle?

shlomitc:
it's more of a general question, but requires some thought - How should the dynamic import be tested in the components? `JSDom` tests are not helping because it's running in `node`. so I guess a browser test is needed. We can run `protractor` on the story page, but what should we test?

yavorsky:
@shlomitc I think we need to just check that component was rendered. Since we have e2e bundled with webpack, we need to build it with `module: esnext` and `NODE_ENV=production`. I'll check how we are doing it currently.

shlomitc:
@yavorsky - ok, so you want me to merge and you'll try to make it in a different PR or wait until we have a solution?

yavorsky:
We can merge it since we are not affecting our users with it. I'll add 1-2 tests for it in separate PR today.

Issue #1135, opened 2019-02-10
wix-ui-core: (Refactor) Better to apply `data-hook` props directly, rather then using stylable's runtime stylesheet function


hepiyellow:
This comes from an issue that some users have been experiencing.
If they don't use @stylable/jest` (or Yoshi) for running the jest tests,
then the default stylesheet is a mock function which does not pass the `data-*` attributes.

For example, in [`Popover.tsx`](https://github.com/wix/wix-ui/blob/5f517608701b2c7cd455f8973127c71ac7cf6811/packages/wix-ui-core/src/components/popover/Popover.tsx#L432
), instead of this:
```js
return (
      <Manager>
        <ClickOutsideWrapper handleClickOutside={this._handleClickOutside}>
          <div
            style={inlineStyles}
            {...style('root', {}, this.props)}
            onMouseEnter={onMouseEnter}
            onMouseLeave={onMouseLeave}
            id={id}
          >
```

We can do this:
```js
return (
      <Manager>
        <ClickOutsideWrapper handleClickOutside={this._handleClickOutside}>
          <div
            style={inlineStyles}
            {...style('root', {}, this.props)}
            data-hook={this.props[data-hook]}
            onMouseEnter={onMouseEnter}
            onMouseLeave={onMouseLeave}
            id={id}
          >
```

Anyway, we shouldn't mix apples and oranges.

shlomitc:
I don't think it's an issue that's the way the component works.
More than that - the stylable function is also adding other data attributes which some are essentials in order to make the testkit driver working correctly.

Closing

PR #1188, opened 2019-03-13
Applying `data-hook` prop directly on Popover

yaelz:
rather than using stylable's runtime stylesheet function (#1135 )


zivl:
@shlomitc - this will really help us to get ahead.
Also, it is part of resolving this issue #1135

shlomitc:
The correct way to solve this is by implementing the `@stylable/node` utility, to parse the stylable style function in your tests.

This is a very specific solution and not the way we work in the library, so I tend not to accept the PR

zivl:
@shlomitc I'd rather keep this open until after our meeting, after we'll show you some stuff you might think different :)

PR #1206, opened 2019-03-27
<FocusableHOC/> - add onblur and onfocus props callbacks 

mykas:
This PR adds additional focus & blur control for consumer when using a component that uses FocusableHOC. There are cases when consumer wants to control focusable components focus and also add additional actions. This change will add onBlur and onFocus props on top and give control over the focusability.

### DIFF
--- a/packages/wix-ui-core/src/hocs/Focusable/FocusableHOC.tsx
+++ b/packages/wix-ui-core/src/hocs/Focusable/FocusableHOC.tsx
@@ -84,7 +84,7 @@ export const withFocusable = Component => {
       const isFocused = this.state.focus || this.state.focusVisible;
       const isBecomeDisabled = !prevProps.disabled && this.props.disabled;
       if (isFocused && isBecomeDisabled) {
-        this.onBlur();
+        this.onBlur({});
       }
     }
 
@@ -102,14 +102,18 @@ export const withFocusable = Component => {
       this.setState({ focus: false, focusVisible: false });
     };
 
-    onFocus = () => {
+    onFocus = event => {
       const { onFocus } = this.props;
-      onFocus ? onFocus(this.focus) : this.focus();
+      onFocus
+        ? onFocus(event, { blur: this.blur, focus: this.focus })
+        : this.focus();
     };
 
-    onBlur = () => {
+    onBlur = event => {
       const { onBlur } = this.props;
-      onBlur ? onBlur(this.blur) : this.blur();
+      onBlur
+        ? onBlur(event, { blur: this.blur, focus: this.focus })
+        : this.blur();
     };
 
     render() {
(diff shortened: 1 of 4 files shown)


packages/wix-ui-core/src/hocs/Focusable/FocusableHOC.tsx
```diff
@@ -97,11 +97,21 @@ export const withFocusable = Component => {
       });
     };
 
-    onBlur = () => {
+    blur = () => {
       inputMethod.unsubscribe(this);
       this.setState({ focus: false, focusVisible: false });
     };
 
+    onFocus = () => {
```
argshook: @mykas 	
i'd probably do this:

```js
onFocus = event => {
  const { onFocus } = this.props;
  onFocus ? onFocus(event, { focus: this.focus, blur: this.blur }) : this.focus();
};
```

same for `onBlur`.

this achieves support for:
1. `withFocusable(<MyComponent onFocus/>)` // previously wasn't possible at all
1. `withFocusable(<MyComponent onFocus={event => console.log(event.target.whatever)}/>)` remain consistent with react world where event callbacks carry event object as first argument
1. second parameter with `{focus, blur}` and space for more in future for fine grained control

```
withFocusable(<MyComponent onFocus={(event, { focus, blur }) =>
  console.log(event.target.whatever)}/>
  blur();
  setTimeout(focus, 1000) // any async behavior is possible (like button with spinner which focuses itself once spinner loads data or smthng)
)
```

mykas: make sense. Thanks!

packages/wix-ui-core/src/hocs/Focusable/FocusableHOC.tsx
```diff
@@ -97,11 +97,25 @@ export const withFocusable = Component => {
...
-    onBlur = () => {
+    blur = () => {
       inputMethod.unsubscribe(this);
       this.setState({ focus: false, focusVisible: false });
     };
 
+    onFocus = event => {
+      const { onFocus } = this.props;
+      onFocus
+        ? onFocus(event, { blur: this.blur, focus: this.focus })
```
hepiyellow: ### Alternative

I think its more simple to run both `onFocus` AND `this.focus`.
And the parent can override the styles of the wrapped component to make it not show the focus-ring.
For example, a `<BannerButton/>` which composes a `<Button/>`.
The BannerButton would override the focus-ring styles of a Button (disabling the focus-ring look),
and BAnnerButton will be wrapped with FocusableHOC, and manage it's own styles.

hepiyellow: ### Why  { blur: this.blur, focus: this.focus } ?
I don't understand why the consumer need the `{ blur: this.blur, focus: this.focus }`?
In this example:  a <BannerButton/> which composes a <Button/>...
The Button's FocusableHOC's `focus` method must be called when `focusableOnFocus` is called, in order to maintain the `inputMethod` state.
So the consumer (BannerButton) must call `focus`.
What the use-case for having `blur` available?

PR #1207, opened 2019-03-27
Add getElementByDataHook protractor helper

boriskor:


packages/wix-ui-test-utils/src/protractor/protractor-helpers.ts
```diff
@@ -69,3 +70,5 @@ export interface ILocation {
 
 export const mouseEnter = async (element: WebElement | ILocation) => await browser.actions().mouseMove(element).perform();
 export const mouseLeave = () => mouseEnter({x: 1000, y: 1000});
+
+export const getElementByDataHook = (dataHook: string) => $(`[data-hook="${dataHook}"]`);
```
hepiyellow: @boriskor 
Maybe it should also receive a 2nd argument which is a wrapper.
```
export const getElementByDataHook = (dataHook: string, wrapper?: ElementFinder) => {
  const selector = `[data-hook="${dataHook}"]`;
  return wrapper ? wrapper.$(selector) : $(selector)
}

PR #1209, opened 2019-03-28
Add <Loadable /> component

yavorsky:
Please, check README.md: https://github.com/wix/wix-ui/compare/feature/loadable?expand=1#diff-28c61f2f6c67484c011484a3c202a2a7


packages/wix-ui-core/src/components/loadable/README.md
```diff
@@ -0,0 +1,45 @@
...
+
+// You probably want to specify child props and module structure in part of the loaded component.
+class LoadableTooltip extends Loadable<
+  TooltipProps,
+  { Tooltip: React.ComponentType<TooltipProps> }
+> {}
+
+const LazyLoadedTooltip = ({ shouldUseAsync, shouldLoadComponent }) => (
+  <LoadableTooltip
+    loader={() => shouldUseAsync ? import('../tooltip') : require('../tooltip')}
```
shlomitc: how webpack knows to handle this import vs. require depending on the `dist/es` module or the regular `dist`?

should everyone use this exact syntax to make the dynamic module and tree shaking work?

packages/wix-ui-core/src/components/loadable/Loadable.driver.ts
```diff
@@ -0,0 +1,57 @@
+import { ComponentFactory, BaseDriver } from 'wix-ui-test-utils/driver-factory';
+import { tooltipDriverFactory } from '../tooltip/Tooltip.driver';
+
+export interface LoadableDriver extends BaseDriver {
+  isLoaded(): boolean;
+}
+
+const tooltipSelector = element => {
```
shlomitc: `data-hook="tooltip"` for example is knowing something about your HOC content which can be anything, not just tooltip. so this driver is useless for any case which is not tooltip

packages/wix-ui-core/src/components/loadable/Loadable.spec.tsx
```diff
@@ -0,0 +1,121 @@
...
+        {Tooltip => {
+          return (
+            <Tooltip data-hook="tooltip" placement="top" content="kek">
+              {fallBackElement}
+            </Tooltip>
+          );
+        }}
+      </LoadableTooltip>,
+    );
+    expect(wrapper.isLoaded()).toBe(true);
```
shlomitc: following my above comment about the driver being too specific, I would make the change a bit differently.

1. mount the component exactly as you do now, but add also `moveArrowTo=10`
2. get from the loadable driver for the rendered content (can be html)
3. search for the `Tooltip`'s `data-hook`, and initialize the tooltip driver.
4. use the `Tooltip`'s driver to verify it's actually the tooltip that is rendered. for example: `expect((await driver.getArrowOffset()).left).toBe('10px');`

packages/wix-ui-core/src/components/loadable/Loadable.spec.tsx
```diff
@@ -0,0 +1,121 @@
...
+              <span data-hook="tooltip-child">Hey!</span>
+            </Tooltip>
+          );
+        }}
+      </LoadableTooltip>,
+    );
+    expect(wrapper.find(tooltipSelector).exists()).toBe(false);
+    wrapper.setProps({ shouldLoadComponent: true });
+    await eventually(() => {
+      wrapper.simulate('mouseEnter');
```
shlomitc: use the tooltip driver for mouseEnter and leave

packages/wix-ui-core/src/components/loadable/Loadable.driver.ts
```diff
@@ -0,0 +1,57 @@
...
+    return childFactory({
+      element: childSelector(element),
+      eventTrigger,
+      wrapper,
+    });
+  };
+
+  return {
+    exists: () => !!element,
+    isLoaded: () => getChildFactory().exists(),
```
shlomitc: I think this can be removed after changing tests according to my comments.

packages/wix-ui-core/src/components/loadable/README.md
```diff
@@ -0,0 +1,45 @@
...
+A `React.Component` which is designed to make lazy-loading easy.<br/>
+It will make part of the code bundled into a separate chunk and load after `shouldLoadComponent` will be set to `true`.
+It's based on the render props approach. Just use function wrapper as a child of Loadable component with loaded component as an argument.
+The main feature is that it could avoid some breaking changes for your codebase.
+
+React-loadable or other libraries for lazy loading is based on dynamic imports + Promises apprach. So it will render lazy loaded component only on the next tick. But if you want to move to separate chunks some code gradually, you would like to support both approaches - sync and async.
+
+
+## API
+`shouldLoadComponent` - A boolean property that triggers component from `loader` callback to be loaded.
```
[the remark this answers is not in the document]
yavorsky: Users are responsible. It will just start lazy-load when it become `true`.

packages/wix-ui-core/src/components/loadable/README.md
yavorsky: Webpack won't know. Users should control what to use `import` or `require` according to their needs.
For our components like `ProxyText`, we can give users options to use `shouldUseAsync` property and according to it load modules sync or async.

packages/wix-ui-core/src/components/loadable/Loadable.driver.ts
yavorsky: yes. that's why we have [https://github.com/wix/wix-ui/pull/1209/files/f561abf15e0bf086a768bef709c3baa90802e449#diff-3fa42051f16b36ba5a7da6f638bd609bR34](createLoadableDriverFactoryWithChild) creator, which could access any selector + driverFactory

shlomitc: I would split my comment into several parts.

1. usage of drivers.

All of our drivers are structured in the following format:
```js
export const componentDriver = ({element}) => {
  return {
    interactWithComponent: () => someAction()
  };
}
```

Drivers play two roles in our eco-system:
1. abstract common actions.
2. use as testkits to our users.

Any other format is not acceptable as it breaks the structure and will not play well with our automated testkit extraction.

2. Drivers should be generic
Functions like `loadableDriverFactoryWithTooltip` are too specific for a generic driver, since it's just one usage, and can be any kind of them.
In that case, the responsibility to create it should be in the hands of the component using it.

let's say that `CircularProgressBar` is using the loaded`Tooltip` to display an error message.
In that case, `CircularProgressBar` driver is the one responsible to create the `Tooltip` driver, and expose the right function to the consumer. For example:

```js
export const circularProgressBarDriver = ({element}) => {
  return {
    getErrorMessage: () => createTooltipDriver({element.querySelector('.tooltip')}).getContent()
  };
}
```

packages/wix-ui-core/src/components/loadable/Loadable.driver.ts
```diff
@@ -0,0 +1,57 @@
...
+    });
+  };
+
+  return {
+    exists: () => !!element,
+    isLoaded: () => getTooltipFactory().exists(),
+  };
+};
+
+export const createLoadableDriverFactoryWithChild = ({
```
shlomitc: According to the above, this doesn't fit our drivers' format, so I wouldn't want to expose it.
You can change it to our regular format (as in the previous comment).
Then - you can query for elements that are only inside the `Loadable` component.

for example - if you have a test that checks if the component is not yet loaded, then the driver's `isLoaded` function will verify that the `defaultComponent` is not rendered.
we usually add `data-hook`s internally in our components in order to get the ability to query the elements.

Since the `Loadable` component doesn't have any markers (because it just renders the children), it might be a bit tricky. maybe you can wrap the default component internally with a `span`?

packages/wix-ui-core/src/components/loadable/Loadable.driver.ts
yavorsky: But this is kinda specific component. It could return any kind of child and only user can specify what kind of child it will be. That's why we can't use some hardcoded selectors or drivers to access children. According to the API we have, we have no chance to pass different drivers..
Let's see at `circularProgressBarDriver `. It always contains tooltip, so we can use `tooltipDriver`. But loadable can contain anything.
It shouldn't create any additional `div`s, so we can't pass any data-properties to handle this. But we need to somehow check that component was loaded w/o any idea about selectors user's will use

shlomitc: The use case here is when you want to expose the component's state, but don't want to apply any behavior on your children, wrap them or mutate them.

My suggestion - pass this information in your render prop, so they can be applied to the rendered node. It gives control to the consumer to decide whether this meta-data should be applied or not.
If the consumer wants to test the behavior and need some hooks - no problem, if he doesn't, no problem :)

yavorsky:
@shlomitc good point. I exactly thought about the same. I think we can use
```
{
  'Tooltip': () => import('./Tooltip'),
  tooltipStyles: () => import('./tooltipStyles')
}
```
and have to
`{({Tooltip, tooltipStyles}) => <Tooltip style={tooltipStyles} />}` via render prop

shlomitc:
@yavorsky - yeah, looks good

PR #1211, opened 2019-03-31
<AutoTestkit /> - add unidriver support based on sections api

mykas:
This PR adds unidriver examples documentation support.


hepiyellow:
### Methods API
 
Methods API should be a generic thing. Not specific to a Unidriver Testkit (It is currently under `AutoTestkit/Unidriver`)

Also components have methods, and we should document them in the same way !

### Documenting Methods

Taking as a reference,  the [Oracle Java Documentation](https://docs.oracle.com/javase/7/docs/api/java/lang/Object.html#clone()).

#### Method Summary
This could be a table with:
- method name
- short description (
- Return type
- Link to Full method docs

#### Full method docs
A "block" that shows:
- Method name
- Method Signature
- Long description
- Return value description
- Other tags
  - Throws
  - Deprecation
  - See Also

@wuwa @milkyfruit 
Can we sit together and design how we should document methods?

@shlomitc FYI

mykas:
@hepiyellow I agree with you. Though I don't think that this PR is the scope of your suggestion.

milkyfruit:
@hepiyellow yes, let's meet.

packages/wix-storybook-utils/src/AutoTestkit/unidriver/index.tsx
```diff
@@ -0,0 +1,29 @@
...
+import { API } from './components/api-table';
+import { CodeExample } from './components/code-example';
+
+const driverNames = ['react', 'enzyme', 'puppeteer', 'protractor'];
+
+export const UnidriverTestkit = ({ component }) => (
+  <div>
+    {component.drivers.map(({ file, descriptor, error }, i) => {
+      if (error) {
+        return <span />;
```
argshook:
```suggestion
        return null;
```
what's the span for? affects layout maybe?

packages/wix-storybook-utils/src/AutoTestkit/unidriver/templates/index.ts
```diff
@@ -0,0 +1,30 @@
+import { UnidriverExample } from '../../../typings/code-example';
+import EnzymeExample from './enzyme';
+import ProtractorExample from './protractor';
+import PuppeteerExample from './puppeteer';
+import ReactTestUtilsExample from './react';
+
+const templates: UnidriverExample[] = [
```
argshook: minor, but: `UnidriverExample` name implies it's an example but in reality it's a template definition

but i see non unidriver is also named `Example` so leave it for now

hepiyellow:
@mykas
Also, should the return tyoe say one if:
Promise<boolean>, Promise<string>, etc...?

mykas:
@hepiyellow Should, but then I need to read the description of the method and parse return type. Only dev can provide return type in description part. For now Promise is enough to tell what is the return.

hepiyellow:
@mykas 
- That's true, only the developer knows the return value (Unless we migrate to TypeScript).
- Parsing the description is very easy, we can use JSDocs tags (e.g. `@returns`). I think we already support `@deprecated`.
- Some drivers return either nested drivers, or nested objects. These methods are not async.
  ```js
  export driver = (base) => {
     const option = index => {
        const opt = base.$$('.option').get(index);
        return  {
          getContent: ()=> opt.text();
        }
     }
     return {
       getText: base.text(), // This is async
       getOption: index => opt(index); // This is not async
     }
  ```

mykas:
@hepiyellow Hey, I will move all of those suggestions to a new issue so that we could iterate on next pull request.

PR #1215, opened 2019-03-31
wix-ui-core: Bump Major version to 3.x - Upgrade Unidriver 

hepiyellow:
### The Unidriver Breaking Change 
Unidriver upgrade introduces a small change in API, but it could  "slightly" break consumer's tests.
Unidriver `attr()` was fixed so that in case the attribute does not exist, it will return `null` and not an empty string.

This could break tests like this:
```js
it('should be disabled', ()=> {
   ...
   expect(driver.attr('data-disabled')).toBe('');
}
it('should NOT be disabled', ()=> {
   ...
   expect(driver.attr('data-disabled')).toBe('true');
}
```

We decided not to consider this as a breaking change. The chance of tests breaking is small.

### DIFF
--- a/packages/wix-ui-core/README.md
+++ b/packages/wix-ui-core/README.md
@@ -1,3 +1,7 @@
 # wix-ui-core
 
 ## [Using Icons](https://github.com/wix/wix-ui/blob/master/docs/icons.md)
+
+## Migration (Major Version)
+
+See [`docs/migration`](./docs/migration/index.md)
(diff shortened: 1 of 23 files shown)


shlomitc:
Let’s talk about it before merging. I don’t want to bump major version at the moment.

What’s the breaking behavior?

hepiyellow:
- Reverted wix-ui-test-utils to version 1 (add upgrade of unidriver in version 1).
- Changed this PR to NOT bump major version

PR #1216, opened 2019-04-01
<Tooltip/> - enhance functionalities

mykas:
This PR adds additional functionality to Tooltip component like a11y, unidriver testkit. Also this provides unit tests.


packages/wix-ui-core/src/components/tooltip/Tooltip.tsx
```diff
@@ -45,77 +41,89 @@ export interface TooltipState {
...
+  state = { isOpen: false };
 
-  state = {isOpen: false};
-
-  handleClickOutside() {
-    if (this.props.onClickOutside) {
-      this.props.onClickOutside();
-    }
-    if (this.props.shouldCloseOnClickOutside) {
+  handleClickOutside = () => {
```
argshook: @mykas 	please add `_` in front of private methods so that they don't show up in storybook

PR #1218, opened 2019-04-01
wix-ui-core: import `UniDriver` from `wix-ui-test-utils`

hepiyellow:
### What was done

- We import `UniDriver` from `wix-ui-test-utils`
- Remove `unidriver` from dependencies.
- (Remove base-driver, not in use, not related)

### Why
This is part of migrating to latest Unidriver version.

We realized that `wix-ui-core` has it's own dependency on `unidriver`.
While it is using the `baseUniDriver` from `wix-ui-test-utils` in each driver.

If we upgrade only the unidriver version in `wix-ui-test-utils`, then we get a TypeError. (Since `UniDriver` API has changed)

I already added today that `wix-ui-test-utils` will export `UniDriver`.

And in this PR I am importing `UniDriver` from `wix-ui-test-utils`.
So it makes sure that we use the same version.

I removed the dependency of `unidriver` from `wix-ui-core`.

After this is merged, we can upgrade the version of `unidriver` in `wix-ui-test-utils`.

## Discussion
- Should we do it as described above?
- Or should wix-ui-core sync versions by `peer-dependency`?


hepiyellow:
@argshook Actually, I don't really like that `wix-ui-test-utils` exports `UniDriver`.
I think its an anti-pattern. I don't really have a real argument, cause I Don't remember why.
Maybe it has something to do with clashing globals in TypeScript.
( I think I encountered it with `protractor`, which re-exports all Node types).

So I've decided to NOT export `UniDriver` from `wix-ui-test-utils`.
And have `wix-ui-core` depend on `@unidriver/core`.

See PR #1219 which replaces this one.

PR #1221, opened 2019-04-02
Captcha as required element

ronen-erez:
This change enables users to use the captcha as required element in a form.
Thanks


packages/wix-ui-core/src/components/captcha/Captcha.driver.ts
```diff
@@ -5,10 +5,10 @@ import {
 import {UniDriver} from 'unidriver';
 
 export interface CaptchaDriver extends BaseUniDriver {
-  /** clicks on the captcha checkbox challenge */
   getTheme: () => Promise<string>;
   getSize: () => Promise<string>;
   getLang: () => Promise<string>;
+  getRequired: () => Promise<string>;
```
shlomitc: `isRequired`, make sure it returns a boolean value

packages/wix-ui-core/src/components/captcha/Captcha.st.css
```diff
@@ -21,3 +23,10 @@
 .root .captcha {
     display: none;
 }
+/* a hidden check box that will give the captcha a required input element capabilities */
+.root .checkbox {
```
shlomitc: I would call it `requiredField`

packages/wix-ui-core/src/components/captcha/test-assets/CaptchaTestComponent.testDriver.ts
```diff
@@ -14,6 +14,11 @@ async function isCaptchaVerified() {
   return verifiedToken !== constants.verifiedTokenMark && verifiedToken.includes(constants.verifiedTokenMark);
 }
 
+async function isCaptchaRequired() {
+  const input = await $('input');
```
shlomitc: why not use the same driver strategy for it? reuse the component driver that checks if the captcha is required

packages/wix-ui-core/src/components/captcha/test-assets/CaptchaTestComponent.tsx
```diff
@@ -36,33 +36,46 @@ export class CaptchaTestComponent extends React.Component {
...
+            <div data-hook={constants.resetDataHook}>{constants.resetMark}</div>
+          )}
+          <button
+            onClick={this.resetCaptcha}
+            type="button"
+            data-hook={constants.resetButtonDataHook}
+          >
+            Click to reset captcha
+          </button>
+          <button type="submit" value="Submit Captcha">
```
shlomitc: I don't see this is in use in our tests, what is it for?

packages/wix-ui-core/src/components/captcha/Captcha.st.css
```diff
@@ -21,3 +23,10 @@
...
+    position: absolute;
+    opacity : 0.0;
+    bottom: 0;
```
shlomitc: is this the right placement for it? shouldn't it be in the center?
this is how it looks like now
image

packages/wix-ui-core/src/components/captcha/test-assets/CaptchaTestComponent.tsx
ronen-erez: if you want to manually test the required field than you can submit and see the results

shlomitc: so let's create another example component that just displays another captcha with this button, to separate it from the test component.
you can also remove the added e2e test (its covered in the JSDOM tests)

packages/wix-ui-core/src/components/captcha/Captcha.st.css
ronen-erez: i know.
is there a way to get the captcha position so i can relate the input to it? if we put center (what i have tried) it comes out in the center of the div ... out of the captcha element borders

packages/wix-ui-core/src/components/captcha/Captcha.st.css
ronen-erez: done

packages/wix-ui-core/src/components/captcha/Captcha.driver.ts
ronen-erez: done

packages/wix-ui-core/src/components/captcha/test-assets/CaptchaTestComponent.testDriver.ts
ronen-erez: can you pls elaborate? 
I assumed that required input is a tested feature and i verified we have an input.

PR #1222, opened 2019-04-02
Add index.ts for convenient named imports

yavorsky:
Allow named imports like `import { Avatar } from 'wix-ui-core'`


yavorsky:
@hepiyellow updated.

packages/wix-ui-core/src/components/deprecated/index.ts
```diff
@@ -0,0 +1,6 @@
+/* This file is autogenerated with scripts/named-export.js */
```
shlomitc: if it's auto-generated, maybe it should be in `.gitignore` and not committed?

packages/wix-ui-core/src/index.ts
```diff
@@ -0,0 +1,40 @@
+/* This file is autogenerated with scripts/named-export.js */
```
shlomitc: same comment here

packages/wix-ui-core/src/components/deprecated/index.ts
yavorsky: Actually, I'm not fully sure. It could be convenient for example for people who just want to review the structure of the repo and don't want to expect some 'magic'. A lot of question will be when you see `import {Comp} from 'wix-ui-core'`, but can't find index file in the repo. Also, we can track if the structure of the file is changed.
But if you have concerns to commiting this file, we can add it to .gitignore

shlomitc: It causes git conflicts and people tend to manually change it, so that's the downside of having it in git

hepiyellow: Another risk i’ve encountered, is that if the output path of the generating script changes by mistake, then the committed file becomes stale without anyone noticing. 

But actually i do think that commiting generated files is convenient. 
If not, then at least run the generator at post-install.

shlomitc: why do you need it in `post-install`? you're not using it in the library

PR #1229, opened 2019-04-03
wix-storybook-utils: create default sections when no `sections` array given

argshook:
this PR releases wix-storybook-utils 3.0.0

> there are visual changes that could be considered breaking, especially
> for users who utilize wix-storybook-utils e2e (which is a mistake-with-fix-in-progress)

we want to make docs look similar regardless if they use sections or
not.

this PR makes all .story.js files to use sections even if they're not explicitly set.

logic of the idea:
* if no `sections`, then render default premade `sections`
* if `sections` and `componentPath` (it's component story), then merge .story.js sections and predefined sections in wix-storybook-utils (provide defaults, allow override)
* if `sections`, then just render them

the result is that all component stories look like sections, but less config
this will make less code changes in wix-style-react

---

other internal changes

* stop using `<SingleComponentLayout/>` and create default sections when no `sections` given in storyConfig
* remove ErrorSection, instead assume react node and render it
* refactor StoryPage tests to fit the change


shlomitc:
Why not use puppeteer instead of introducing cypress? We had already some experiments on Wix and decided to go with puppeteer

argshook:
> Why not use puppeteer instead of introducing cypress? We had already some experiments on Wix and decided to go with puppeteer

cypress here is not the important change, there's not much about it.
not sure i'll leave it for good, it's just experiment.

i wanted to introduce lightweight browser testing, cause unit testing all the story quirks is complicated.
installed cypress on a whim just to familiarize, see if it works and such.

it does work and is pretty nice but not sure i'll keep it for long.

as it is not relevant for users, no testkits or anything, it's not really important anyway, imo.

shlomitc:
If it’s not an issue, let’s keep our standards across the different projects.
Wix is investing in puppeteer all the way, so I want to keep the consistency in our projects for better maintanece.

Feel free to experiment on side projects. If you think it’s a better tool than puppeteer then I would love to hear the reasons for

argshook:
> Wix is investing in puppeteer all the way

i wouldn't be so sure, some companies use cypress and are happy with it. we might some day need to make an adapter for unidriver.

anyway it's not an issue so i cleaned it up

PR #1235, opened 2019-04-05
<Tooltip/> - add disabled prop support

mykas:
This PR adds additional prop `disabled` which disables tooltip showup on mouse enter or focus event.

### DIFF
--- a/packages/wix-ui-core/src/components/tooltip/Tooltip.spec.tsx
+++ b/packages/wix-ui-core/src/components/tooltip/Tooltip.spec.tsx
@@ -130,6 +130,16 @@ describe('Tooltip', () => {
     });
   });
 
+  describe('`disabled` prop', () => {
+    it('[when] given should not show tooltip on mouse enter', async () => {
+      const children = 'kido';
+      const { driver } = render(tooltip({ children, disabled: true }));
+      expect(await driver.tooltipExists()).toBe(false);
+      await driver.mouseEnter();
+      expect(await driver.tooltipExists()).toBe(false);
+    });
+  });
+
   describe('`onClickOutside` & `shouldCloseOnClickOutside` props', () => {
     it('should call onClickOutside when clicked outside', async () => {
       const onClickOutside = jest.fn();
--- a/packages/wix-ui-core/src/components/tooltip/Tooltip.spec.tsx
+++ b/packages/wix-ui-core/src/components/tooltip/Tooltip.spec.tsx
@@ -138,6 +138,13 @@ describe('Tooltip', () => {
       await driver.mouseEnter();
       expect(await driver.tooltipExists()).toBe(false);
     });
+    it('[when] given should not show tooltip on focus', async () => {
+      const children = 'kido';
+      const { driver } = render(tooltip({ children, disabled: true }));
+      expect(await driver.tooltipExists()).toBe(false);
+      await driver.tabIn();
+      expect(await driver.tooltipExists()).toBe(false);
+    });
   });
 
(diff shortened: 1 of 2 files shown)


mykas:
True, but the question is how should button disable tooltip show up?

mykas:
We are thinking that in core the tooltip itself should have `disabled` prop to make it inactive, but in wix-style-react we would read Buttons props and disable it internally.

shlomitc:
You mean if you still want to display a tooltip on top of a disabled button?

mykas:
No. I mean that it is true that when Button is disabled -  the tooltip should not show up. Questions is: how to achieve this? One way is to let users do it both: on Button and Disabled. Another way @argshook suggested is to read Tooltips childs props and disabled it internally.

shlomitc:
I don’t like this magic of reading internal props. There can be Many use cases / different buttons and I don’t want to couple the tooltip to this specific disable behavior.


, please check how other libraries solve it and let’s pick the one that gives us most

mykas:
ok, I'll do a research on it.

argshook:
this is our tooltip:

```js
<Tooltip>
  <Button>
</Tooltip>
```

consider case:

```js
<Tooltip>
  <Button disabled>
</Tooltip>
```

the result is that button looks disabled, but tooltip still works
that's because currently tooltip doesn't know that it should be disabled.

simple approach could be

```js
<Tooltip show={false}>
  <Button disabled>
</Tooltip>
```

but this delegates work on the user. they might forget, not know about it, see issue on production which is too late.

to prevent this melody, we can add logic to the tooltip:

```pseudo
if
  children is known
  and has prop disabled, then
    do not render tooltip
else
  render tooltip
```

`children is known` means we have a whitelist (we don't handle ALL cases, we handle known cases)
`has prop disabled` means we peek into `children.props`, which is fine

---

this is a small thing that takes away some labor from the user and
prevents potential bugs. not sure why it should be considered magic

shlomitc:
I would like to go with more conventional ways. White list is too opinionated, hard to extend and not suitable for a core component

argshook:
> White list is too opinionated,

that's exactly the point: we're explicit and say what components disabled state to consider for tooltip, it's a good thing.

> hard to extend and

i'd say the opposite, should be just an array

> not suitable for a core component

agree, it should be defined within wix-style-react

---

there're no drastic changes here, it's a nice-to-have, easy-to-implement, no-drawbacks idea which can save some pingpongs during support

mykas:
Material-ui case:
Screenshot 2019-04-08 at 10 18 06

So they rely solely on html to disable hover functionality. What I also noticed that they  rely on children props internally to pass mouse or touch callbacks to parent div. So @argshook idea is not that bad. 

Though we can't rely on html tag solely to just disable interactivity. Our Buttons supports other kind of tags like `a` or `span` which does not have native `disabled`.

mykas:
`Semantic-ui case`:

They pass prop disabled and then pass it on the children too internally. One place to show the disabled state I guess is quite nice.

mykas:
Attlassian tooltip case: 

Relies on html tag. Which means that all abstracted cases will not work.

mykas:
So basically this is the conclusion:

The idea of having `disabled` state in core Tooltip is great, because as we seen in Material UI case - tooltip might need to be shown even on disabled button. 

Wix-style-react case does not have such cases so we need to control it internally. Reading from children props is not a bad idea so as simple as `disabled={children.props.disabled? true: false}` can work internally in wix-style-react tooltip

PR #1248, opened 2019-04-18
wix-ui-core: (UniDriver) - rename drivers to have `.uni` prefix

hepiyellow:
## Exported `drivers/vanilla`
These were exporting UniDrivers as `vanilla`. Which is a mistake.
- Now I export UniDrivers under `drivers/unidriver`
- Added deprecationLog when consumer import UniDrivers from `drivers/vanilla` (still exporting the same driver for backward compatibility)


packages/wix-ui-core/scripts/create-drivers-export.js
```diff
@@ -2,4 +2,5 @@
 const createExports = require('./driver-export');
 
 createExports('vanilla', '.driver');
+createExports('unidriver', '.uni.driver');
```
NitayRabi: Do we want to keep the direct exporting of drivers (and not wrap them similar to testkit exporting in WSR (wrapped with protractor/vanilla/enzyme/puppeteer wrapper)) ?

PR #1249, opened 2019-04-20
wix-storybook-utils: create persistent data url mechanism

shlomitc:
The `Playground` story in `wix-style-react` becomes very popular for prototyping and creating snippets.
With this mechanism, we can save data to the URL and send it to users instead of long code snippets.

### DIFF
--- a/packages/wix-storybook-utils/package.json
+++ b/packages/wix-storybook-utils/package.json
@@ -48,6 +48,7 @@
   },
   "dependencies": {
     "babel-runtime": "^6.25.0",
+    "base64-url": "^2.2.1",
     "classnames": "^2.2.5",
     "copy-to-clipboard": "^3.0.8",
     "downshift": "^3.1.4",
@@ -56,7 +57,7 @@
     "loader-utils": "^1.1.0",
     "lodash.merge": "^4.6.1",
     "prettier": "^1.16.2",
-    "query-string": "^6.0.0",
+    "query-string": "^6.4.2",
     "react-autodocs-utils": "^3.0.0",
     "react-collapse": "^4.0.3",
     "react-docgen": "^2.20.0",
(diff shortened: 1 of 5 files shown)


argshook:
this is not good, i wouldn't go with this approach, because:

1. multiple code examples on one page would overwrite each others changes
1. the urls will be filled with gibberish
1. base64 is ~30% longer than original (e.g. 10 characters is 14 characters in base64)
1. url length limit is around 2000, this means ~1500 chars before base64. current example in playground is 1378, we can minify to 585, which is already 1/3 of the limit
1. twiddling with window history imperatively can have unexpected issues
   when testing components (this is already the case with Tabs in wix-storybook-utils)

clearly a more sophisticated approach is required. perhaps localstorage

shlomitc:
1. that's why you have a configurable unique key which is up to the right page example to decide.
2. yes, but it's only for sharing code snippets, it won't show on every story.
3. don't forget that the original needs to be escpaed, so every `<` char short be encoded to a longer string. that's why I don't see an issue with base64
4. right, but see point 2, it's just for code sharing.
5. right, but that's the scenario, you do need to change the url to store the values.

another approach will be holding a server storing the code snippets, but then our documentation will be coupled with an external server which I'm not sure if it's a good idea

argshook:
> I don't see an issue with base64

the issue is not with base64, issue is with squeezing code snippets into url.

i don't like this cause it doesn't scale, but for short snippets it's ok.

> another approach will be holding a server storing the code snippets, but then our documentation will be coupled with an external server which I'm not sure if it's a good idea

yes, that would be the proper approach and it's the whole zelda (or now Brix) thing http://apps.wix.com/hack-photoshare/edita/

it's not far from actually using it, maybe someday :)

argshook:
closed by https://github.com/wix/wix-ui/pull/1852 & https://github.com/wix/wix-ui/pull/1850

PR #1251, opened 2019-04-22
<ToggleSwitch/> - fix knob positioning

gilmeir:
According to UX the spacing between the knob and the track should be 1px.

### DIFF
--- a/packages/wix-ui-core/src/components/toggle-switch/ToggleSwitchLayout.st.css
+++ b/packages/wix-ui-core/src/components/toggle-switch/ToggleSwitchLayout.st.css
@@ -15,7 +15,7 @@
   trackHeight: value(rootHeight);
   knobWidth: calc(value(rootHeight) / 1.2);
   knobHeight: calc(value(rootHeight) / 1.2);
-  knobMovementRange: calc(value(trackWidth) - value(knobWidth) - value(trackHeight) + value(knobHeight));
+  knobMovementRange: calc(value(trackWidth)/2 - value(knobWidth)/2 - 1);
 }
 
 .root {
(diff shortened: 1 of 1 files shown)


packages/wix-ui-core/src/components/toggle-switch/ToggleSwitchLayout.st.css
```diff
@@ -44,14 +44,7 @@
   align-items: center;
   justify-content: center;
   border-radius: calc(value(knobHeight) / 2);
-  transform: translateX(calc(value(knobMovementRange) / -2));
-
-  /* Workaround for Firefox, which doesn't center absolutely positioned flex items */
```
gilmeir: rethink, maybe check on older firefox

packages/wix-ui-core/src/components/toggle-switch/ToggleSwitchLayout.st.css
```diff
@@ -65,11 +58,11 @@
   height: 100%;
 }
 
-.root:checked::knob, 
+.root:checked::knob,
 :global([dir="rtl"]) .root::knob {
-  transform: translateX(calc(value(knobMovementRange) / 2));
+  transform: translateX(value(knobMovementRange));
```
gilmeir: check if I can put value() in a var

packages/wix-ui-core/src/components/toggle-switch/ToggleSwitchLayout.st.css
gilmeir: Checked on Firefox 60.0.1, works ok, so still removing this redundant code

packages/wix-ui-core/src/components/toggle-switch/ToggleSwitchLayout.st.css
gilmeir: skipped

PR #1252, opened 2019-04-22
<Icons/> - add small size icons

mosheKerbel:
https://github.com/wix/wix-style-react/issues/2639
@shlomitc 
@argshook 
@mykas 
in terms of checks, we are fine to merge with the failing one of the branch, it is checking stuff of ui-core that are not related to the PR.


argshook:
@mosheKerbel 
i understand im late to the party, but am just wondering, was something like `<Icon size="small"/>` considered?

mosheKerbel:
amm no
we are now thinking of another iteration - to split the icons into families, so i will also discuss it with Shlomi and keep u in the loop.

Issue #1264, opened 2019-04-25
`<Popover/>` - testkit method  getContentElement() does not find popover content properly when multiple popovers given


mykas:
Given such wrapper component for testkit:
```jsx
class App extends React.Component {
  render() {
    return (
      <div>
        <Popover
          dataHook="popover1"
          appendTo="window"
          showArrow
          shown={true}
          placement="right"
        >
          <Popover.Element>
            <Button onClick={() => this.toggle()}>Click me to toggle</Button>
          </Popover.Element>
          <Popover.Content>popover1</Popover.Content>
        </Popover>
        <Popover dataHook="popover2" showArrow shown={true} placement="right">
          <Popover.Element>
            <Button onClick={() => this.toggle()}>Click me to toggle</Button>
          </Popover.Element>
          <Popover.Content>popover2</Popover.Content>
        </Popover>
      </div>
    );
  }
}
```

Test:

```js
import React from 'react';
import { render, cleanup } from 'react-testing-library';

import { eventually } from '../eventually';

import { popoverTestkitFactory } from 'wix-style-react/dist/testkit';

import App from './App';

describe('App', () => {
  afterEach(cleanup);

  it('finds correct content text for popover1', async () => {
    const { container } = render(<App />);

    const driver = popoverTestkitFactory({
      dataHook: 'popover1',
      wrapper: container,
    });

    expect((await driver.getContentElement()).textContent).toBe('popover1');
  });

  it('finds correct content text for popover2', async () => {
    const { container } = render(<App />);

    const driver = popoverTestkitFactory({
      dataHook: 'popover2',
      wrapper: container,
    });

    expect((await driver.getContentElement()).textContent).toBe('popover2');
  });
});
````

Test `finds correct content text for popover1` will fail and will show content coming from the second popover component.

Both uni and simple drivers first search for available `popover-content` datahook first in the wrapper and only then in the body. 

UNI:
```js
 const queryDocumentOrElement = async (query: string) => {
    const elm = base.$$(query).get(0);
    if (await elm.exists()) {
      return elm;
    }
    return body.$(query);
  }
```

SIMPLE:
```js
const queryDocumentOrElement = (element, query) =>
  (element && element.querySelectorAll(query)[0]) ||
  (document && document.querySelector(query));
```

hepiyellow:
Please look at the solution in wsr Tooltip. 
[...]

PR #1266, opened 2019-04-26
add(wix-ui-test-utils): add additional argument to testkit factories

mykas:
This PR will add additional argument for testkit factories. This will solve some cases where testkit needs to determine components parts in document.body by searching specific dataHook.

https://github.com/wix/wix-ui/issues/1264


hepiyellow:
I don’t get it. 
It looks fishy...
Can you add a usage example?
Why isn’t the wsr Tooltip solution good enough here?

mykas:
It seemed kind of hackish for me to generate some kind of random content-data-hook and the search based on it for the data-hook. But now that I looked at it more it make sense. I am closing this instead of that Tooltip case.

shlomitc:

PR #1268, opened 2019-04-26
<Popover/> - change testkit method getContentElement handler

mykas:
This PR should fix this issue: https://github.com/wix/wix-ui/issues/1264

### DIFF
--- a/packages/wix-ui-core/src/components/popover/Popover.tsx
+++ b/packages/wix-ui-core/src/components/popover/Popover.tsx
@@ -169,6 +169,8 @@ export class Popover extends React.Component<PopoverProps, PopoverState> {
   portalNode: HTMLElement = null;
   stylesObj: AttributeMap = null;
   appendToNode: HTMLElement = null;
+  testId: string;
+  contentHook: string;
 
   popperScheduleUpdate: () => void = null;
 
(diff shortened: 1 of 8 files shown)


packages/wix-ui-core/src/components/popover/Popover.tsx
```diff
@@ -169,17 +169,30 @@ export class Popover extends React.Component<PopoverProps, PopoverState> {
...
-    shown: this.props.shown || false,
-  };
+  constructor(props) {
+    super(props);
+    this.state = {
+      isMounted: false,
+      shown: this.props.shown || false,
+    };
+
+    this.testId = Math.floor((1 + Math.random()) * 0x10000)
```
hepiyellow: I think in this case a simple counter would be good enough as an id. 
I’d rather not include random stuff. It could interfere with snapshot testing. (Regression)

packages/wix-ui-core/src/components/popover/Popover.tsx
```diff
@@ -169,17 +169,30 @@ export class Popover extends React.Component<PopoverProps, PopoverState> {
...
+      isMounted: false,
+      shown: props.shown || false,
+    };
+
+    this.testId = Math.floor((1 + Math.random()) * 0x10000)
+      .toString(16)
+      .substring(1);
+
+    this.contentHook = `popover-content-${props['data-hook'] || ''}-${
+      this.testId
```
mykas: Hey @Zombiefruit & @JoMarton  here we have a new change which is important for Popover. This will generate uniq `data-content-hook` on the target div, but then I start to get issues from `autotools sanity`:
```
 AssertionError: console error was called with:
 Warning: Prop `data-content-hook` did not match. Server: "popover-content--b835" Client: "popover-content--f1c2": expected error to not have been called
```

I understand that this is reacts hydration error ,but whenever I disable the hydration check in meta file on the composite component that gets this issue (AdressInput) https://github.com/wix/wix-ui/pull/1268/files#diff-c6c1d4163a4d7e880cd373cade4d5865R13 - other components start to fail on autotools sanity that are note even related to this change. 

Everything works if I remove the hydration disable part. Can you please help?

packages/wix-ui-core/src/components/popover/Popover.tsx
mykas: How this can interfere with snapshot testing?

hepiyellow:
We dont have snapshot testing yet. But it is common practice to have DOM snapshot regression tests. (Built in with jest). 
If dom has randomness, then you can not compare dom to last “baseline”. It will always fail.

mykas:
I've decided that wix-ui-core popover should not include random or incremental values in the data-content-hook. It might be possible for us to do it on wix-style-react side keeping mind that fact that we would not able to use `autotools sanity` on popover based components, but in general the id generation part only covers the issue when we forget in our composite components to pass the unique dataHook.

PR #1275, opened 2019-05-02
<Popover/> - add random id to data-content-hook

mykas:


mykas:
@argshook I think I have finally found one way to slip in the id. It looks ugly, but it was done before with environment variable anyway. If we think about it the test id only matters for the test phase when testkit is used. Other cases it can be undefined.

PR #1286, opened 2019-05-06
Remove requireindex, export eslint no-full-wsr-lib rule manually

aarnoldaas:
requireindex doesn't work with jest and we want to run tests on a package that contains this rule

### DIFF
--- a/packages/eslint-plugin-wix-style-react/lib/index.js
+++ b/packages/eslint-plugin-wix-style-react/lib/index.js
@@ -4,9 +4,9 @@
  */
 "use strict";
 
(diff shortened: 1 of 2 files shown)


ranyitz:
@shlomitc @argshook @mykas Can someone assist?

We want to test the `eslint-preset-yoshi` but unfortunately, this rule made some problems. @aarnoldaas fixed the problem in this PR.

argshook:
@ranyitz @aarnoldaas thanks for contribution!

ranyitz:
Thanks @argshook ❤️

Issue #1301, opened 2019-05-13
`<Popover/>` - appending to scrollParent does not work  when parent has  overflowY:auto or overflowX: auto.


mykas:
https://github.com/FezVrasta/popper.js/issues/535

PR #1304, opened 2019-05-15
<Popover/> - fix unexpected popover positioning

mykas:
This PR fixes this issue: https://github.com/wix/wix-ui/issues/1301


mykas:
@shlomitc it seems that this is a serious bug from Popper.js. Most of the people who use the Popper.js like for example Material UI guys or Bootstrap UI they just changed the default value of appendTo to `window` and added a description explaining it how to change it.

For example:
`Overflow auto and scroll
Tooltip position attempts to automatically change when a parent container has overflow: auto or overflow: scroll like our .table-responsive, but still keeps the original placement's positioning. To resolve, set the boundary option to anything other than default value, 'scrollParent', such as 'window':`

https://github.com/twbs/bootstrap/pull/26844

mykas:
Again we have the owner of Popper.js suggesting to append the Popover content to window and it should fix the issue: https://github.com/FezVrasta/react-popper/issues/57#issuecomment-331180720

PR #1308, opened 2019-05-16
`<CircularProgressBar/>` - Adding unidriver and refactoring unit tests to test both driver and unideriver

TalZa1810:
Refactoring also the tests to have a more generic layout


packages/wix-ui-core/src/components/circular-progress-bar/constants.ts
```diff
@@ -0,0 +1,5 @@
+export enum dataHooks {
+    progressIndicator = 'progress-indicator',
+    successIcon = 'success-icon',
+    errorIcon = 'error-icon',
```
NitayRabi: Implement usage within `CircularProgressBar.tsx`

Issue #1296, opened 2019-05-12
can't focus on a wix-ui-tpa button using ref


eranshabi:
It seems that the `ref` we get from `<Button>` doesn't provide `focus()` method so we can't focus on the button dynamically.

for example:
```
<ButtonNext ref={r => (this.r = r)} />
```

`this.r` won't have the `focus` property in this case.

PR #1310, opened 2019-05-20
convert button-next to class so focusableHOC will pass ref to it

eranshabi:


eranshabi:
https://github.com/wix/wix-ui/issues/1296

Issue #1303, opened 2019-05-15
`<Popover/>` - appending to predicate is not working and is not tested


mykas:
Popover appending to predicate is not working. Follow this PR to implement it properly: https://github.com/wix/wix-style-react/pull/1754

mykas:
Done.

PR #1313, opened 2019-05-24
<Popover/> - improve appendTo functionality

mykas:
This PR should improve and fix popover appendTo functionality by adding appending by predicate.

Fixes this issue:
https://github.com/wix/wix-ui/issues/1303


splintor:
This should be properly documented.

mykas:
It will be on all the storybook doc pages. This here is just implementation.

PR #1321, opened 2019-05-28
Button next with ref fix

eranshabi:


packages/wix-ui-core/src/components/button-next/button-next.spec.tsx
```diff
@@ -128,6 +128,12 @@ describe('ButtonNext', () => {
       });
     });
 
+    it('should render component with tabIndex -1 when disabled', async () => {
```
jonathanadler: The description should be changed

packages/wix-ui-core/src/components/button-next/button-next.spec.tsx
```diff
@@ -128,6 +128,12 @@ describe('ButtonNext', () => {
...
+      const ref = React.createRef<any>();
+      await testContainer.render(<ButtonNext ref={ref}/>);
+      expect(ref.current.wrappedComponentRef.innerComponentRef instanceof HTMLButtonElement).toEqual(true);
```
jonathanadler: Where did `wrappedComponentRef` come from? From the `Focusable`?

packages/wix-ui-core/src/components/button-next/button-next.tsx
```diff
@@ -49,9 +49,7 @@ class ButtonNextComponent extends React.Component<ButtonProps> {
     } = this.props;
     const htmlTabIndex = disabled ? -1 : rest.tabIndex || 0;
     const htmlHref = disabled ? undefined : href;
-    const reference = isStatelessComponent(Component)
-        ? undefined
-        : ref => (this.innerComponentRef = ref);
+    const reference = canPassRef(Component) ? ref => (this.innerComponentRef = ref) : undefined;
```
jonathanadler: Just to understand, why do we need to check if `canPassRef`?

eranshabi: because if it's a stateless component you can pass `ref` to it, only if it's class component or html tag

packages/wix-ui-core/src/components/button-next/button-next.spec.tsx
eranshabi: `wrappedComponentRef` is from `FocusableHOC` (I would add a type so it would be more clear but `FocusableHOC` is not exported)
then `innerComponentRef` comes from `ButtonNextComponent`

packages/wix-ui-core/src/components/button-next/button-next.spec.tsx
eranshabi: oops

packages/wix-ui-core/src/utils/index.ts
```diff
@@ -45,3 +45,5 @@ export const isReactElement = <T>(
 export const isStatelessComponent = Component =>
     !(Component.prototype && Component.prototype.render);
 
+export const canPassRef = Component =>
+    (Component.prototype && Component.prototype.render) || typeof Component === 'string';
```
jonathanadler: I would write: 
```
export const canPassRef = Component => !isStatelessComponent(Component) || typeof Component === 'string'
```

eranshabi:
> forwardRef



> I would really love it if we could use `forwardRef` here instead of using the class's inner property.
> In addition, Shlomi, and I agree with him, doesn't want to have multiple levels of references, meaning: `this.ref.curent.inner1.inner2.focus()`.
> Our goal is to have only one level, so users can just use `this.ref.focus()`.

can't use forwardRef because you need to support react 15

eranshabi:
is it possible not to use `inner1.inner2`?
I don't think FocusableHOC should know its ref has another inner ref, wdyt?

jonathanadler:
> is it possible not to use `inner1.inner2`?
> I don't think FocusableHOC should know its ref has another inner ref, wdyt?

That's what I meant. At the end it should work like this: `this.ref.focus()`

PR #1322, opened 2019-05-28
change image-client-api package to @wix/image-client-api

eranshabi:


eranshabi:
> This change should not be approved. wix-ui-core is an open source package and this takes a private registry package. Why do you need this change?

Because in `image-client-api` there are some `const`s that get inside our bundle

shlomitc:
What do you mean by some `const`s? what's the difference between the two packages?
They should be exactly the same package, only published to the public registry

eranshabi:
image-client-api was fixed so if you use latest you won't publish es6 code in the next version you publish

shlomitc:
Thanks!

PR #1329, opened 2019-06-05
`<Add/>`- Removing id from the Add icon

TalZa1810:


packages/wix-ui-icons-common/src/general/raw/Add.svg
```diff
@@ -1,6 +1,6 @@
 <svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" width="24" height="24" viewBox="0 0 24 24">
   <defs>
-    <polygon id="add-a" points="12 12 12 6 11 6 11 12 5 12 5 13 11 13 11 19 12 19 12 13 18 13 18 12"/>
+    <polygon points="12 12 12 6 11 6 11 12 5 12 5 13 11 13 11 19 12 19 12 13 18 13 18 12"/>
   </defs>
   <use xlink:href="#add-a"/>
```
jonathanadler: This `use` tag is referencing the `add-a` id, so this breaks the svg.
I think what we need is to move the first polygon out of the `defs` element, and remove all the rest.

PR #1333, opened 2019-06-10
#SCHED-10310 | dev (media-image) added resize options and quality props

poseneror:
The added options are ready for code review.
We've discussed the typing of the "options" prop in "MediaProps", would like another opinion:
It is currently stricted to MediaImageOptions interface defined in this library, the dilemma is if it should stay like that or be "any".
"any" would be less limited but will exposes the user to inner implementations.

### DIFF
--- a/packages/wix-ui-core/src/components/media-image/README.md
+++ b/packages/wix-ui-core/src/components/media-image/README.md
@@ -19,8 +19,7 @@ The **MediaImage** component accepts **MediaPlatformItem** object and convert it
 | onError | (event: ImageEvent) => void; | noop | ✖ | An event handler setting an Error state. |
 | errorMediaPlatformItem | MediaPlatformItem | ✖  | ✖ | media platform item to be used as the source for error media |
 | alt | string |   ✖   | No | alternative text for the image used ny screen readers |
-| resize | 'fill' / 'fit' |   'fill'   | No | change image sizing options inside the container |
-| options | MediaImageOptions | ✖ | No | additional options for the media image |
+| scale | 'fill' / 'fit' |   'fill'   | No | change image sizing options inside the container |
 
 
 #### `MediaPlatformItem`
@@ -31,20 +30,32 @@ The **MediaImage** component accepts **MediaPlatformItem** object and convert it
 | height | number |   ✖   | Yes | height of the source image. |
 | uri | string |   ✖   | Yes | relative url of the image |
 | mediaType | 'picture' &#124; 'video' | ✖ |   Yes   | type of the the media item |
+| options | MediaItemOptions | ✖ | No | additional options for the media image |
 
-#### `MediaImageOptions`
+#### `MediaItemOptions`
 
 | Property         |                 Type                  | default | Required | Description                              |
 | ---------------- | :-----------------------------------: | :------: | :------: | ---------------------------------------- |
-| focalPoint | FocalPointCoordinates |   ✖   | No | coordinates for positioning a cropped image (0-100) |
-| quality | number |   80   | Yes | the quality of the image (5-80) |
+| focalPoint | FocalPoint |   ✖   | No | coordinates for positioning a cropped image (0-100) |
+| quality | number |   80   | No | the quality of the image (5-80) |
+| unsharpMask | UnsharpMaskOptions |   ✖   | No | apply an unsharp mask to the image |
+| watermark | string |   ✖   | No | watermark manifest id |
+| upscaleMethod | MediaImageUpscaleMethod |   'auto'   | No | upscale method used |
+| isSEOBot | boolean |   false   | No | - |
+| filter | MediaItemFilters |   ✖   | No | filters applied to image |
 
-#### `FocalPointCoordinates`
+#### `FocalPoint`
 
 | Property         |                 Type                  | default | Required | Description                              |
 | ---------------- | :-----------------------------------: | :------: | :------: | ---------------------------------------- |
 | x | number |   ✖   | Yes | x coordinates for positioning a cropped image (0-100) |
 | y | number |   ✖   | Yes | y coordinates for positioning a cropped image (0-100) |
+
+#### `FocaMediaItemFilters`
+
+| Property         |                 Type                  | default | Required | Description                              |
+| ---------------- | :-----------------------------------: | :------: | :------: | ---------------------------------------- |
+| blur | number |   0   | No | image blur (0-100) |
 
 **Example:**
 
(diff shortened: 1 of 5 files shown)


packages/wix-ui-core/src/components/media-image/media-image.tsx
```diff
@@ -8,6 +9,21 @@ export interface MediaPlatformItem {
   uri: string;
 }
 
+export enum MediaImageResizing {
```
jonathanadler: Let's call this `MediaImageScale`

packages/wix-ui-core/src/components/media-image/media-image.tsx
```diff
@@ -8,6 +9,21 @@ export interface MediaPlatformItem {
...
+  FIT = 'fit',
+  FILL = 'fill',
+}
+
+export interface FocalPointCoordinates {
```
jonathanadler: Let's call this `FocalPoint`

packages/wix-ui-core/src/components/media-image/media-image.tsx
```diff
@@ -16,26 +32,34 @@ export interface MediaProps {
   onError?(event: React.SyntheticEvent<HTMLImageElement>): void;
   onLoad?(event: React.SyntheticEvent<HTMLImageElement>): void;
   alt?: string;
+  resize?: MediaImageResizing;
```
jonathanadler: So this will be called `scale`

packages/wix-ui-core/src/components/media-image/media-image.tsx
```diff
@@ -8,6 +9,21 @@ export interface MediaPlatformItem {
...
+  FIT = 'fit',
+  FILL = 'fill',
+}
+
+export interface FocalPointCoordinates {
+  x: number,
+  y: number,
+}
+
+export interface MediaImageOptions {
```
jonathanadler: Regarding your comment: 
Maybe we can add all option params, as stated here
https://github.com/wix-private/santa-core/tree/master/image-client-api#options
It hasn't changed in the last 7 months.
what do you think?
It might save us another PR in the future.
And I don't know how the server will behave if someone sends an unknown property (if we define it as `any`).
We don't want them to accidentally break their app. 

Btw, I thought you guys wanted to use the `filter.blur` option, or `unsharpMask`.

packages/wix-ui-core/src/components/media-image/media-image.tsx
```diff
@@ -16,26 +32,34 @@ export interface MediaProps {
   onError?(event: React.SyntheticEvent<HTMLImageElement>): void;
   onLoad?(event: React.SyntheticEvent<HTMLImageElement>): void;
   alt?: string;
+  resize?: MediaImageResizing;
+  options?: MediaImageOptions;
```
jonathanadler: I'm trying to think if I prefer having an `options` prop on the component, or add a `options` property to the `MediaPlatformItem` object.
I think the more natural place is the `MediaPlatformItem`, however I don't like nested objects in props;
It's less readable, and might be harder for developers to re-render the component.
What do you think?

packages/wix-ui-core/src/components/media-image/media-image.tsx
poseneror: Ok, seems legit. I think the image is already a bit blurred if you use a low quality query, and we haven't got any specification since it's more of a tech thing, I'll see how it looks and consider it with our designer when we get to the tpa implementation or "ours".

jonathanadler: Oh, no problem, use whatever's best for you. I just thought it works with `blur`.

packages/wix-ui-core/src/components/media-image/media-image.tsx
poseneror: I've considered that with Lior, we now believe that it makes a lot of sense to put the options on the MediaPlatformItem itself, since you might want different options for the "errorMediaPlatformItem" for example...

poseneror:
@jonathanadler fixed according to comments :)

packages/wix-ui-core/src/components/media-image/README.md
```diff
@@ -28,6 +30,32 @@ The **MediaImage** component accepts **MediaPlatformItem** object and convert it
...
+| filter | MediaItemFilters |   ✖   | No | filters applied to image |
+
+#### `FocalPoint`
+
+| Property         |                 Type                  | default | Required | Description                              |
+| ---------------- | :-----------------------------------: | :------: | :------: | ---------------------------------------- |
+| x | number |   ✖   | Yes | x coordinates for positioning a cropped image (0-100) |
+| y | number |   ✖   | Yes | y coordinates for positioning a cropped image (0-100) |
+
+#### `FocaMediaItemFilters`
```
jonathanadler: I think this is a typo

packages/wix-ui-core/stories/MediaImage/MediaImage.story.tsx
```diff
@@ -33,6 +33,32 @@ export default {
...
+          height: 400,
+          uri: '506418dbb019414f951a61670f3255a8.jpg',
+          options: {
+            filters: {
+              blur: 100,
+            }
+          }
+        },
+      },
       {
```
jonathanadler: If you look at the storybook of this pr:
https://wix-wix-ui-core-pr-1333.surge.sh/?activeTab=Usage&path=%2Fstory%2Fcomponents--media-image

You can see that changing the `mediaPlatformItem` option, doesn't change the image, and that's what I feared would happen.
It's because the options are nested inside the `mediaPlatformItem`, and this is something that might happen to users.
So I think maybe your initial suggestion of having a designated prop for this might work better.
What do you think?

packages/wix-ui-core/src/components/media-image/media-image.tsx
jonathanadler: @poseneror 
If you look at the storybook of this pr:
https://wix-wix-ui-core-pr-1333.surge.sh/?activeTab=Usage&path=%2Fstory%2Fcomponents--media-image

You can see that changing the mediaPlatformItem option, doesn't change the image, and that's what I feared would happen.
It's because the options are nested inside the mediaPlatformItem, and this is something that might happen to users.
So I think maybe your initial suggestion of having a designated prop for this might work better.
What do you think?

packages/wix-ui-core/src/components/media-image/README.md
poseneror: fixed.

PR #1336, opened 2019-06-11
add Check icon

eranshabi:


shlomitc:
Currently, the icons components are dedicated to the backoffice design system.
I don't want to mix here new icons that are not defined for it

if we need a different set of icons for the on-stage components, it should be in the `wix-ui-tpa` lib

PR #1338, opened 2019-06-12
fix(ui-core): playable-overlay

yteslenk:


packages/wix-ui-core/src/components/video/players/Playable.tsx
```diff
@@ -135,13 +136,13 @@ class PlayablePlayer extends React.PureComponent<
...
       src,
       playing,
+      poster,
       muted,
       title,
       showTitle,
       loop,
       volume,
       controls,
-      preload,
```
kupriyanenko: I don't think that we need to remove the ability to pass `preload` explicitly.

packages/wix-ui-core/src/components/video/players/Playable.tsx
```diff
@@ -209,6 +209,10 @@ class PlayablePlayer extends React.PureComponent<
 
     this.player.on(ENGINE_STATES.ENDED, () => {
       this.setState({ hasBeenPlayed: false });
+      if (controls) {
+        this._hidePlayableUI();
+      }
+      this.player.seekTo(0);
```
kupriyanenko: why do we need it?

packages/wix-ui-core/src/components/video/players/Playable.tsx
```diff
@@ -240,9 +260,13 @@ class PlayablePlayer extends React.PureComponent<
...
         />
-        {!this.state.hasBeenPlayed && poster && (
+        {!this.state.hasBeenPlayed && (
           <div
-            className={styles.cover}
+            className={
+              !poster
+                ? classNames(styles.cover, styles.transparentOverlay)
+                : styles.cover
+            }
```
kupriyanenko:
```
className={classNames(styles.cover, {styles.transparentOverlay: !poster})}
```

packages/wix-ui-core/src/components/video/Video.st.css
```diff
@@ -26,6 +26,10 @@
     cursor: pointer;
 }
 
+.cover.transparentOverlay {
+    background: transparent 50% 50% no-repeat;
```
kupriyanenko: why not just `background: transparent;`?

packages/wix-ui-core/src/components/video/players/Playable.tsx
yteslenk: useful knowledge)

kupriyanenko: check more here: https://github.com/JedWatson/classnames

packages/wix-ui-core/src/components/video/players/Playable.tsx
yteslenk: I deleted from this position pleload, cause I need to reinitialized it below
const preload = !poster ? 'metadata' : this.props.preload;

yteslenk: We don't lose an ability to pass it explicitly.

kupriyanenko: you can override this value if poster is not provided, but respect it in another case.

kupriyanenko: > We don't lose an ability to pass it explicitly.

from your code looks like we loose

packages/wix-ui-core/src/components/video/players/Playable.tsx
kupriyanenko: so, if we really need it, please add a comment in the code, and think, maybe we need to fix something on the playable side.

yteslenk: SeekTo I use to get a first frame of video.
Conditional check I could remove. If user removes controls in editor stage, there is nothing to remove.

packages/wix-ui-core/src/components/video/Video.st.css
yteslenk: Agree.

packages/wix-ui-core/src/components/video/players/Playable.tsx
kupriyanenko: > SeekTo I use to get a first frame of video.

for me, it looks like a hack, that you need to document in your code. And also looks like something that we need to fix on the playable side. please discuss with @BrooklynKing

kupriyanenko: My bad, nothing to do here

kupriyanenko:
@yteslenk I think in this PR we need to update playable version (package.json) as well.

PR #1340, opened 2019-06-12
backward/forward uni-driver

barak007:


shlomitc:
Build is failing
` src/unidriver/stylable-unidriver-util-legacy.ts(46,26): error TS2345: Argument of type '{ className: string; }' is not assignable to parameter of type 'string'.`

http://pullrequest-tc.dev.wixpress.com/viewLog.html?buildId=3011024&buildTypeId=FedInfra_Package_WixUiTestUtils&tab=buildLog&_focus=455#_state=422

barak007:
Passing now

packages/wix-ui-test-utils/package.json
```diff
@@ -38,6 +38,7 @@
     "registry": "https://registry.npmjs.org/"
   },
   "devDependencies": {
+    "@stylable/uni-driver": "^2.1.0",
```
shlomitc: this should be a dependency and not a dev-dependency as the unidriver module is being used from other libraries

PR #1348, opened 2019-06-17
Initial standalone generation

ranyitz:
Based on this RFC - https://github.com/wix-a/wsr-output-standalone

A PR for building a standalone version of WSR

### DIFF
--- a/packages/wix-ui-core/package.json
+++ b/packages/wix-ui-core/package.json
@@ -60,9 +60,7 @@
     "test:browser": "wix-ui-mocha-runner"
   },
   "dependencies": {
-    "@stylable/cli": "^2.1.0",
-    "@stylable/core": "^2.1.0",
-    "@stylable/dom-test-kit": "^0.1.0",
+    "@stylable/dom-test-kit": "2.1.0",
     "classnames": "^2.2.5",
     "eventemitter3": "^3.1.0",
     "hoist-non-react-methods": "^1.1.0",
@@ -84,7 +82,7 @@
   },
   "devDependencies": {
     "@storybook/react": "^5.0.0",
-    "@stylable/cli": "1.2.10",
+    "@stylable/cli": "^2.1.0",
     "@types/classnames": "^2.2.3",
     "@types/enzyme": "^3.1.9",
     "@types/jest": "^22.1.1",
--- a/packages/wix-ui-core/package.json
+++ b/packages/wix-ui-core/package.json
@@ -56,7 +56,7 @@
     "generate-stylable-components": "stc --srcDir=\"./dist/src/components\" --diagnostics --indexFile=index.st.css",
     "generate-es-stylable-components": "stc --srcDir=\"./dist/es/src/components\" --diagnostics --indexFile=index.es.st.css",
     "generate-stylable-hocs": "stc --srcDir=\"./dist/src/hocs\" --diagnostics --indexFile=hocs.st.css",
-    "build-standalone": "stc --srcDir='./src' --outDir='dist/standalone' --css",
+    "build-standalone": "./scripts/build-standalone",
     "test:browser": "wix-ui-mocha-runner"
   },
   "dependencies": {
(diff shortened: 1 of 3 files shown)


packages/wix-ui-core/package.json
```diff
@@ -56,9 +56,12 @@
     "generate-stylable-components": "stc --srcDir=\"./dist/src/components\" --diagnostics --indexFile=index.st.css",
     "generate-es-stylable-components": "stc --srcDir=\"./dist/es/src/components\" --diagnostics --indexFile=index.es.st.css",
     "generate-stylable-hocs": "stc --srcDir=\"./dist/src/hocs\" --diagnostics --indexFile=hocs.st.css",
+    "build-standalone": "stc --srcDir='./src' --outDir='dist/standalone' --css",
     "test:browser": "wix-ui-mocha-runner"
   },
   "dependencies": {
+    "@stylable/cli": "^2.1.0",
+    "@stylable/core": "^2.1.0",
     "@stylable/dom-test-kit": "^0.1.0",
```
ranyitz: bring version 2.1

packages/wix-ui-core/package.json
```diff
@@ -56,9 +56,12 @@
     "generate-stylable-components": "stc --srcDir=\"./dist/src/components\" --diagnostics --indexFile=index.st.css",
     "generate-es-stylable-components": "stc --srcDir=\"./dist/es/src/components\" --diagnostics --indexFile=index.es.st.css",
     "generate-stylable-hocs": "stc --srcDir=\"./dist/src/hocs\" --diagnostics --indexFile=hocs.st.css",
+    "build-standalone": "stc --srcDir='./src' --outDir='dist/standalone' --css",
     "test:browser": "wix-ui-mocha-runner"
   },
   "dependencies": {
+    "@stylable/cli": "^2.1.0",
```
ranyitz: Should be dev dependency

ranyitz: maybe...

shlomitc: I think it's ok to be a direct `devDependency`, why not?

packages/wix-ui-core/package.json
```diff
@@ -81,7 +82,7 @@
   },
   "devDependencies": {
     "@storybook/react": "^5.0.0",
-    "@stylable/cli": "1.2.10",
+    "@stylable/cli": "^2.1.0",
```
shlomitc: I'm not sure v2 syntax is backward compatible.
it might be, but maybe let's start using v1 just to make sure there will be no difference in the CSS creation

packages/wix-ui-core/scripts/build-standalone.js
```diff
@@ -0,0 +1,33 @@
+const path = require('path');
+const execa = require('execa');
+
+(async () => {
+  const outDir = path.join('dist', 'standalone');
```
shlomitc: we are also generating a `dist/es` folder, as you can see here https://github.com/wix/wix-ui/pull/1348/files#diff-5424cc8fdc700f83a28f75167ffcc053L57
should this be generated there as well?

packages/wix-ui-core/scripts/build-standalone.js
```diff
@@ -0,0 +1,33 @@
...
+
+  console.log('running tsc...');
```
shlomitc: `stc`

packages/wix-ui-core/package.json
ranyitz: @shlomitc I agree that it's a good thing to separate the two things. @barak007 says it's complex for him to make this change in `@stylable/cli` version 1.

packages/wix-ui-core/scripts/build-standalone.js
ranyitz: If people are consuming WSR using `/es` and expect tree shaking it means we need to think of a solution for the standalone as well. @shlomitc Is this the case?

packages/wix-ui-core/package.json
ranyitz: It was in dependencies when I wrote this comment (it is outdated)

packages/wix-ui-core/scripts/build-standalone.js
ranyitz: Done

PR #1349, opened 2019-06-17
rgbToHex- move util function to a different folder

TalZa1810:
move hexToRgb util function to "mixin" folder so that it will be compiled and added to the "dist" folder and prevent warning messages on WSR.


NitayRabi:
Looks good, there's no longer a warning?

TalZa1810:
I’ve compiled the `wix-ui-core` after the change, checked it on the local host and it worked fine. It added it to the `dist` folder. Afterwards, I copied the `dist` folder of the project to `wix-style-react` and run the build again. I haven’t gotten any warning messages related to it (on `wix-style-react`) so it seems to fix the issue :)

PR #1355, opened 2019-06-18
Icon updates

ZilvinasMaciulis:
Add 10 new icons
Replace and rename 2 existing icons:
NewFolder/NewFolderSmall  >  FolderAdd/FolderAddSmall


argshook:
Renaming icons is a breaking change, is it really required?

ZilvinasMaciulis:
Didn't know that. Let's leave the naming for new NewFolder just replace the
Icon itself. Never the less I think we should stick with suggested naming
in the future so other folder icons should still be named Folder, FolderAdd.

argshook:
> Never the less I think we should stick with suggested naming in the future so other folder icons should still be named Folder, FolderAdd.

i'm not against change in general, but we must consider it as breaking, because the file name is used for importing:
`import SomeGeneralIcon from 'wix-ui-icons-common/general/SomeGeneralIcon';`

so if we change then name, user will have to change code and that's considered a breaking change

PR #1357, opened 2019-06-18
fix(ui-core): Remove autotools-snap from tests commands

mosheKerbel:
@shlomitc @argshook


mosheKerbel:
approved by @shlomitc

PR #1359, opened 2019-06-19
InputWithOptions - avoid calling onManualInout when focus is lost to DropdownOption

giladartzi:


packages/wix-ui-core/src/components/input-with-options/InputWithOptions.tsx
```diff
@@ -158,6 +158,10 @@ export class InputWithOptions extends React.PureComponent<
     onBlur && onBlur(event);
   };
 
+  _onOptionMouseDown = () => {
+    this.isEditing = false;
```
giladartzi: When we are about to lose focus of the input, we exit "editing mode", in order to avoid calling `onManualInput`

packages/wix-ui-core/src/components/dropdown-content/DropdownContent.tsx
```diff
@@ -156,13 +164,16 @@ export class DropdownContent extends React.PureComponent<
...
       selectedIds,
       onOptionClick,
+      onMouseDown,
     } = this.props;
     const { hoveredIndex } = this.state;
 
     return (
       <div
         {...style('root', {}, this.props)}
+        data-hook="dropdown-content"
```
shlomitc: the `data-hook` should be provided through the props and not hard-coded in the component. so InputWithOptions will just propagate it

packages/wix-ui-core/src/components/dropdown-content/DropdownContent.driver.ts
```diff
@@ -16,5 +23,9 @@ export const dropdownContentDriverFactory = ({ element, eventTrigger }) => {
       const option = element ? getOptions()[index] : null;
       return dropdownOptionDriverFactory({ element: option, eventTrigger });
     },
+    dropdownContentMouseDown: () => {
```
shlomitc: isn't it enough to simulate the click on the item itself?
this might give a better indication of what the users actually intents so we don't need to expose this method

giladartzi: It's not enough, because JSDom click event does not trigger a mouseDown event as well. In the e2e test, we do test only a click, as a mouse click on the browse triggers a mouseDown event and then a click

packages/wix-ui-core/src/components/dropdown-content/DropdownContent.tsx
giladartzi: This data-hook is internal, only for testing. Unfortunately it has to be hard coded. The reason is that in Dropdown tests, we create the Dropdown driver, that eagerly creates an instance of DropdownContent driver (with the Dropdown's root as the element). At this point, the DropdownContent component does not exist yet (it will only be created once the dropdown is opened) - see the `getDropdownContentElement` function in DropdownContent.driver.tsx

PR #1363, opened 2019-06-20
add(ui-core) - <Popover/> - set zIndex from props

mosheKerbel:
@argshook 
could you plz pull it and help me figure out why my new example does appear in the storybook but does not apear in once the e2e tests are running?

@argshook 
neverminde - i just had to build again the modules locally and then run the tests
in ci they ran

### DIFF
--- a/packages/wix-ui-core/src/components/popover/Popover.st.css
+++ b/packages/wix-ui-core/src/components/popover/Popover.st.css
@@ -54,10 +54,6 @@
   margin: value(contentArrowSize);
 }
 
-.popover {
-  z-index: 1000;
-}
-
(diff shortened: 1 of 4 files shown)


mykas:
Looks good!

packages/wix-ui-core/src/components/popover/Popover.st.css
```diff
@@ -54,10 +54,6 @@
   margin: value(contentArrowSize);
 }
 
-.popover {
```
mykas: This shouldn't be deleted. Just remove z-index value and leave empty.

mosheKerbel:
merged.
the failing tests were screenshots that i approved becuase of the new z-index example

PR #1366, opened 2019-06-23
fix(ui-core) - <Tooltip/> zIndex

mosheKerbel:
pass zIndex top pop-over
@mykas


mosheKerbel:
build fail is good! it fails only because of snpashot of tooltip which now get also the next exmplae rendrered
https://eyes.applitools.com/app/test-results/00000251841017822280/?accountId=mE99eP8CUn9UTSXSv98L3OSaJ0NZmAcY6Wndx6WSHODdY110

mosheKerbel:
change looks pretty trivial to me and i've added tests.
if you find any issue with it plz let me know and i will fix.
merging.

Issue #1351, opened 2019-06-17
`wix-ui-core`- (Refactor) Renaming "mixin" folder name to "utils"


TalZa1810:
The naming of the folder is confusing and does not describe the files clearly. 
https://github.com/wix/wix-ui/tree/master/packages/wix-ui-core/src/mixins

Should do:
1. Renaming it to "utils".
2. Making sure that all the `calc` and `hexToRgb` usages (the paths) are updated with the new folder name.
3. Updating the `package.json` accordingly:
https://github.com/wix/wix-ui/blob/master/packages/wix-ui-core/package.json#L55
https://github.com/wix/wix-ui/blob/master/packages/wix-ui-core/package.json#L38

PR #1376, opened 2019-07-03
Export & allow to reuse css-utils

gilmeir:
* moved src/mixins/calc, src/mixins/hexToRgb --> src/utils/css-utils

* hexToRgb becomes public since it allows us to reuse existing colors when defining new ones with opacity.

* For example:
  Given the color D30 with 0.2 opacity, it will output the rgba value for D30-20.

* Those are js functions to be imported in css, like `-st-from` in stylable

* We should consider moving it outside to a separate ui-utils module


gilmeir:
#1351

packages/wix-ui-core/src/utils/css-utils/hexToRgb/index.ts
```diff
@@ -1,4 +1,4 @@
-module.exports = function(hex, opacity = 1) {
+export default function(hex: string , opacity:number = 1) {
```
TalZa1810: I used module.export as described in `stylable` [documentation](https://stylable.io/docs/references/formatters). Just make sure that it works properly when exporting it this way.

packages/wix-ui-core/scripts/create-named-export.js
```diff
@@ -6,5 +6,5 @@ const src = path.resolve(__dirname, '../src');
 const deprecatedSrc = path.join(src, 'components/deprecated');
 const ignoreDeprecated = `${deprecatedSrc}/**/*`;
 
-createExports(['components', 'hocs', 'createHOC'], src, { ignore: ignoreDeprecated });
+createExports(['components', 'hocs', 'createHOC', 'utils/css-utils'], src, { ignore: ignoreDeprecated });
```
TalZa1810: I recommend making sure that the files are transpiled properly to the `dist` and `es` directories so that `wsr` users won't get warning messages that the utils functions cannot be found.

shlomitc:
@gilmeir - let's close it?

gilmeir:
Closing it, instead of exporting hexToRgb, I added the colors to the color scheme [here](https://github.com/wix/wix-ui/pull/1385)

PR #1391, opened 2019-07-09
Move "@stylable/dom-test-kit" to devDeps.

yavorsky:
Don't think we need testkits as a dependencies.
It's creating some issues and conflicts with other dependencies.
For example,
image


ranyitz:
@yavorsky @argshook If you are exposing it through testkits (https://github.com/wix/wix-ui/search?p=2&q=dom-test-kit&unscoped_q=dom-test-kit) please add it as a `peerDependency`.

yavorsky:
@ranyitz good point

argshook:
@yavorsky 	@ranyitz 	but it's directly used in `wix-ui-core` testkits and because of peerDep is now failing. Why can't it be `devDep`?

ranyitz:
@argshook The short answer, it can be a dev dependency.

__The problem:__
**tl;dr** longer deployment time for fullstack projects.

When you run `npm install --production` you also get `dom-testkit`. This is of course not a big problem, just a bit longer installation on servers that actually run this command. (For example, fullstack projects)

I would say that the ideal solution would be to supply different packages for testing, which comes with a cost of harder maintenance.

It will also require the consumers of the library to change their code, and therefore I would not rush into doing it.

It's all just trade-offs.

ranyitz:
@argshook It is interesting though, why would it fail... ?

argshook:
@ranyitz 	it is typescript complaining about missing module for few components

```
src/components/avatar/avatar.spec.tsx(2,33): error TS2307: Cannot find module '@stylable/dom-test-kit'.
```

let's say this is mitigated but im guessing (didn't try) that it would still
fail. that's because some testkits use `@stylable/dom-test-kit` and
wix-ui-core tests components using its own testkits so it is a
`devDep` from wix-ui-core perspective

so perhaps it can stay as devDep for now, i would think a bit longer
deployment time is neglible, no? wdyt?

ranyitz:
@argshook I agree that there is might be more to lose than to earn.

yavorsky:
> wix-ui-core exposes testkits as public API, which use it (this constitutes a dep)

I would say it constitutes as a `peerDep`.
We can't use `@stylable/dom-test-kit` w/o any stylable infrastructure and `stylable-core` itself.

argshook:
i had to revert move to `devDep`, as builds started to crumble fast. need to investigate further

PR #1392, opened 2019-07-09
<Slider> - add aria-label

mosheKerbel:
@argshook @shlomitc 
will open an issue in wix-ui-santa, the only consumers, after merged.

### DIFF
--- a/packages/wix-ui-core/src/components/slider/Slider.meta.ts
+++ b/packages/wix-ui-core/src/components/slider/Slider.meta.ts
@@ -5,5 +5,7 @@ const sliderMetadata = Registry.getComponentMetadata(Slider);
 sliderMetadata.nonA11yCompliant = true;
 sliderMetadata.addSim({
   title: 'sliderSim',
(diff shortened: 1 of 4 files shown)


packages/wix-ui-core/src/components/slider/Slider.meta.ts
```diff
@@ -5,5 +5,7 @@ const sliderMetadata = Registry.getComponentMetadata(Slider);
 sliderMetadata.nonA11yCompliant = true;
```
argshook:
```suggestion
```
@mosheKerbel i think a11y test should work now?

mosheKerbel: yeah that was the whole point..a glitch
will fix

mosheKerbel:
ignoring the wix-ui-framework failure not related
also does the tooltop test that fails

mosheKerbel:
apprvoed by Arijus.

PR #1394, opened 2019-07-10
lock uniDriver

mosheKerbel:
@shlomitc @argshook 
i want to lock it so i will be able to test this UniDriver change locally when released, to go on the safe side...
https://github.com/wix-incubator/unidriver/pull/86


shlomitc:
In general, I don't think we should be locking versions. if someone breaks us - it's his fault and he should fix it. Also, by locking stuff you don't affect previously installed versions (which are using `^`), so they will break.
And in addition - if we lock stuff, we don't know if something is updated incorrectly and the feedback loop is slow.

So I say no

PR #1396, opened 2019-07-11
Correct import path

alifr:
`npm start` from `wix-ui/packages/wix-storybook-utils/` is failing to compile because the Sections module path is incorrect.

Not sure if this should go into `master` but that's where I discovered the issue.


argshook:
Hello, do you experience the same issue if you run `npm run build` at least once before `npm start`?

shlomitc:
been stale for a while so I'm closing. please reopen if still relevant

PR #1407, opened 2019-07-24
<LinearProgressBar> - Fix aria-valuenow

NitayRabi:
`aria-valuenow` prop should be the value passed to progress (and not relative progress) 

- [x] Create red test to describe issue
- [x] Fix issue

### DIFF
--- a/packages/wix-ui-core/src/components/linear-progress-bar/LinearProgressBar.tsx
+++ b/packages/wix-ui-core/src/components/linear-progress-bar/LinearProgressBar.tsx
@@ -43,12 +43,12 @@ const resolveIndicationElement = (props: LinearProgressBarProps) => {
 
   return wrapped(
     ProgressBarDataHooks.progressPercentage,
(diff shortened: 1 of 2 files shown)


packages/wix-ui-core/src/components/linear-progress-bar/LinearProgressBar.tsx
```diff
@@ -114,16 +114,16 @@ const getAriaAttributes = (
...
 > = (props: LinearProgressBarProps) => {
-  const {error, showProgressIndication} = props;
+  const { error, showProgressIndication } = props;
   const _props = normalizeProps(props);
   const success = _props.value === FULL_PROGRESS;
   return (
     <div
       {...getDataAttributes(_props)}
-      {...getAriaAttributes(_props)}
+      {...getAriaAttributes(props)}
```
NitayRabi: This is the prod change

PR #1413, opened 2019-07-29
Stylable compat 3

barak007:


shlomitc:
getting `src/unidriver/stylable-unidriver-util-legacy.ts(124,12): error TS2349: Cannot invoke an expression whose type lacks a call signature. Type '((base: ElementRemoteApi, stateName: string, param?: string | number | boolean | undefined) => Pr...' has no compatible call signatures.`

shlomitc:
ping?

PR #1418, opened 2019-07-30
<ButtonNext> - touch action

giladsegal:
Adding touch-action: manipulation removes the need for browsers to
delay the generation of click events when the user taps the screen
for touchscreen


NitayRabi:
Is there a way for this to be tested? if it has functional effects, maybe running e2e with user-agent to test functional effect?

giladsegal:
It only affects touch devices, not sure how we can test it.
I can add it directly to the 'clear button' in signature input but I think it will miss the point

PR #1421, opened 2019-08-02
feat(RadioButton): Add tabIndex prop

bordi:
Add `tabIndex` prop.

### DIFF
--- a/packages/wix-ui-core/src/components/radio-button/RadioButton.tsx
+++ b/packages/wix-ui-core/src/components/radio-button/RadioButton.tsx
@@ -41,8 +41,6 @@ export interface RadioButtonProps {
   disabled?: boolean;
   /** Sets the tabindex to the input */
   tabIndex?: number;
-  /** Disable focus on radio input */
-  disableFocus?: number;
   /** Sets the required status of the radio */
   required?: boolean;
   /** aria-label - Accessibility */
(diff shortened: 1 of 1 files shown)


packages/wix-ui-core/src/components/radio-button/RadioButton.tsx
```diff
@@ -39,6 +39,10 @@ export interface RadioButtonProps {
   label?: React.ReactNode;
   /** Sets the disabled status of the radio */
   disabled?: boolean;
+  /** Sets the tabindex to the input */
+  tabIndex?: number;
+  /** Disable focus on radio input */
+  disableFocus?: number;
```
jonathanadler: I think we don't need this prop at the moment.
We can start with just adding the `tabIndex` prop
I'll write you in an email why.

bordi: @jonathanadler removed this `prop`.

packages/wix-ui-core/src/components/radio-button/RadioButton.tsx
```diff
@@ -110,6 +113,7 @@ export class RadioButton extends React.Component<
           checked={checked}
           value={value}
           name={name}
+          tabIndex={tabIndex}
```
mykas: Now actually wondering about scenario if this is set and its undefined - will it default to inputs tabindex which is 0?

jonathanadler: From what I know React will ignore attributes that have undefined as their value

PR #1428, opened 2019-08-09
feat(wix-ui-core): add Control Button theme

ipanasenko:
`standard` and `dark` skins, `active` prop

https://zpl.io/bz66eGz
image


shlomitc:
Hey, themes should no longer be applied in wix-ui-core but rather in wix-style-react.

I agree it should be a new kind of button, maybe the name control bar is not that good. ToggleButton would probably describe better its use case

PR #1431, opened 2019-08-14
Signature input 

giladsegal:


packages/wix-ui-core/src/components/signature-input/signing-pad/SigningPad.tsx
```diff
@@ -10,25 +10,102 @@ export interface SigningPadProps
...
 
   static propTypes = {
+    /* Testing ID */
     'data-hook': PropTypes.string,
+    /* Color of the signature drawn over the canvas */
+    penColor: PropTypes.string,
+    /* Width of the signature line drawn over the canvas */
+    penWidth: PropTypes.number,
+    /* Callback to get imperative signature pad instace API  */
+    onPadInit: PropTypes.func,
```
NitayRabi: Why no `onInit` ? it's a `SignaturePad` prop, `pad` seems redundant IMO

packages/wix-ui-core/src/components/signature-input/signing-pad/SigningPad.tsx
```diff
@@ -10,25 +10,102 @@ export interface SigningPadProps
   extends WithSignaturePadProps,
     React.CanvasHTMLAttributes<HTMLCanvasElement> {
   'data-hook'?: string;
+  penColor?: IOptions['penColor'];
+  penWidth?: number;
+  onPadInit?(padAPI: Pick<SignaturePad, 'clear' | 'toDataURL'>): void;
+  canvasRef?(instance: HTMLCanvasElement): void;
 }
 
+const calculatePenColor = (penColor?: string) => penColor || 'black';
```
NitayRabi: How do we handle invalid values? (let's say a random string that doesn't represent color)

packages/wix-ui-core/src/components/signature-input/test/SignatureInputTestFixture.tsx
```diff
@@ -9,11 +9,18 @@ export const TEST_IDS = {
   CLEAR_BUTTON: 'signature-input--clear-pad',
 };
 
-export class SignatureInputTestFixture extends React.Component<{
+export interface SignatureInputTestFixtureProps {
   onClear?(): void;
-}> {
+  onPadInit?(): void;
```
NitayRabi: Now this is in the context of SignatureInput so maybe `pad` makes some sense but in general I think `onInit` is enough.

PR #1436, opened 2019-08-22
Update existing icons

ZilvinasMaciulis:
Existing Facebook and FacebookSmall icons need to be updated with uploaded SVG.


ZilvinasMaciulis:
@argshook really need to merge these

ZilvinasMaciulis:
@argshook also I need to let users know when the icons will be available for public use?

argshook:
we're releasing almost every day but if it's more urgent, we can make a release specifically for this

PR #1441, opened 2019-08-26
remove fdescribe

giladsegal:


shlomitc:
Lets add a lint rule to prevent this from happening

PR #1448, opened 2019-08-27
Add RadioButton onKeyDown prop

charmingelle:


packages/wix-ui-core/src/components/radio-button/RadioButton.tsx
```diff
@@ -13,6 +13,11 @@ export interface RadioButtonClickEvent
   value: string;
 }
 
+export interface RadioButtonKeyDownEvent
+  extends React.KeyboardEvent<HTMLInputElement> {
+  value: string;
```
jonathanadler: I'm not sure I understand this.
Why do we need to extend the original event?

charmingelle: Oh, sorry, I didn't pass <code>value</code> to <code>onKeyDown</code> call. Just corrected this. As far as I can see, all other events (in <code>onChange</code>, <code>onHover</code> methods) are extended with <code>value</code> prop for it to be available there I guess.

jonathanadler:
@charmingelle I think we just need to add a test for this (probably an e2e test, without `eyes`)
and then we'll merge

packages/wix-ui-core/src/components/radio-button/RadioButton.tsx
```diff
@@ -158,8 +166,11 @@ export class RadioButton extends React.Component<
     this.focusedByMouse = false;
   };
 
-  handleInputKeyDown = () => {
-    this.setState({ focusVisible: true });
+  handleInputKeyDown = (event: RadioButtonKeyDownEvent) => {
+    if (!this.props.disabled) {
+      this.setState({ focusVisible: true });
```
jonathanadler: I thinks it's better to leave this line (171) outside of the `if` block.
Because this is a change in behavior that I don't know if we should introduce.
It will be safer IMO.

charmingelle: Done)

PR #1460, opened 2019-09-04
wix-ui-core: fix(Input): use namespaced import specifier for `classnames`

argshook:
a small change which fixes error like this:

```
TypeError: classnames_1.default is not a function
    at Input.render (Input.js:31)
```

this error is not visible in wix-ui-core but rather in component
libraries that depend on wix-ui-core (e.g. wix-ui-adi).

code transpilers in those libraries fail to resolve such import

### DIFF
--- a/packages/wix-ui-core/src/components/input/Input.tsx
+++ b/packages/wix-ui-core/src/components/input/Input.tsx
@@ -1,7 +1,7 @@
 import * as React from 'react';
 import style from './Input.st.css';
 import { Omit } from 'type-zoo';
-import classnames from 'classnames';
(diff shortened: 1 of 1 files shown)


argshook:
@NitayRabi 	just fyi

NitayRabi:
How about adding a lint rule for this?

argshook:
> How about adding a lint rule for this?

good idea, however it turns out `lint` is not part of wix-ui-core build process at all, so enabling it would require a bit more work (as in, fix lint errors :) )

issue about it https://github.com/wix/wix-ui/issues/1461

PR #1473, opened 2019-09-12
Adds <TagsList/> component

zonzujiro:


packages/wix-ui-core/src/components/tags-list/TagsList.tsx
```diff
@@ -0,0 +1,58 @@
...
+  return (
+    <label data-hook={DataHook.Tag} className={nameOfClass} onChange={onChange}>
+      <input
+        data-hook={DataHook.TagInput}
+        className={style.tagInput}
+        type="checkbox"
+        checked={checked}
+        value={value}
+      />
+      {children}
```
NitayRabi: Do we have any children prop validation? can it get a whole node? prefix/suffix?

packages/wix-ui-core/src/components/tags-list/TagsList.tsx
```diff
@@ -0,0 +1,58 @@
...
+Tag.displayName = 'Tag';
+
+export const TagsList: React.FunctionComponent<TagsListProps> = ({
+  children,
+  className,
+}) => {
+  return (
+    <div className={className} {...style('root')}>
+      <div data-hook={DataHook.TagsList} className={style.tagsList}>
+        {children}
```
NitayRabi: Do we have any children prop validation? can it get any node? just `Tag` components?

packages/wix-ui-core/src/components/tags-list/test/TagsList.uni.driver.ts
```diff
@@ -0,0 +1,37 @@
...
+export const tagsListUniDriver = (base: UniDriver): TagsListUniDriver => {
+  const getInput = index =>
+    base
+      .$$(`${byDataHook(DataHook.TagInput)}`)
+      .get(index)
+      .getNative();
+
+  return {
+    ...baseUniDriverFactory(base),
+    getContent: () =>
```
NitayRabi: You use this to check the count of tags - change to `getTagCount`

packages/wix-ui-core/src/components/tags-list/test/TagsList.uni.driver.ts
```diff
@@ -0,0 +1,37 @@
...
+      base.$$(`${byDataHook(DataHook.Tag)}`).map(ud => ud.getNative()),
```
NitayRabi: I would change the approach 
1. Create Tag unidriver (with `getValue`, `getText`, `isChecked`, `click` etc...)
2. `getTags` method returns an array of Tag unidriver.

packages/wix-ui-core/src/components/tags-list/test/TagsList.uni.driver.ts
```diff
@@ -0,0 +1,37 @@
...
+    clickOnTagInput: async (index = 0) => {
```
NitayRabi: - change to `clickOnTagByIndex`, more informative and doesn't expose internal component input.
- Use `Tag` unidriver to invoke click and not native code

packages/wix-ui-core/src/components/tags-list/test/TagsList.uni.driver.ts
```diff
@@ -0,0 +1,37 @@
...
+import { UniDriver } from 'wix-ui-test-utils/unidriver';
+import { DataHook } from '../TagsList.helpers';
+
+export interface TagsListUniDriver extends BaseUniDriver {
+  getContent(): Promise<any[]>;
+  clickOnTagInput(): Promise<void>;
+  getIsChecked(): Promise<string>;
+}
+
+export const tagsListUniDriver = (base: UniDriver): TagsListUniDriver => {
```
NitayRabi: I would change the approach 
1. Create Tag unidriver (with `getValue`, `getText`, `isChecked`, `click` etc...)

packages/wix-ui-core/src/components/tags-list/test/TagsList.uni.driver.ts
```diff
@@ -0,0 +1,37 @@
...
+  return {
+    ...baseUniDriverFactory(base),
+    getContent: () =>
+      base.$$(`${byDataHook(DataHook.Tag)}`).map(ud => ud.getNative()),
+    clickOnTagInput: async (index = 0) => {
+      const native = await getInput(index);
+
+      return native.click();
+    },
+    getIsChecked: async (index = 0) => {
```
NitayRabi: - change to `isTagCheckedByIndex`
- Use Tag unidriver to check `checked` state.

packages/wix-ui-core/src/components/tags-list/test/TagsList.spec.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+        expect(isChecked).toBe(false);
+
+        await driver.clickOnTagInput();
+
+        expect(onChangeSpy).toHaveBeenCalled();
+
+        isChecked = await driver.getIsChecked();
+
+        expect(isChecked).toBe(true);
```
NitayRabi: No need for empty lines between each line.

packages/wix-ui-core/src/components/tags-list/TagsList.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+Tag.displayName = 'Tag';
+
+export const TagsList: React.FunctionComponent<TagsListProps> = ({
+  children,
+  className,
+}) => {
+  return (
+    <div className={className} {...style('root')}>
+      <div data-hook={DataHook.TagsList} className={style.tagsList}>
```
NitayRabi: Usually the root component is also the one with the `data-hook`, if you want to pass `className` please concat it to the one coming from the stylesheet.(and remove the wrapping div)

packages/wix-ui-core/src/components/tags-list/test/TagsList.spec.tsx
```diff
@@ -0,0 +1,59 @@
...
+            <Tag value="value">Some tag</Tag>
+          </TagsList>
+        );
+
+        const tags = await driver.getContent();
+        expect(tags.length).toBe(4);
+      });
+    });
+
+    describe('Tag', () => {
```
NitayRabi: Some missing tests in my opinion - 
- 'should render passed children'
- 'should pass `value` prop to `onChange` callback'
- 'should not break when clicked and no `onChange` prop passed' (onChange is not a function)

packages/wix-ui-core/src/components/tags-list/TagsList.tsx
zonzujiro: Right now - yes, any.

We really need validation of children? I like idea when `TagsList` is simple as possible.

packages/wix-ui-core/src/components/tags-list/test/TagsList.spec.tsx
zonzujiro: I have added test for rendering a child.

Regarding passing the value - it is a vanilla event so I don't think that such test is necessary.
Regarding testing missing `onChange` - we have default prop for such situation. Do we really need to test it?

NitayRabi: I think it's a good safety and quite a simple test to add

packages/wix-ui-core/src/components/tags-list/TagsList.tsx
NitayRabi: I agree about keeping it simple, But maybe using propTypes would do the trick

NitayRabi:
I still see you use `getNative` - avoid this, it breaks the whole unidriver abstraction approach (what returns from `getNative` is different in Puppeteer/JSDOM/Protractor)

packages/wix-ui-core/src/components/tags-list/TagsList.tsx
zonzujiro: Yeah, good idea. Also because `TagList` will be used in JS projects.

PR #1476, opened 2019-09-16
Update @stylable/dom-test-kit

NitayRabi:
Use compat class


NitayRabi:
Not relevant for time being.

PR #1484, opened 2019-09-20
wix-ui-core: feat(wuf): add component generator

argshook:
This PR adds new npm script which generates component in wix-ui-core.

`npm run generate`

template is at `.wuf/generator/template`

---

This is to prevent "copy-paste" technique when creating new components
and potentially copy-pasting bad practices from the past (like what
happened with tags-list)

---

the template is same as one in wix-style-react. Therefore, please check
and see if practices from there are applicable for wix-ui-core


argshook:
merging

it's non breaking and can be improved over time

PR #1487, opened 2019-09-23
changing layout of <Tag/>

zonzujiro:


packages/wix-ui-core/src/components/tags-list/Tag.tsx
```diff
@@ -11,7 +11,7 @@ export interface TagProps {
   checked?: boolean;
   value: string;
   label?: string;
-  children: React.ReactNode;
+  children: string;
```
piratus: So we do not allow anything else, like `<a>` or `<strong>`?

zonzujiro: According to specs it is can be only strings. 
`<a>` is a different tag and `strong` can be simulated with CSS.

Also, I believe that if we will need to change it - we can change.

zonzujiro: Forget, it can be reused by other developers - I will return `ReactNode`. Thank you for mention it. 👍

PR #1502, opened 2019-10-15
<AutoExample/> - Fix [object Object] in a list prop

EgoziE:
For example when there's a prop like this:
`type: PropTypes.oneOf([
    '',
    'compact',
    'compactSide',
    'uniformSide',
    'uniformFull',
  ]),`

The docs looks like this:
Screen Shot 2019-10-15 at 17 41 23

Adding a check for a string value fix the problem


argshook:
thanks for the fix, good catch!

unfortunately tslint isn't too happy about that `if`:
`ERROR: 47:7  curly  if statements must be braced`

however, i implemented your fix here:
https://github.com/wix/wix-ui/pull/1503/files#diff-e56bdda575f8670fd300a662d4b551acL47

thus, closing this PR.

thanks again, i know it's not too easy to dig around wix-storybook-utils :)

PR #1509, opened 2019-10-18
wix-ui-framework: compound components support

argshook:


argshook:
Closing, this was an interesting algorithmic challenge. It's not solved 100% but we anyway seem to not need it.

shlomitc:
I actually still think this feature has a lot of value.
Currently, We need to manually write the props of compound components  (like [here](https://wix-style-react.now.sh/?activeTab=Compound%20components%20API&path=%2Fstory%2Fcomponents-api-components--sidebar)) or sometimes just skip it (like [here](https://wix-style-react.now.sh/?activeTab=Description&path=%2Fstory%2Fcomponents-api-components--segmentedtoggle)) which could lead to mistakes.

Was this far from being finished?

argshook:
i don't remember exactly as it was a while ago, but there were some cases that weren't covered. I can get back on it on some sprint maybe?

argshook:
apparently i had some unpushed changes on my machine. There are 3 tests failing, which is not a big number, but i remember those were some not trivial ones.

shlomitc:
not prioritized, closing

PR #1510, opened 2019-10-18
Avatar - support for complex graphemes

zemlanin:
fixes https://github.com/wix/wix-ui-tpa/issues/145


packages/wix-ui-core/src/components/avatar/util.ts
```diff
@@ -1,3 +1,9 @@
+// `const ... = require()` instead of `import ... = require()`
```
piratus: And whats wrong with `import ... from ...`?

zemlanin: I think that `grapheme-splitter`'s [module definition](https://github.com/orling/grapheme-splitter/blob/0609d90dcbc93b42d8ceebb7aec0eda38b5d916d/index.js#L1741-L1743) is too confusing for TS:

- `import GraphemeSplitter from 'grapheme-splitter'` results in:
    ```
    error TS1202: Import assignment cannot be used when targeting ECMAScript modules. Consider using 'import * as ns from "mod"', 'import {a} from "mod"', 'import d from "mod"', or another module format instead.
    ```
- `import * as GraphemeSplitter from 'grapheme-splitter'` results in:
    ```
    error TS2497: Module '"[...]/wix-ui/packages/wix-ui-core/node_modules/grapheme-splitter/index"' resolves to a non-module entity and cannot be imported using this construct.
    ```
- adding `esModuleInterop` results in _a lot_ of
    ```
    error TS2349: Cannot invoke an expression whose type lacks a call signature. Type '{}' has no compatible call signatures.
    error TS2349: Cannot invoke an expression whose type lacks a call signature. Type 'typeof WixEventually' has no compatible call signatures.
    ```
    and fixing them is _way_ out of scope. Plus, `grapheme-splitter` still can't be `import`ed
    ```
    error TS2497: Module '"[...]/wix-ui/packages/wix-ui-core/node_modules/grapheme-splitter/index"' resolves to a non-module entity and cannot be imported using this construct.
    ```

mykas:
Guys, this pr just doubled our bundle size by 20Kb of GZIP. Can you do something on this?

mykas:
Do we really need another library just to split the name property?

zemlanin:
1. minified and gzipped `grapheme-splitter` is 7.5Kb
2. [unicode is hard](https://begriffs.com/posts/2019-05-23-unicode-icu.html) so _if we want_ to have `<Avatar>`s with initials in them, we'll need to somehow compute the first visible/comprehendible symbol/grapheme (instead of the first code point)

mykas:
I thinking of reverting this PR and rethinking the solution. Our heaviest core component is Popover which has popper.js calculation mechanism and it only weights 10Kb. It doesnt make sense that Avatar to support emojis adds up extra 7 (according to our perfer its 24KB)

mykas:
Can you check if js-emoji good library to use?

zemlanin:
Not only emojis, but also something like [Bengali `"মৌ"`](https://runkit.com/embed/3lpp1soipcfm) or [German `"Ü"`](https://runkit.com/embed/j1jdjflpyk6l)

mykas:
Ok for cases like German and Bengali I understand. What if for emojis we use library called is-emoji which is not actually 42Kb but [332b](https://bundlephobia.com/result?p=is-emoji@0.1.1) and check if the string is emoji - we dont split. Then for lanaguage cases we can think of something else

zemlanin:
bundlephobia doesn't count the size of (outdated) [`emoji-data-short.txt`](https://github.com/smurthas/is-emoji/blob/master/emoji-data-short.txt) because it is loaded with [`fs.readFileSync`](https://github.com/smurthas/is-emoji/blob/f7f70c7acade0083c758733fcae6004517868a25/index.js#L4-L5)

```
> curl -sL "https://unpkg.com/is-emoji@0.1.1/emoji-data-short.txt" | gzip -c | wc -c
    3702
```

So `is-emoji`:
1. is outdated;
2. not as small as it seems;
3. won't work in the browser (at least, without some webpack ✨magic✨)

zemlanin:
> can this feature be lifted to users of Avatar? as in, if they need support for emoji they opt-in to use 3rd party lib. We document how this can be done.

I'm okay with that. Also we could make `<InitialsAvatar>` (or whatever) to use it like this:

```jsx
{user.avatarUrl
  ? <Avatar src={user.avatarUrl} name={user.name} />
  : <InitialsAvatar name={user.name} />}
```

mykas:
After investigating more the use case I guess we don't have other choice. We will increase our perfer threshold in wix-style-react.

jonathanadler:
Hey guys,

I want to open this discussion again.
I'm really not sure this is something we need to support, and in any case I'm not sure its worth the extra bundle size.

I went over the added tests as an example for why we need to use the graphemes splitter.
I encountered these examples: `"Иван Тесто"`, `"🏳️‍🌈"`, `"अनुच्छेद"` and `"ĺo͂řȩm̅ ipsum"`.

For `"Иван Тесто"` and `"अनुच्छेद"` it's enough to just take the first character of the string (ie `str[0]`)
Regarding avatars, I'm not sure we really need to support this.

Regarding "ĺo͂řȩm̅ ipsum", is this a real language? 
I mean I know "lorem ipsum" is gibberish, but the letters used in the test, are they really something we should expect in real life? Are the correct initials really `ĹI`?

I'm asking, cause I'm not sure "accents" should be part of initials, because they're mainly for word pronunciation, and not for letters alone.
I also guess the rules for this might be different for different locales.
For instance my name in Hebrew, (with our sort of "accents"),  is `"יוֹנָתָן אַדְלֶר"`, but I don't expect my initials to be `"יאַ"`, but rather `"יא"`.

If this a real issue, maybe this should be defined better, and maybe even moved to Wix Atlas?

@shlomitc wdyt?

(relates to #1925)

Attached is an image of a bundle size analysis of a TPA, but you can see how the GraphemeSplitter size is unproportional
Screenshot from 2020-09-15 11-16-21

shlomitc:
@zemlanin - was this PR intended to solve some bug or just a nice-to-have feature?

zemlanin:
> was this PR intended to solve some bug

Screen Shot 2020-09-15 at 11 47 30

shlomitc:
@zemlanin yes I saw it, but was this a real production bug/feature? where is it used?

zemlanin:
> Regarding "ĺo͂řȩm̅ ipsum", is this a real language?

While "ĺo͂řȩm̅ ipsum" is not a real name, there are [names with `Ü`](http://magyarnevek.com/nevek/ferfinevek/Ü) for which `[0]` [may not work](https://runkit.com/embed/j1jdjflpyk6l)

> was this a real production bug/feature? where is it used?

https://github.com/wix/wix-ui-tpa/issues/145 is/was a real production issue. My rational for why we need `grapheme-splitter` [hasn't changed](https://github.com/wix/wix-ui/pull/1510#issuecomment-545813974):

> _if we want_ to have `<Avatar>`s with initials in them, we'll need to somehow compute the first visible/comprehendible symbol/grapheme (instead of the first code point)

jonathanadler:
> wix/wix-ui-tpa#145

Actually https://github.com/wix/wix-ui-tpa/issues/145 was **not** opened as a real production issue, but as a nice to have, by the developer who created the wix-ui-tpa `Avatar` component, while he was working on it, and guided by myself.

If you've encountered a production issue maybe we can find a better solution.
It would help to see the exact use-case.

BTW, in my dev console `"Üdved"[0] === "Ü"`.
And again, as I wrote before, there might be a special use case where we need to support these cases, but as I showed you, in Hebrew for instance, it's not needed.

zemlanin:
> BTW, in my dev console `"Üdved"[0] === "Ü"`

[Unicode is hard](https://mathiasbynens.be/notes/javascript-unicode#accounting-for-lookalikes)

On the other hand, we could use [`String.prototype.normalize("NFC")`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/normalize) to [account for that](https://runkit.com/embed/722zfojhomu7) 🤔

jonathanadler:
> > BTW, in my dev console `"Üdved"[0] === "Ü"`
> 
> [Unicode is hard](https://mathiasbynens.be/notes/javascript-unicode#accounting-for-lookalikes)

Very true 😞 

> On the other hand, we could use [`String.prototype.normalize("NFC")`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/normalize) _(needs polyfill in IE)_ to [account for that](https://runkit.com/embed/f9ldbivavplw) 🤔

Interesting 🤔

PR #1512, opened 2019-10-18
Prevent enter key event only if there is selection option

tuchk4:
Related to https://github.com/wix/wix-ui/pull/1508

### DIFF
--- a/packages/wix-ui-core/src/components/dropdown/Dropdown.tsx
+++ b/packages/wix-ui-core/src/components/dropdown/Dropdown.tsx
@@ -151,10 +151,14 @@ export class DropdownComponent extends React.PureComponent<
     this.setState({ isOpen: false });
   }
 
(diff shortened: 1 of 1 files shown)


packages/wix-ui-core/src/components/dropdown/Dropdown.tsx
```diff
@@ -151,10 +151,14 @@ export class DropdownComponent extends React.PureComponent<
     this.setState({ isOpen: false });
   }
 
-  onKeyboardSelect() {
-    const selectedOption = this.dropdownContentRef
+  getSelectedOption() {
+    return this.dropdownContentRef
```
tuchk4: https://github.com/wix/wix-ui/blob/master/packages/wix-ui-core/src/components/dropdown-content/DropdownContent.tsx#L114-L118

PR #1513, opened 2019-10-18
Prevent Tab key event only if there is selection option 

tuchk4:


shlomitc:

PR #1521, opened 2019-10-22
wix-storybook-utils: fix(function-to-string): guard `ensureShorthandProperties`

argshook:
ocasionally built transpilled storybook yields `Couldn't find a Program`
error. Does not happen on local environment.


argshook:
oops, wrong branch

PR #1528, opened 2019-10-27
Remove enyzme import from a non-enzyme interface

mosheKerbel:
in order to fix this issue
https://github.com/wix/wix-style-react/issues/4397
@NitayRabi @saarkuriel


giladsegal:
This might break a lot of consumers!
If you install enzyme at version 3 and some consumers use enzyme 2 - we might encounter an error
I think you should specify it as a peer dependency

PR #1533, opened 2019-11-03
wsr-types: readme - include by pattern

mosheKerbel:
I checked it, it works with a single include
@argshook @shlomitc @giladsegal

### DIFF
--- a/packages/wsr-types/readme.md
+++ b/packages/wsr-types/readme.md
@@ -23,11 +23,7 @@ To use install and update `tsconfig.json`:
     // ... compiler options
   },
   "include": [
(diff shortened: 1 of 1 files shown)


mosheKerbel:
> If this is recommended approach to import types, then maybe all of them should be in one file anyway?

Initially it was one file, but we ran into issues,
https://github.com/wix/wix-ui/pull/1527#discussion_r338796715

giladsegal:
We didn't want it to be a single file because we were afraid it would enforce consumers to import enzyme even if they use the vanilla testkit.
If we would include the `baseDrivers` only once and create a single file, and you claim it doesn't cause the user to install unnecessary libraries then we can merge

PR #1534, opened 2019-11-03
More types

giladsegal:
@mosheKerbel please verify that I didn't miss any prop - really easy to miss something
it will be easier to view those types by commit

### DIFF
--- a/packages/wsr-types/__tests__/Text-tests.tsx
+++ b/packages/wsr-types/__tests__/Text-tests.tsx
@@ -15,13 +15,22 @@ function TextWithAllProps() {
     <Text
       className="cssssss"
       size="tiny"
-      ellipsed
       light
       secondary
-      showTooltip
       skin="standard"
       tagName="marquee"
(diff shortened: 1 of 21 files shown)


packages/wsr-types/__tests__/Text-tests.tsx
```diff
@@ -15,13 +15,22 @@ function TextWithAllProps() {
     <Text
```
mosheKerbel: test children pass?

packages/wsr-types/src/components/Text.d.ts
```diff
@@ -1,8 +1,7 @@
 declare namespace __WSR {
   namespace Text {
-    export interface TextProps {
-      showTooltip?: boolean;
-      ellipsed?: boolean;
+    export interface TextProps extends __WSR.BaseComponents.EllipsisHOCProps {
```
mosheKerbel: add children prop?

PR #1535, opened 2019-11-03
Update tests

giladsegal:


packages/wsr-types/package.json
```diff
@@ -22,16 +22,17 @@
     "url": "git+https://github.com/wix/wix-ui.git"
   },
   "scripts": {
-    "test": "tsc",
+    "test": "npm run concat && tsc -p tsconfig-test.json",
+    "test:watch": "tsc --watch",
     "prerelease": "npm run test",
     "release": "npm run concat && yoshi release",
     "concat": "npm run concat:components && npm run concat:testkits",
-    "concat:components": "cat src/components/*d.ts >> types.d.ts",
```
giladsegal: @argshook I switched `>>` to `>` because running the `cat` command multiple times with `>>` will concat results to existing files, so if you run it twice in local environment it will create wrong output

giladsegal:
I'm running the tests with the concatenated files which is what our consumers will do, I think it will give us more safety

packages/wsr-types/package.json
argshook: Good catch, somehow I thought that `*` needs to be used with append pipe

PR #1538, opened 2019-11-05
wix-ui-core - Update React version

EgoziE:
Updated React version in wix-ui-core package


shlomitc:
We need to make sure that consumers of wix-UI-core and it’s wrappers are actually using it. For example, we need to make sure which react version the business manager and ADI are using

mykas:
Hey, yes we double checked that wix-ui-backoffice is using 16.4.2 so it will start breaking after this merge. Then we can fix it. In the business manager we already have 16.8.3 so we should update everywhere this part :)

mykas:
https://github.com/wix-private/business-manager/blob/master/business-manager-web/package.json#L82

PR #1540, opened 2019-11-06
Add <DropdownLayout> type definitions

giladsegal:
That's a big one :)

### DIFF
--- a/packages/wsr-types/__tests__/Input-tests.tsx
+++ b/packages/wsr-types/__tests__/Input-tests.tsx
@@ -33,6 +33,7 @@ function InputWithAllProps() {
       help
       helpMessage="msg"
       hideStatusSuffix
+      id="1"
       max={10}
       maxLength={100}
       menuArrow
--- a/packages/wsr-types/__tests__/Input-tests.tsx
+++ b/packages/wsr-types/__tests__/Input-tests.tsx
@@ -1,0 +1,108 @@
+import * as React from 'react';
+import Input from 'wix-style-react/Input';
+import { inputTestkitFactory } from 'wix-style-react/dist/testkit';
+import { inputTestkitFactory as inputEnzymeTestkitFactory } from 'wix-style-react/dist/testkit/enzyme';
+import { inputTestkitFactory as inputPuppeteerTestkitFactory } from 'wix-style-react/dist/testkit/puppeteer';
+import * as enzyme from 'enzyme';
+import * as puppeteer from 'puppeteer';
+
+function inputWithMandatoryProps() {
+  return <Input />;
+}
+
+function InputWithAllProps() {
+  return (
+    <Input
+      ariaControls="label"
+      ariaDescribedby="label"
+      ariaLabel="text"
+      autoFocus
+      autoSelect
+      autocomplete="off"
+      className="cls"
+      clearButton
+      customInput={<span />}
+      dataHook="hook"
+      defaultValue="value"
+      disableEditing
+      disabled
+      error
+      errorMessage="msg"
+      forceFocus
+      forceHover
+      help
+      helpMessage="msg"
+      hideStatusSuffix
+      max={10}
+      maxLength={100}
+      menuArrow
+      min={5}
+      name="name"
+      noLeftBorderRadius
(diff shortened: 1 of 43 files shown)


packages/wsr-types/src/drivers/TextButton.uni.driver.d.ts
```diff
@@ -1,5 +1,3 @@
 declare namespace __WSRTests {
-  type TextButtonUniDriverBase = import('wix-ui-core/drivers/unidriver').ButtonNextDriver;
```
giladsegal: I don't know why it didnt break my tests earlier - this was imported in multiple test files - I had to move it to `baseDrivers.d.ts`
BTW if we would have only 1 big file containing all the type for all test environments and production, we wouldn't have this problem because each type would be included only once.
If you say you've tested a client project without a dependency in enzyme (or some other library) and importing our type definitions which depend on enzyme doesn't break, maybe we should consider doing it

packages/wsr-types/src/components/Page.d.ts
```diff
@@ -0,0 +1,74 @@
+declare namespace __WSR {
+  namespace Page {
+    export type PagePropsNew = BaseComponents.WixComponentProps & {
```
giladsegal: @mosheKerbel  lets talk about this example at Sunday
type New = { upgrade: true; new: number, meh: boolean };
type Old = { upgrade?: false; new: string };
type Maybe = New | Old;
const a: Maybe = { meh: true };

packages/wsr-types/src/drivers/DropdownLayout.driver.d.ts
```diff
@@ -0,0 +1,60 @@
+declare namespace __WSRTests {
+  export interface DropdownLayoutDriver extends BaseDriver {
```
mosheKerbel: missing type
https://github.com/wix/wix-style-react/blob/a1cf61edda0ab5a3f1863ebb0f4a6b367e7a6356/src/DropdownLayout/DropdownLayout.js#L501

packages/wsr-types/__tests__/Input-tests.tsx
```diff
@@ -0,0 +1,108 @@
...
+import { inputTestkitFactory as inputPuppeteerTestkitFactory } from 'wix-style-react/dist/testkit/puppeteer';
+import * as enzyme from 'enzyme';
+import * as puppeteer from 'puppeteer';
+
+function inputWithMandatoryProps() {
+  return <Input />;
+}
+
+function InputWithAllProps() {
+  return (
```
mosheKerbel: missing param - `id`

packages/wsr-types/__tests__/Input-tests.tsx
```diff
@@ -0,0 +1,108 @@
...
+      suffix={<div />}
+      tabIndex={0}
+      textOverflow="clip"
+      theme="amaterial"
+      title="title"
+      tooltipPlacement="auto"
+      type="text"
+      updateControlledOnClear
+      value="value"
+      withSelection>
```
mosheKerbel: remove >

packages/wsr-types/__tests__/Modal-tests.tsx
```diff
@@ -0,0 +1,58 @@
...
+import { modalTestkitFactory as modalPuppeteerTestkitFactory } from 'wix-style-react/dist/testkit/puppeteer';
+import * as enzyme from 'enzyme';
+import * as puppeteer from 'puppeteer';
+
+function ModalWithMandatoryProps() {
+  return <Modal isOpen />;
+}
+
+function ModalWithAllProps() {
+  return (
```
mosheKerbel: missing `onOk`

packages/wsr-types/src/components/Modal.d.ts
```diff
@@ -0,0 +1,47 @@
+declare namespace __WSR {
+  namespace Modal {
+    export interface ModalProps extends BaseComponents.WixComponentProps {
```
mosheKerbel: missing `onOk: PropTypes.func,`

PR #1542, opened 2019-11-06
Created `<ClickOutside/>` component

EgoziE:
**Behavioural component `<ClickOutside/>`**
Triggers a given callback when clicking outside `<ClickOutside/>`'s children
```
<ClickOutside> ~children~ </ClickOutside>
```
Created a story page

Also added `disableOnClickOutside` prop in order to remove listeners when there's no need for them.
Fixing https://github.com/wix/wix-style-react/issues/4468

### DIFF
--- a/packages/wix-ui-core/src/components/click-outside/ClickOutside.tsx
+++ b/packages/wix-ui-core/src/components/click-outside/ClickOutside.tsx
@@ -17,7 +17,7 @@ export interface ClickOutsideProps {
 /**
  * Click outside behavior
  */
-class ClickOutside extends React.PureComponent<ClickOutsideProps> {
+export class ClickOutside extends React.PureComponent<ClickOutsideProps> {
   private readonly _boundEvents: string[];
 
   constructor(props) {
@@ -100,5 +100,3 @@ class ClickOutside extends React.PureComponent<ClickOutsideProps> {
     return this.props.children;
   }
 }
-
-export default ClickOutside;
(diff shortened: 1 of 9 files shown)


packages/wix-ui-core/src/components/click-outside/ClickOutside.tsx
```diff
@@ -0,0 +1,102 @@
...
+      document.removeEventListener(eventName, this._onClickOutside, true);
+    }
+  }
+
+  componentDidMount() {
+    if (!this.props.disableOnClickOutside) {
+      this._registerEvents();
+    }
+  }
+  componentDidUpdate(prevProps) {
```
mykas: This should be done with `getDerivedStateFromProps` instead

packages/wix-ui-core/src/components/click-outside/ClickOutside.tsx
```diff
@@ -0,0 +1,102 @@
...
+    const { rootRef, excludeClass } = this.props;
+    let target: HTMLElement = event.target;
+    while (target) {
+      if (
+        rootRef.current === target ||
+        target.classList.contains(excludeClass)
+      ) {
+        return false;
+      }
+      target = target.parentElement;
```
mykas: So I am not sure if this would really work because popover content could attached on a child near the parent node element. And this happens with popover as it attaches to document.body
Screenshot 2019-11-06 at 15 20 30

mykas:
I wondering if we should include excluding scrollbar click: https://github.com/Pomax/react-onclickoutside/blob/master/src/index.js#L42

packages/wix-ui-core/src/components/click-outside/ClickOutside.tsx
mykas: -- document.body
   --- div
      --- div
          --- div
   --- popover portal

mykas: So the solution should be:
1. Search through parent elements
2. Search parent elements direct children but don't go deep.

PR #1548, opened 2019-11-10
Switch to Node v12 LTS

simast:
- Old `yoshi` versions (`1.x` and `2.x`) no longer build with Node  v12 (due to native `node-sass` dependency). Had to update yoshi in a couple of packages.
- Due to `yoshi` update having a new webpack version - had to update Storybook in `wix-ui-icons-common` package.


simast:
Split up changes into separate PRs for each package.

PR #1544, opened 2019-11-07
Migrate <Popover/> to the new <ClickOutside/>

EgoziE:
Migrate `<Popover/>` to the new `<ClickOutside/>`

PR #1551, opened 2019-11-10
Revert "Migrate <Popover/> to the new <ClickOutside/>"

mykas:
Reverts wix/wix-ui#1544


OriMarron:
@mykas why did you revert? will you migrate to new version again?

mykas:
It caused our users builds to fail on Dropdown changes. The popover side will be extracted to a separate pr and for dropdown we will need to do a breaking change at some time in the future.

OriMarron:
haha funny 😀
we broke after the initial PR, made a fix, and now broke again for the revert.
but no problem, I reverted as well.

mykas:
sorry :D

PR #1552, opened 2019-11-10
Add <CloseButton> type definitions

giladsegal:


packages/wsr-types/src/drivers/Text.driver.d.ts
```diff
@@ -8,9 +8,4 @@ declare namespace __WSRTests {
     isLight: () => boolean;
     isSecondary: () => boolean;
   }
-
-  interface TextPuppeteerDriver {
```
giladsegal: This interface exists in `Text.puppeteer.driver.d.ts`

PR #1557, opened 2019-11-11
Update Node to v12 LTS in wix-ui-jss package

simast:


argshook:
@simast FYI, this package was created before stylable existed, while we were reasearching how to do theming for UI libraries and considered a js-in-css approach

we should remove this package but i see that it's used at least here https://github.com/wix-private/wix-ui-santa/blob/master/src/legacy/components/CoreToggleSwitch/index.js#L1

this is anyway job for another PR

PR #1565, opened 2019-11-11
Infra: unify stylable@1 versions

AviVahl:
- the older `@stylable/dom-test-kit` is causing *multiple* stylable versions for each `wix-ui-core`/`wix-style-react` user.
- the `2.2.8-alpha.0` is an abandoned effort. reverting to currently compatible v1.
- `@stylable/uni-driver` only has v2, but is compatible with both v1 and v2 of stylable, and has no dependencies (so doesn't bring along v2).

~~- tests passed locally, except for the re-captcha test, which fails for master as well.~~

### DIFF
--- a/packages/wix-ui-core/package.json
+++ b/packages/wix-ui-core/package.json
@@ -73,7 +73,7 @@
     "update-components-standalone": "bash .wuf/update-components-standalone.sh"
   },
   "dependencies": {
-    "@stylable/dom-test-kit": "^0.1.0",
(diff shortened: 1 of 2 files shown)


AviVahl:
hey @ranyitz this is improve installation times many Wix projects which use this library.

ranyitz:
@AviVahl That's cool. I don't know/remember why those versions were configure this way at the first place. Maybe @NitayRabi could help here.

AviVahl:
rebased to master and tests now pass.

PR #1574, opened 2019-11-13
Add <InputWithOptions> type definitions

giladsegal:


packages/wsr-types/src/components/InputWithOptions.d.ts
```diff
@@ -0,0 +1,39 @@
...
+    export interface InputWithOptionsProps extends Omit<Input.InputProps, 'theme'>, Omit<DropdownLayout.DropdownLayoutProps, 'theme'> {
+      // TODO: there is a bug in WSR - theme exists in InputProps and DropdownLayoutProps
+      // and it has different set of values
+      theme?: string;
+
+      autocomplete?: string,
+      inputElement?: React.ReactElement,
+      closeOnSelect?: boolean,
+      // TODO: should it be DropdownLayout.DropdownLayoutOption or DropdownLayout.DropdownLayoutValueOption
+      onManuallyInput?: (inputValue: string, suggestedOption: DropdownLayout.DropdownLayoutValueOption) => void,
```
giladsegal: should it be DropdownLayout.DropdownLayoutOption or DropdownLayout.DropdownLayoutValueOption for both functions?

packages/wsr-types/src/components/InputWithOptions.d.ts
```diff
@@ -0,0 +1,39 @@
...
+      dropdownOffsetLeft?: string,
+      showOptionsIfEmptyInput?: boolean,
+      highlight?: boolean,
+      native?: boolean,
+      popoverProps?: object; // TODO update when PopoverProps are implemented
+    }
+
+    export class InputWithOptions extends React.Component<InputWithOptionsProps> {
+      // TODO: should we include all other methods as well?
+      focus: (options: FocusOptions) => void;
```
giladsegal: what about other instance methods that are not prefixed with `_` like `renderInput`?
I think they are internal but exposed to storybook because they are not prefixed accordingly

packages/wsr-types/src/drivers/InputWithOptions.uni.driver.d.ts
```diff
@@ -0,0 +1,22 @@
...
+      focus: () => Promise<void>;
+      blur: () => Promise<void>;
+      pressKey: (key: string) => Promise<void>;
+      outsideClick: () => Promise<void>;
+      isOptionWrappedToHighlighter: (
+        optionId: string | number
+      ) => Promise<boolean>;
+    };
+    inputDriver: InputUniDriver;
+    dropdownLayoutDriver: DropdownLayoutUniDriver; // TODO: is this right?
```
giladsegal: I think this is the API although the implementation is kinda funny

packages/wsr-types/src/components/InputWithOptions.d.ts
```diff
@@ -0,0 +1,47 @@
...
+      extends Omit<Input.InputProps, "theme">,
+        Omit<DropdownLayout.DropdownLayoutProps, "theme"> {
+      // TODO: there is a bug in WSR - theme exists in InputProps and DropdownLayoutProps
+      // and it has different set of values
+      theme?: string;
+
+      autocomplete?: string;
+      inputElement?: React.ReactElement;
+      closeOnSelect?: boolean;
+      onManuallyInput?: (
```
giladsegal: should it be DropdownLayout.DropdownLayoutOption or DropdownLayout.DropdownLayoutValueOption

packages/wsr-types/src/components/Search.d.ts
```diff
@@ -0,0 +1,22 @@
+declare namespace __WSR {
+  namespace Search {
+    export interface SearchProps
+      extends InputWithOptions.InputWithOptionsProps {
+      expandable?: boolean;
+      predicate?: (option: DropdownLayout.DropdownLayoutValueOption) => boolean;
```
giladsegal: We need to think if it's DropdownLayout.DropdownlayoutValueOption or just option

packages/wsr-types/src/components/InputWithOptions.d.ts
mosheKerbel: i think it is safe enough to use the actual option with value `DropdownLayout.DropdownLayoutValueOption`

packages/wsr-types/src/components/Search.d.ts
mosheKerbel: `DropdownLayout.DropdownLayoutValueOption` is safe enough IMO

packages/wsr-types/src/components/CounterBadge.d.ts
```diff
@@ -0,0 +1,32 @@
+declare namespace __WSR {
+  namespace CounterBadge {
+    export interface CounterBadgeProps
+      extends React.HTMLAttributes<HTMLElement> {
```
mosheKerbel: hmm why we need that one?

giladsegal: all props are spread into `stylable-badge/Badge.tsx` into a span in wix-ui-core
so you can pass regular HTML props, am I wrong?

packages/wsr-types/src/drivers/InputWithOptions.uni.driver.d.ts
mosheKerbel: yes this is it, the proxy looks weird, but i don't think that it break the original one, same interface should work.

packages/wsr-types/src/components/InputWithOptions.d.ts
mosheKerbel: I tried to run a search on some of these props inside `wix-private` org in github and got no results. 
I agree it's probably a mistake there are exposed inside storybook because of the missing "_" prefix.
Anyways, i think we better of not adding them to the types. In a case we break someone we then add specific ones or maybe all. Agree?

giladsegal: 👍

packages/wsr-types/src/components/CounterBadge.d.ts
mosheKerbel: it is spreading the `props` in stylable function not on the actual span, 
so i don't think it necessary.
```jsx
export const Badge: React.FunctionComponent<BadgeProps> = props => (
  <span {...style('root', {}, props)}>{props.children}</span>
);
```

PR #1576, opened 2019-11-14
Add <CardGalleryItem /> typings

simast:


packages/wsr-types/src/components/CardGalleryItem.d.ts
```diff
@@ -0,0 +1,35 @@
+declare namespace __WSR {
+  namespace CardGalleryItem {
+    export interface CardGalleryItemProps {
+      badge?: React.ReactNode;
+      title?: React.ReactNode;
+      subtitle?: React.ReactNode;
+      backgroundImageUrl?: string;
+      backgroundImageNode?: React.ReactNode;
+      primaryActionProps?: {
```
giladsegal: This prop is not optional

packages/wsr-types/src/components/CardGalleryItem.d.ts
```diff
@@ -0,0 +1,35 @@
...
+        label?: React.ReactNode;
+        onClick?: React.MouseEventHandler<HTMLElement>;
+        disabled?: boolean;
+        disabledMessage?: string;
+      };
+      secondaryActionProps?: {
```
giladsegal: This prop is not optional as well

packages/wsr-types/src/components/CardGalleryItem.d.ts
```diff
@@ -0,0 +1,35 @@
...
+        label?: React.ReactNode;
+        onClick?: React.MouseEventHandler<HTMLElement>;
```
giladsegal: I think this should be `HTMLButtonElement`

packages/wsr-types/src/components/CardGalleryItem.d.ts
simast: technically it is! There is a `defaultProps` on the component - so React will actually not complain if you do not provide this prop (but it's definitely a bug in the component).

Anyway - what I did is marked `primaryActionProps` as required but left `secondaryActionProps` as optional. As there is some code in the wild that already expect `secondaryActionProps` to be optional, for example:

https://github.com/wix-private/form-builder-resources/blob/10890e46628fc0fb2a8a4a45b5592808f1abfb11/src/components/PresetsPanel/PresetsPanel.tsx#L110
https://github.com/wix-private/promote/blob/5c28e5121780b68814106cb18fb3d16919ac16d1/email-marketing-backoffice/src/home/home.tsx#L16

PR #1579, opened 2019-11-15
InputWithOptions: a11y

zemlanin:
See also: part 6 of https://uxmastery.com/anatomy-of-an-accessible-auto-suggest/

### DIFF
--- a/packages/wix-ui-core/src/components/dropdown-content/DropdownContent.tsx
+++ b/packages/wix-ui-core/src/components/dropdown-content/DropdownContent.tsx
@@ -4,11 +4,17 @@ import { Option, DropdownOption } from '../dropdown-option';
 
 const NOT_HOVERED_INDEX = -1;
 
+export interface IDOMid {
+  _DOMid: string;
+}
+
 export interface DropdownContentProps {
   /** Component class name */
   className?: string;
   /** The dropdown options array */
   options: Option[];
+  /** A callback for when hovering an option */
+  onOptionHover?(option: (Option & IDOMid) | null): void;
   /** A callback for when clicking an option */
   onOptionClick(option: Option | null): void;
   /** A callback for mouse down event */
@@ -19,6 +25,8 @@ export interface DropdownContentProps {
   fixedHeader?: React.ReactNode;
   /** An element that always appears at the bottom of the options */
   fixedFooter?: React.ReactNode;
+  /** DOM id of optionsContainer element */
+  optionsContainerId?: string;
 }
 
 export interface DropdownContentState {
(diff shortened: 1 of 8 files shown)


packages/wix-ui-core/src/components/dropdown-content/DropdownContent.tsx
```diff
@@ -112,11 +130,24 @@ export class DropdownContent extends React.PureComponent<
...
     const { options } = this.props;
     const { hoveredIndex } = this.state;
     const isValidIndex = hoveredIndex >= 0 && hoveredIndex < options.length;
-    return isValidIndex ? options[hoveredIndex] : null;
+    return isValidIndex
+      ? {
+          ...options[hoveredIndex],
+          _DOMid: this.getOptionDOMid(options[hoveredIndex]),
+        }
+      : null;
```
Oleksiika: It's not clear to me - the method has name `onKeyboardSelect` which seems like it's an event handler, but it returns `(Option & IDOMid) | null` and inside it checks if hoveringIndex is in valid options range. Also in line 70 it's result is being used as an argument to `onOptionHover` prop. What's happening here? 🤔

zemlanin: You're right. Renamed `DropdownContent#onKeyboardSelect` to more accurate `DropdownContent#getSelectedOption` in https://github.com/wix/wix-ui/pull/1579/commits/def7e45f94868a71dd5bcfaed5a2b984bcfff2a4

_Just to be clear: before this PR, method was already called `onKeyboardSelect` and already was returning `Option | null`_

packages/wix-ui-core/src/components/dropdown/Dropdown.tsx
```diff
@@ -131,11 +138,20 @@ export class DropdownComponent extends React.PureComponent<
     this.close();
   }
 
+  _onExpandedChange() {
```
zonzujiro: Underscore for consistency? You can use `private`

packages/wix-ui-core/src/components/dropdown-content/DropdownContent.tsx
```diff
@@ -4,11 +4,17 @@ import { Option, DropdownOption } from '../dropdown-option';
 
 const NOT_HOVERED_INDEX = -1;
 
+export interface IDOMid {
+  _DOMid: string;
+}
```
NitayRabi: Do we expect all consumers to add `_DOMid` for each option passed to Dropdown? maybe `onHover` pass the index for option identification?

zemlanin: `_DOMid` value is [added by `DropdownContent`](https://github.com/wix/wix-ui/pull/1579/files#diff-bc057c87b9515f41998a9cecf830e35eR148) and not by consumers of `Dropdown`

`_DOMid` is here to be used as value for [`aria-activedescendant`](https://www.w3.org/TR/wai-aria/#aria-activedescendant) attribute. If `onHover` will receive the option's index instead of DOM `id`, then, to keep `aria-` attribute valid, event consumer will need to reconstruct DOM `id` in the same format as [`DropdownContent#getOptionDOMid`](https://github.com/wix/wix-ui/pull/1579/files#diff-bc057c87b9515f41998a9cecf830e35eR133), making components too tightly coupled

packages/wix-ui-core/src/components/dropdown/Dropdown.tsx
zemlanin: > Underscore for consistency?

Yep

zemlanin:
@NitayRabi added tests for `aria-` attributes and `onOptionHover` in https://github.com/wix/wix-ui/pull/1579/commits/517f9c4ece8b693195d110b2e15ba662e6503f3a

PR #1587, opened 2019-11-19
wix-storybook-utils: Implemented plugin section

vytenisu:
Plugin sections allows consumer of storybook utils to inject their own section while still getting full configuration context.

### DIFF
--- a/packages/wix-storybook-utils/docs/sections.md
+++ b/packages/wix-storybook-utils/docs/sections.md
@@ -15,6 +15,7 @@ import {
   importexample,
   mdx,
   playground,
+  plugin,
   tab,
   table,
   tabs,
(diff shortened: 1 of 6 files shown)


packages/wix-storybook-utils/docs/sections.md
```diff
@@ -210,6 +211,15 @@ accepts no parameters.
 
 ---
 
+## `plugin()` section
```
argshook: this file is generated with `npm run docs:sections`, so if you add content here it might be lost.

All comments and descriptions should go to `typings/story-section.ts`

vytenisu: Got it. Can I suggest .gitignor'ing this file then? I think this file was in fact generated when I launched build locally.

argshook: Perhaps file name should be more explanatory. Otherwise gitignoring is not good, I want this file to be visible in git. Currently there's no dedicated documentation website for wix-storybook-utils

PR #1597, opened 2019-11-20
<PreviewWidget/>, <MobilePreviewWidget/>, <BrowserPreviewWidget/>- adding types to testkits

TalZa1810:
Adding types to `<PreviewWidget/>`, `<MobilePreviewWidget/>` and `<BrowserPreviewWidget/>`.

### DIFF
--- a/packages/wsr-types/src/testkits/enzyme.d.ts
+++ b/packages/wsr-types/src/testkits/enzyme.d.ts
@@ -203,4 +203,7 @@ declare module 'wix-style-react/dist/testkit/enzyme' {
   export const cardSubheaderTestkitFactory: any;
   export const tooltipTestkitFactory: any;
   export const TooltipTestkit: any;
+  export const previewWidgetTestkit: any;
+  export const mobilePreviewWidgetTestkit: any;
+  export const browserPreviewWidgetTestkit: any;
 }


shlomitc:
If it’s a new component - why not to create the types next to the component?

TalZa1810:
I've added the types next to the Components. We talked about this today in the meeting. Moshe asked us to add these as a temporary solution for all new components until we merge all types to WSR.

shlomitc:
we're adding here types for existing components because we don't want to break users, but I think that for new components, types should be in `wsr` from day one (that's why we want to add it to the generator)

mosheKerbel:
> we're adding here types for existing components because we don't want to break users, but I think that for new components, types should be in `wsr` from day one (that's why we want to add it to the generator)

@shlomitc Tal did added the types next to the component
this is just an addition for `wsr-types` so our testers would still be able to import the teskits, because the testkits d.ts over-ride the ones in wix-style-react.

shlomitc:
so it looks like an issue needs to be solved.
Can't we declare this module as a union for `any` with the testkits that are here?

we can merge it anyway if that's how it works now

mosheKerbel:
> so it looks like an issue needs to be solved.
> Can't we declare this module as a union for `any` with the testkits that are here?
> 
> we can merge it anyway if that's how it works now

no, once you declare this kind of type
`declare module 'wix-style-react/dist/testkit/enzyme'`
and import it from a project, it completly over-ride the old d.ts, it does not make a union of the two.

mosheKerbel:
@TalZa1810 plz add it also for the other testkits

PR #1614, opened 2019-11-26
wix-ui-framework: chore(pom.xml): remove Marius from developers

argshook:
@mariusju fyi, im happy to not remove you if you wish, just thought maybe you don't need that spam :)


mariusju:
@argshook thanks, i was too lazy to do it myself

PR #1618, opened 2019-11-26
Use Omit polyfill and loosen Input types

giladsegal:


packages/wsr-types/src/components/Input.d.ts
```diff
@@ -35,17 +35,17 @@ declare namespace __WSR {
       name?: string;
       noLeftBorderRadius?: boolean;
       noRightBorderRadius?: boolean;
-      onBlur?: React.FocusEventHandler<HTMLElement>;
-      onChange?: React.ChangeEventHandler<HTMLElement>;
-      onClear?: React.MouseEventHandler<HTMLElement>;
+      onBlur?: React.FocusEventHandler<HTMLInputElement>;
```
mosheKerbel: was it also necessary to change it?
`HTMLElement ` -> `HTMLInputElement`

PR #1619, opened 2019-11-27
<Popover/> - Remove ClickOutside listeners when popover is not shown

EgoziE:
Added an upgrade prop to remove ClickOutside listeners when popover is not shown
+ added 2 tests


packages/wix-ui-core/src/components/popover/Popover.tsx
```diff
@@ -121,6 +121,11 @@ export interface PopoverProps {
    * - `string` value that contains `px`
    */
   width?: number | string;
+
+  /**
+   * Fix removing ClickOutside listeners
```
mykas: It would be better to explain what it does. The problem our users have that our clickoutside is firing events when popover content is closed. So lets write: `when true - onClickOutside will be called only when popover content is shown`

PR #1628, opened 2019-11-28
<MarketingLayout /> testkit exports added

mantaspa:


argshook:
@mosheKerbel fyi, this is for new `MediaLayout` component. Please see

PR #1647, opened 2019-12-07
<TextButton/> - change tiny size height to 15px

EgoziE:
Changed `<TextButton/>` height to 15px when size `tiny`
According to the design, the icons stays 18px with negative margins

Pay attention that visuals will break!

### DIFF
--- a/packages/wix-ui-core/src/themes/backoffice/text-button/text-button.st.css
+++ b/packages/wix-ui-core/src/themes/backoffice/text-button/text-button.st.css
@@ -383,8 +383,8 @@
   -st-mixin: sizeSmall;
 }
 
-.textButton.tiny{
-  height: 18px;
+.textButton.tiny {
+  height: 15px;
   -st-mixin: sizeTiny;
 }
 
@@ -417,9 +417,11 @@
 .textButton.tiny .prefix {
   width: 18px;
   height: 18px;
+  margin: -2px 0 -1px 0;
 }
 
 .textButton.tiny .suffix {
   width: 18px;
   height: 18px;
+  margin: -2px 0 -1px 0;
 }


packages/wix-ui-core/src/themes/backoffice/text-button/text-button.st.css
```diff
@@ -417,9 +417,11 @@
 .textButton.tiny .prefix {
   width: 18px;
   height: 18px;
+  margin: -2px 0 -1px 0;
 }
 
 .textButton.tiny .suffix {
   width: 18px;
   height: 18px;
+  margin: -2px 0 -1px 0;
```
milkyfruit: @EgoziE code looks based on [specs](https://zpl.io/a7Oj8wW). However can't check how it visually looks, couldn't access PR's storybook, does it have it?

@wuwa this change is to fix `ListItemSection` [baseline issue](https://github.com/wix/wix-style-react/issues/4669). Maybe you remember why tiny `TextButton` was developed 18px?

shlomitc: @wuwa @milkyfruit ?

milkyfruit: @shlomitc I suggest to merge it and release it together with breaking-change release (in case this causes layout break for someone).

shlomitc: it's not how it works. if we merge it - it will be immediately released.
The question here - is this going to make any harm to the products UX or only a slight change in how things look

milkyfruit: not critical, but might distort layout in some places.

shlomitc: @EgoziE - I think it's safe to merge

PR #1651, opened 2019-12-09
`<LinearProgressBar/>` - adding classname to progress percentage span

TalZa1810:


packages/wix-ui-core/src/components/linear-progress-bar/LinearProgressBar.st.css
```diff
@@ -30,3 +35,12 @@
...
 .indicationContainer {}
+
+.root .progressPercentage {
+  color: value(D10);
+}
+
+.root:light .progressPercentage {
+  color: value(D80);
+
+}
```
nitayneeman: We're doing efforts to ditch the styles from ui-core so that eventually we could remove the backoffice theme completely - maybe we should add them to WSR instead and avoid adding new styles to ui-core?

TalZa1810: backoffice theme? I'm not sure I understand. 
Currently, the progress percentage is rendered in `wix-ui-core`.  Maybe trying to extend the styling in `wix-style-react` and adding those values there? is that what you were referring to?

TalZa1810: In order to do that, I need to add the class-name here and add the colors in `wix-style-react` if I understand correctly.

nitayneeman: Yes. I meant that these changes use styles from the backoffice theme (colors actually), which will be removed at some point - and in general, I think our way is to make ui-core style-less as much as possible.

TalZa1810: ok sababa so I just added classname to the span and I'll implement the colors in `wix-style-react`. Can I merge this?

nitayneeman: Not sure if it's the most proper way to do that (`data-attr` maybe better?), but I see this is the convention here so let's proceed this way and merge.

PR #1652, opened 2019-12-09
react-autodocs-parser: metadata parser should work with tree shakable imports

mykas:
I am adding a skipped unit tests that shows that situation.


shlomitc:
@mykas - is this still relevant?

mykas:
Yes, we might want to migrate our core imports in wix-style-react to import with named imports instead of going through dist/src. 

Now users need to specifically know that they need to import from dist/src otherwise our transpilation code will throw an error

PR #1677, opened 2019-12-18
fix loop for youtube player

yteslenk:


packages/wix-ui-core/src/components/video/players/YouTube.tsx
```diff
@@ -121,7 +123,6 @@ class YouTubePlayer extends React.PureComponent<IYouTubeProps> {
         autoplay: playing ? 1 : 0,
         mute: muted ? 1 : 0,
         controls: controls ? 1 : 0,
-        loop: loop ? 1 : 0,
         origin: window.location.origin,
```
mokhovyk: is it work without loop property?

packages/wix-ui-core/src/components/video/players/YouTube.tsx
```diff
@@ -47,7 +47,9 @@ const mapPropsToPlayer: IPropsToPlayer = {
     }
   },
   volume: 'setVolume',
-  loop: 'setLoop',
+  loop: instance => {
```
mokhovyk: pls, left a comment why we are using reload instead of setLoop here

packages/wix-ui-core/src/components/video/players/YouTube.tsx
yteslenk: yes

packages/wix-ui-core/src/components/video/players/YouTube.tsx
yteslenk: done

PR #1678, opened 2019-12-19
wix-ui-icons-common: Publish SVG icons

simast:
Raw SVG icons are now published together with React icon components. This change was implemented so the icons can be used within CSS files directly.

```js
import Add from 'wix-ui-icons-common/Add'; // React component
import Add from 'wix-ui-icons-common/Add.svg'; // SVG document
```

```css
background-image: url('~wix-ui-icons-common/Add.svg');
```


simast:
@argshook I have implemented svg optimisation by referencing non-public `svg2react-icon/src/lib/svg-optimizer` import. Not happy about this, we could add `svgo` dependency and run optimizer directly. WDYT?

argshook:
If this matter is not urgent, then I'd suggest contribute to svg2react-icon and make svgo config public or even better expose some command that does optimization while hiding svgo details.

But if this is more urgent, then referencing file is good enough

PR #1683, opened 2019-12-22
More types 18

giladsegal:


packages/wsr-types/src/components/Sidebar.d.ts
```diff
@@ -0,0 +1,88 @@
...
+    interface SidebarPersistentFooterProps {}
+
+    class SidebarBackButton extends React.Component<SidebarBackButtonProps> {}
+    interface SidebarBackButtonProps {
+      disable?: boolean;
+      onClick?: React.MouseEventHandler<HTMLElement>;
+    }
+
+    export class SidebarContextConsumer extends React.Component<
+      SideBarContextConsumerProps
```
nitayneeman: Maybe we should rename just to be aligned?

```suggestion
      SidebarContextConsumerProps
```

packages/wsr-types/src/components/Sidebar.d.ts
```diff
@@ -0,0 +1,88 @@
...
+    > {}
+    interface SideBarContextConsumerProps {
```
nitayneeman: Maybe we should rename just to be aligned?

```suggestion
    interface SidebarContextConsumerProps {
```

PR #1685, opened 2019-12-29
Fix Notification.ActionButton link prop type

splintor:
According to https://github.com/wix/wix-style-react/blob/master/src/Notification/README.md, the `link` prop type is string, as it is a "url to navigate to on click".


splintor:
Already fixed in https://github.com/wix/wix-style-react/pull/5082

PR #1686, opened 2019-12-29
<ClickOutside/> - Add useCapture/options prop

EgoziE:
Add the native parameter useCapture/options
For more information, take a look at https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener


nitayneeman:
What's the real use-case? Why do we need to expose it?

PR #1695, opened 2020-01-06
[VOD-5044] Update playable in wix-ui-core

yuriibut:


packages/wix-ui-core/package.json
```diff
@@ -138,7 +138,7 @@
     "react-testing-library": "^6.0.3",
     "slash": "^2.0.0",
     "ts-jest": "^22.0.3",
-    "typescript": "~3.0.3",
+    "typescript": "^3.7.4",
```
piratus: I'd suggest to keep the tilde
```suggestion
    "typescript": "~3.7.4",
```

PR #1696, opened 2020-01-06
Add font definitions

EgoziE:
Added font definition through css variables
IE11 compatible

The name `--wsr-font` is meant to be as unique as possible because it's global


EgoziE:
Not relevant atm

PR #1698, opened 2020-01-06
Update <Stepper> component types

simast:
Updated for latest changes introduced with https://github.com/wix/wix-style-react/pull/4786

### DIFF
--- a/packages/wsr-types/src/drivers/Stepper.uni.driver.d.ts
+++ b/packages/wsr-types/src/drivers/Stepper.uni.driver.d.ts
@@ -1,9 +1,12 @@
 declare namespace __WSRTests {
   interface StepperUniDriver extends BaseUniDriver {
-    clickStep: (idx: number) => Promise<void>;
-    hoverStep: (idx: number) => Promise<void>;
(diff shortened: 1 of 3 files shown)


packages/wsr-types/src/drivers/Stepper.uni.driver.d.ts
```diff
@@ -1,9 +1,12 @@
...
-    hoverStep: (idx: number) => Promise<void>;
+    getType: () => Promise<string>;
+    getFit: () => Promise<string>;
+    clickStep: (index: number) => Promise<void>;
+    hoverStep: (index: number) => Promise<void>;
     getNumberOfSteps: () => Promise<number>;
-    isStepActive: (id: number) => Promise<string | null>;
-    getStepType: (idx: number) => Promise<string | null>;
+    isStepActive: (index: number) => Promise<boolean>;
+    getStepType: (index: number) => Promise<string>;
```
giladsegal: Technically unidriver `attr` returns `string | null` if you state that you return a string you can return `__WSR.Stepper.StepperType`

PR #1710, opened 2020-01-10
[VOD-5043] Add support for FB portrait videos to wix-ui-core/Video

yuriibut:


packages/wix-ui-core/src/components/video/Video.st.css
```diff
@@ -3,7 +3,8 @@
 }
 
 .playerContainer {
-    position: relative;
+    position: relative !important;
+    display: block !important;
```
mokhovyk: why are you using !important here?

PR #1735, opened 2020-01-26
wix-ui-codemode - introduce named-imports codemode

mykas:


packages/wix-ui-codemod/src/wix-style-react/named-imports.ts
```diff
@@ -0,0 +1,104 @@
...
+      return;
+    }
+
+    const cleanImport = name => {
+      path.value.specifiers = path.value.specifiers.filter(
+        spec => spec.local.name !== name,
+      );
+    };
+
+    path.value.specifiers.forEach((specifier, index) => {
```
simast: we have a lot of `path.value.specifiers` - maybe we can alias to a separate const var?

mykas: Left it as there is  reasign happening here

PR #1744, opened 2020-01-30
(wix-storybook-utils) - Expose transparent background prop for code example

mosheKerbel:
@argshook 
Motivation is to allow also non-compact examples to be displayed with a transparent background, for `wix-base-ui` per @Zohar-Telor request


packages/wix-storybook-utils/src/LiveCodeExample/LiveCodeExample.js
```diff
@@ -36,6 +36,7 @@ export default class LiveCodeExample extends Component {
     previewProps: PropTypes.object,
     autoRender: PropTypes.bool,
     darkBackground: PropTypes.bool,
+    transparentBackground: PropTypes.bool,
```
argshook: i'd call it `noBackground` just shorter to type :)

mosheKerbel: 👍

PR #1752, opened 2020-02-06
<Avatar/> - Add onClick event prop

simast:


packages/wix-ui-core/src/components/avatar/avatar.tsx
```diff
@@ -126,23 +128,16 @@ export class Avatar extends React.Component<AvatarProps, AvatarState> {
...
+    const { name, title, ariaLabel, onClick } = this.props;
     const contentType = this.getCurrentContentType();
+
     return (
       <div
         data-content-type={contentType} // for testing
         data-img-loaded={this.state.imgLoaded} // for testing
         title={title || name}
         aria-label={ariaLabel || name}
+        onClick={onClick}
```
shlomitc: Please notice that clickable elements should support a11y features (like rounded border, keyboard navigation).
we do it using the focusableHOC like in `button-next`

simast:
Added focusableHOC behaviour.
@shlomitc @argshook Please take another look at this PR.

packages/wix-ui-core/src/components/avatar/avatar.tsx
```diff
@@ -125,24 +136,39 @@ export class Avatar extends React.Component<AvatarProps, AvatarState> {
     delete this.img;
   };
 
+  handleKeyDown: React.KeyboardEventHandler<HTMLElement> = event => {
```
argshook:
```suggestion
  _handleKeyDown: React.KeyboardEventHandler<HTMLElement> = event => {
```

use underscore for name, otherwise autodocs will show this method as public api

PR #1755, opened 2020-02-10
`<Slider/>`- adding drivers functions types and prop type

TalZa1810:
Related PR:
https://github.com/wix/wix-style-react/pull/4962/

### DIFF
--- a/packages/wsr-types/src/components/Slider.d.ts
+++ b/packages/wsr-types/src/components/Slider.d.ts
@@ -15,7 +15,7 @@ declare namespace __WSR {
       pushable?: boolean | number;
       value?: number[] | number;
       disabled?: boolean;
-      marks?: {[key:number]: number} | {[key:number]: string};
+      marks?: {[key:number]: number | string};
     }
 
     export class Slider extends React.PureComponent<SliderProps> {}
--- a/packages/wsr-types/src/components/Slider.d.ts
+++ b/packages/wsr-types/src/components/Slider.d.ts
@@ -15,6 +15,7 @@ declare namespace __WSR {
       pushable?: boolean | number;
       value?: number[] | number;
       disabled?: boolean;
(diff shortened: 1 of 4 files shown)


packages/wsr-types/src/components/Slider.d.ts
```diff
@@ -15,6 +15,7 @@ declare namespace __WSR {
       pushable?: boolean | number;
       value?: number[] | number;
       disabled?: boolean;
+      marks?: object;
```
giladsegal: `marks` is not in the `Slider.PropTypes` so you should either delete it from here or add it there as well
Anyway if you decide to add it, it shouldn't be object (it doesn't supply any autocomplete to users)
You can find the actual type here: https://github.com/schrodinger/rc-slider

TalZa1810: It's part of a PR that I want to merge.
It is in the prop types here:
https://github.com/wix/wix-style-react/blob/e04e45c0bdf1ce0d0918ca08c30b00616d8e1ca9/src/Slider/Slider.js#L163

It's part of a new feature (custom marks) that I'm implementing in the wsr slider. 
I've added tests for this as well.
Check out this PR:
wix/wix-style-react#4962

TalZa1810:
Hi @giladsegal ,
Can you please have another look before merging? :)

Thanks!

packages/wsr-types/src/components/Slider.d.ts
```diff
@@ -15,6 +15,7 @@ declare namespace __WSR {
       pushable?: boolean | number;
       value?: number[] | number;
       disabled?: boolean;
+      marks?: {[key:number]: number} | {[key:number]: string};
```
giladsegal: I think you can do:
```ts
marks?: {[key:number]: number|string};
```

PR #1757, opened 2020-02-10
<Avatar/> - Add data-hook prop

simast:
Should it be `dataHook` or `data-hook` in core? Other components suggest it should be `data-hook`.


argshook:
> Should it be `dataHook` or `data-hook` in core? Other components suggest it should be `data-hook`.

let's use `data-hook`

PR #1774, opened 2020-02-11
Add contactItemBuilder types

giladsegal:


packages/wsr-types/src/components/ContactItemBuilder.d.ts
```diff
@@ -0,0 +1,26 @@
...
+    export type ContactItemBuilderFn = (data: {
+      id: string | number,
+      title: string,
+      subtitle?: string,
+      imageUrl?: string,
+      disabled?: boolean,
+    }) => {
+      id: string | number,
+      disabled: boolean | undefined,
+      value:  (data: Partial<{ selected: boolean }>) => React.ReactNode
```
mosheKerbel: hmm can u explain the Partial and boolean?

mosheKerbel: what I'd expect a function that gets an object with at least a `selected` prop of type boolean
does it achieve that?

giladsegal: An object must be passed in order for the destructuring to work - `{selected}`
If we pass an empty object `selected` will be undefined, afterward it is being used like this `!selected` so whether you pass it or not - it will become a boolean.
The object cannot contain anything else besides `selected` because it is not being spread into the component (only destructed and passed specific props)

mosheKerbel: got it

PR #1776, opened 2020-02-11
<PopoverNext/> - add typing internally

mykas:


packages/wix-ui-core/src/components/popover/utils/shouldAnimatePopover.ts
```diff
@@ -1,4 +1,4 @@
-export const shouldAnimatePopover = (timeout = undefined) => {
+export const shouldAnimatePopover = ({ timeout }) => {
```
simast: if it's just one param then we may not need an object here. Also we have an implicit `any` type here. What do you think about changing this to?

```ts
import { PopoverProps } from '../Popover';

export const shouldAnimatePopover = (timeout: PopoverProps['timeout']) => {
```

mykas: Thanks

PR #1779, opened 2020-02-12
wsr-types: Add missing types for exported TableToolbar items

simast:
We do reexport more items from `wix-style-react/TableToolbar`:
https://github.com/wix/wix-style-react/blob/180020600474351c648ce32a3a8be8ac32181fed/src/TableToolbar/index.js#L1-L2


nitayneeman:
The thing is there are two ways to use them:
```jsx
// option 1 - cherrypicking each of them
import { TableToolbar, Item } from 'wix-style-react/TableToolbar';

// option 2 - the compound approach
import { TableToolbar } from 'wix-style-react/TableToolbar';

render() {
 return <TableToolbar.Divider/>;
}
```

We didn't export them consciously, in order to encourage our consumers to use the compound approach. 

But maybe it's not the best way to discipline our consumers? @mosheKerbel

mosheKerbel:
I say we merge it and when deprecating the old usage in next major version we will also clear this type export.

PR #1784, opened 2020-02-24
remove function that prevents contextMenu on right click cause this l…

yteslenk:
…ogic is handled by playable


packages/wix-ui-core/src/components/video/types.ts
```diff
@@ -55,6 +55,7 @@ export interface IPlayableConfig {
   poster?: string;
   modules?: {};
   texts?: ITextMapConfig;
+  preventContextMenu?: boolean;
```
mokhovyk: What is default value for preventContextMenu?

yteslenk: false

shlomitc:

PR #1796, opened 2020-03-09
Patch stylable files with comments to unify clasnamess

mykas:


mykas:
https://github.com/wix/wix-style-react/issues/5048

PR #1802, opened 2020-03-10
Add optional data-hook prop to LinearProgressBar component

karolinawix:


mykas:
Hi @karolinawix , in wix-ui-core our components has a pattern to accept: `data-hook` and not `dataHook`. You can check an example in here: https://github.com/wix/wix-ui/blob/master/packages/wix-ui-core/src/components/popover/Popover.tsx#L529

PR #1803, opened 2020-03-10
[Popover] allow moveBy to accept either x or y

ipanasenko:


mosheKerbel:
Your change makes sense but the test is failing because of the `Partial` use
https://pullrequest-tc.dev.wixpress.com/viewLog.html?buildId=6707835&buildTypeId=FedInfra_Package_WixUi&tab=buildLog&_focus=2447#_state=2417
Do you understand why?
I tried to debug it locally but the test pass.

PR #1804, opened 2020-03-11
NOQA|fix call player when ref instance is null

mokhovyk:


mokhovyk:
> Theoretically we don't even need the 'if' because of the try/catch. But it's probably cleaner this way.

Yes, it is true, but that try/catch was created in order to track (onError callback) unused/deprecated external player API method. e.g Youtube will change some method name.

PR #1807, opened 2020-03-16
wix-ui-core|TagsList: fixed driver type for clickOnTagByIndex

liorMar:


jonathanadler:
@zonzujiro 
Will this change break the santa build?
Are there any tests using this method?

jonathanadler:
> @zonzujiro
> Will this change break the santa build?
> Are there any tests using this method?

Wait I take that back, it shouldn't break anything, right 🤔 ?

zonzujiro:
@jonathanadler as far as I remember, this method, in general, not used anywhere.

jonathanadler:
> @jonathanadler as far as I remember, this method, in general, not used anywhere.

Cool, so i'm merging

PR #1808, opened 2020-03-16
`<ButtonNext />` - content width inherit, so ellipsis could be added

frontned:
This change is needed so I could add Text element with ellipsis in wix-style-react button


mykas:
Hey, I wouldn't add this here. I just tried applying this in wix-style-react button and it broke current implementation.

mykas:
Better if you do it directly in wix-style-react :)

frontned:
Closing the pull request

PR #1812, opened 2020-03-23
fixing fb player view in responsive editor

yteslenk:


packages/wix-ui-core/src/components/video/Video.st.css
```diff
@@ -9,8 +9,10 @@
     background-color: #000;
 }
 
-.playerContainer_facebook {
-    display: block !important;
+.responsiveMode{
```
yuriibut: add space

PR #1816, opened 2020-03-25
[wix-ui-core] <Pagination/> - added unidriver

jjaskulis:


packages/wix-ui-core/src/components/pagination/Pagination.spec.tsx
```diff
@@ -452,13 +455,18 @@ describe('Pagination', () => {
...
       const onChange = jest.fn();
-      const p = await render(
+      const driver = await createDriver(
         <Pagination totalPages={3} currentPage={1} onChange={onChange} />,
       );
 
-      Simulate.click(p.nextButton);
-      Simulate.click(p.getPage(2));
-      expect(onChange).toHaveBeenCalledTimes(2);
+      await eventually(
```
mykas: Any reason we need eventually here?

jjaskulis: turns out it's not really necessary, updated the pr.

PR #1824, opened 2020-04-01
`<Popover/>` - make sure that popover potal does not bubble up the events to its parent

mykas:
This PR is suppose to fix the issue where Popover portal for some unknown reason gets a noop function on onClick method coming from React itself. Could not find the reason for it, but this causes issues like this for our users: 
click
> I have a FloatingHelper on a TableActionCell in that table I also have onRowClick
the problem that I get onRowClick when I click on the FloatingHelper as well.. is there a way to prevent it?

### DIFF
--- a/packages/wix-ui-core/src/components/popover-next/popover-next.tsx
+++ b/packages/wix-ui-core/src/components/popover-next/popover-next.tsx
@@ -273,6 +273,7 @@ export class PopoverNext extends React.Component<
 
     if (this.appendToNode) {
       this.portalNode = document.createElement('div');
+      this.portalNode.addEventListener('click', e => e.stopPropagation());
       this.portalNode.setAttribute('data-hook', 'popover-portal');
       this.portalNode.setAttribute('data-loaded', 'false');
       /**
--- a/packages/wix-ui-core/src/components/popover/Popover.tsx
+++ b/packages/wix-ui-core/src/components/popover/Popover.tsx
@@ -387,6 +387,7 @@ export class Popover extends React.Component<PopoverProps, PopoverState> {
     this.appendToNode = getAppendToElement(appendTo, this.targetRef);
     if (this.appendToNode) {
       this.portalNode = document.createElement('div');
+      this.portalNode.addEventListener('click', e => e.stopPropagation());
       this.portalNode.setAttribute('data-hook', 'popover-portal');
       /**
        * reset overlay wrapping layer


mykas:
Here is an example in firefox that shows that event has been attached by the react library
Screenshot 2020-04-01 at 12 32 50

mykas:
The big question is: does this change could have impact on our users? could there be any case that rely on this?

shlomitc:
This is intentional behavior by React
https://reactjs.org/docs/portals.html#event-bubbling-through-portals
Here's a small reproducible example:
```jsx
() => {
  const [ shown, setShown] = React.useState(false);
  return (
    <div onClick={() => alert('will be triggered even on content click')}>
      <Popover shown={shown} appendTo={'window'}>
        <Popover.Element>
           <Button onClick={() => setShown(!shown)}>Click to open</Button> 
        </Popover.Element>
        <Popover.Content>
          <div>some content</div>
        </Popover.Content>
      </Popover>
    </div>
  )
}
```

you will see that even though the Popover is in a portal, the event is stilled bubbled, 

If some component would like to prevent this from happening (because it knows he is in a more prominent portal), then it's the responsibility of it to prevent it

```jsx
() => {
  const [ shown, setShown] = React.useState(false);
  return (
    <div onClick={() => alert('will not be triggered when stopping propagation from the portal element')}>
      <Popover shown={shown} appendTo={'window'}>
        <Popover.Element>
           <Button onClick={() => setShown(!shown)}>Click to open</Button> 
        </Popover.Element>
        <Popover.Conatent>
          <div onClick={e => e.stopPropagation()}>some content</div>
        </Popover.Content>
      </Popover>
    </div>
  )
}
```

I am not sure if it's a good idea to do the change you suggest by default, it will appear as a "magic" and unexpected behavior to how React works.

mykas:
I understand how to workaround it for users I just dont get the idea that popover content might be needed to be bubbled up outside the popover content 🤷‍♂️

mykas:
In the react example they showed that its great example of Modals when we want to catch events from Modal content click, but do we really need such case on our popover usecases? If not then maybe its worth doing thins in our wsr popover to prevent it 🤷‍♂️

mykas:
Honestly I got like two people complaining about this saying that clicking on popover content  triggers their parent component. Usually people are not aware about the bubblling mechanisms and they think this is a bug.

shlomitc:
@mykas - Can you compare this behavior to other common libraries? See if they handle it differently.

mykas:
sure

PR #1825, opened 2020-04-01
Allow passing popoverZIndex to AddressInput

NitayRabi:
Related to https://jira.wixpress.com/browse/BOLT-2029

### DIFF
--- a/packages/wix-ui-core/src/components/address-input/AddressInput.tsx
+++ b/packages/wix-ui-core/src/components/address-input/AddressInput.tsx
@@ -117,6 +117,8 @@ export type AddressInputProps = Pick<
   converterType?: Converter;
   /** Pass a custom class to the input element */
   inputClassName?: string;
+  /** popover z-index */
+  popoverZIndex?: number;
 };
 
 export interface AddressInputState {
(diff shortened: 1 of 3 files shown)


packages/wix-ui-core/src/components/address-input/AddressInput.tsx
```diff
@@ -117,6 +117,8 @@ export type AddressInputProps = Pick<
   converterType?: Converter;
   /** Pass a custom class to the input element */
   inputClassName?: string;
+  /** popover z-index */
+  popoverZIndex?: number;
```
shlomitc, line 121: I don't think we should mention `popover` in this level of API.
Popover is the implementation detail, what you actually want to pass is the z index for the options list or for the dropdown list

NitayRabi: So `optionsListZIndex`?

shlomitc: `optionsContainerZIndex` will be good

NitayRabi: Naming is the worst

PR #1827, opened 2020-04-03
[wix-ui-core] `<Popover/>` - refactor story

mykas:
This PR is refactoring Popover because currently it is very hard to add an example of usage.


packages/wix-ui-core/package.json
```diff
@@ -188,7 +188,7 @@
   },
   "babel": {
     "presets": [
-      "env"
+      "@babel/preset-env"
```
mykas, line 191: Here we need to switch to `@babel/preset-env` because yoshi 3 was already using babel 7 and this part was preventing it from switching to babel 7.

PR #1829, opened 2020-04-07
allow limiting initials to 1

ayal:


packages/wix-ui-core/src/components/avatar/avatar.spec.tsx
```diff
@@ -320,18 +320,55 @@ describe('Avatar', () => {
         expect(nameToInitials('John Doe')).toBe('JD');
       });
 
-      it('should be 3 initials given 3 name parts', () => {
+      it('should be 2 initials given 3 name parts', () => {
```
ayal, line 323: just a boyscout fixing wrong number here

packages/wix-ui-core/src/components/avatar/avatar.tsx
```diff
@@ -98,8 +98,8 @@ export class AvatarComponent extends React.Component<
   UNSAFE_componentWillReceiveProps(nextProps: AvatarProps) {
     if (
       !nextProps.imgProps ||
-      (!this.props.imgProps ||
-        nextProps.imgProps.src !== this.props.imgProps.src)
+      !this.props.imgProps ||
```
ayal, line 101: lint change

PR #1832, opened 2020-04-07
update appveyor node version

shlomitc:


AviVahl:
use "12" to get the latest stable v12 (and its security fixes!)

shlomitc:
`node  x86 package not found`, not going to put effort on fixing this as it's not in use anyway.

PR #1833, opened 2020-04-08
wix-ui-core: upgrade stylable to v3 and yoshi to v4

argshook:
Main goal of this PR is to update wix-ui-core to use stylable v3

it requires a few packages to be upgraded:

```diff
-"@stylable/dom-test-kit": "^1.0.7"
+"@stylable/dom-test-kit": "^3.4.1"

-"@stylable/cli": "^1.2.11"
+"@stylable/cli": "^3.4.0"

-"@stylable/core": "^1.0.7"
+"@stylable/core": "^3.4.0"

-"@ui-autotools/a11y": "^2.4.0"
+"@ui-autotools/a11y": "^5.0.0"

-"@ui-autotools/registry": "^4.0.0"
+"@ui-autotools/registry": "^5.0.0"

-"@ui-autotools/sanity": "^2.5.0"
+"@ui-autotools/sanity": "^5.0.0"

-"wix-ui-mocha-runner": "^0.1.6"
+"wix-ui-mocha-runner": "^1.0.0"

-"yoshi": "^3.28.0"
+"yoshi": "^4.55.1"

-"yoshi-style-dependencies": "^3.16.0"
+"yoshi-style-dependencies": "^4.37.1"

+"yoshi-stylable-dependencies": "^4.49.0"
```

# TODO

* [x] no typescript errors
  * [x] no errors in `*.meta.ts` files
* [x] successful `npm run build`
* [x] successful `yoshi lint`
* [x] green `yoshi test`
* [x] green `wix-ui-mocha-runner` https://github.com/wix/wix-ui/pull/1840
* [x] green `yoshi test --protractor`
* [x] green `autotools-sanity`
* [x] green `autotools-a11y`
* [x] green `test:performance`
  locally passes, but there's no bundle size comparison.
  unable to check without merge to master :(
* [x] no visual diffs


mykas:
Hey, just adding some more places where it needs to be upgraded.

Here we have direct babel cli usage https://github.com/wix/wix-ui/blob/master/packages/wix-ui-core/package.json#L72

It is important to switch from `babel-cli` to `@babel/cli` just like I did in my PR https://github.com/wix/wix-ui/pull/1827/files#diff-5424cc8fdc700f83a28f75167ffcc053R111

Another thing important is to unfreeze storybook version. Just make sure that storybook style-loader & css-loader versions are the same as in yoshi 4. (but that should be ok when you migrate).

shlomitc:
If this PR focuses only on the yoshi upgrade, I think we can release a patch version, we will get instant feedback from consumers as well

argshook:
> If this PR focuses only on the yoshi upgrade, I think we can release a patch version, we will get instant feedback from consumers as well

it's not only yoshi, it's also stylable v3.

with yoshi3 i and stylable 3 i didn't manage to get build to pass, so i upgraded yoshi too. as this will be a major which shouldn't be delayed, i think we can move forward with both upgrades

packages/wix-ui-core/package.json
```diff
@@ -120,8 +121,7 @@
     "@ui-autotools/a11y": "^2.4.0",
     "@ui-autotools/registry": "^4.0.0",
     "@ui-autotools/sanity": "^2.5.0",
-    "babel-cli": "^6.26.0",
-    "babel-preset-env": "^1.6.1",
+    "@ui-autotools/snap": "^3.0.0",
```
AviVahl, line 124: shouldn't be re-added. was removed in https://github.com/wix/wix-ui/pull/1831

argshook: thanks! i branched out not from most recent master. rebased now.

shlomitc:
@argshook - I don't see any changes in the actual components, which means it you are still using the stylable v1.

`yoshi` provides webpack configuration that supports v1 and v3, but unless did anything different, it would keep using v1. According to [this pr in Yoshi](https://github.com/wix/yoshi/pull/2023) in order to opt-in for v3 you should add add `yoshi-stylable-dependencies`.

Once this is done, webpack will kick in and then the component should no-longer support the old syntax. if it does - something is not working correctly

argshook:
> @argshook - I don't see any changes in the actual components, which means it you are still using the stylable v1.
> 
> `yoshi` provides webpack configuration that supports v1 and v3, but unless did anything different, it would keep using v1. According to [this pr in Yoshi](https://github.com/wix/yoshi/pull/2023) in order to opt-in for v3 you should add add `yoshi-stylable-dependencies`.
> 
> Once this is done, webpack will kick in and then the component should no-longer support the old syntax. if it does - something is not working correctly

thanks for this comment, i absolutely missed this dependency

argshook:
just fyi, since [the enlightenment](https://github.com/wix/wix-ui/pull/1833#issuecomment-610983373) i'm continuing to work on this and am updating stylable usage in each component.

it is not really possible to do this in smaller chunks, because stylable
is widely used. It could only be done if we merge PRs in chunks but
until everything is 100% migrated the merged chunks would make
wix-ui-core broken in many aspects.

reason above and the urgency of the matter leads me to just do this all
in one mega PR.

packages/wix-ui-core/src/components/autocomplete/Autocomplete.tsx
```diff
@@ -126,7 +128,8 @@ export class Autocomplete extends React.PureComponent<
     const inputProps = this._createInputProps();
     return (
       <InputWithOptions
-        {...style('root', { disabled }, this.props)}
+        className={style(classes.root, { disabled })}
```
shlomitc, line 131: Same here regarding props.className

packages/wix-ui-core/src/components/icon-with-options/IconWithOptions.tsx
```diff
@@ -45,11 +44,13 @@ export const IconWithOptions: React.FunctionComponent<
     iconUrl,
     fixedHeader,
     fixedFooter,
+    dataHook,
   } = props;
 
   return (
     <Dropdown
-      {...style('root', {}, props)}
+      className={style(classes.root)}
```
shlomitc, line 52: prop.className

packages/wix-ui-core/src/components/menu-item/menu-item.tsx
```diff
@@ -23,7 +23,7 @@ export const MenuItem: React.FunctionComponent<MenuItemProps> = props => {
 
   return (
     <div
-      {...style('root', { selected, highlighted, disabled }, props)}
+      className={style(classes.root, { selected, highlighted, disabled })}
```
shlomitc, line 26: prop.className

packages/wix-ui-core/package.json
```diff
@@ -82,7 +82,7 @@
   },
   "dependencies": {
     "@loadable/component": "^5.12.0",
-    "@stylable/dom-test-kit": "^1.0.7",
+    "@stylable/dom-test-kit": "^3.4.1",
```
NitayRabi, line 85: Can we use this major to get this out of dependencies and into `peer` ? This really made our life hard trying to update, and it's getting to consumers not really using it, either that or abstracting the usage in wix-ui-test-utils

packages/wix-ui-core/package.json
```diff
@@ -110,10 +110,10 @@
   "devDependencies": {
     "@applitools/eyes-storybook": "^3.3.1",
     "@babel/cli": "^7.8.4",
-    "@babel/preset-env": "^7.9.0",
     "@storybook/react": "^5.0.0",
-    "@stylable/cli": "^1.2.11",
-    "@stylable/core": "^1.0.7",
+    "@stylable/cli": "^3.4.0",
+    "@stylable/core": "^3.4.0",
+    "@stylable/runtime": "^3.4.1",
```
NitayRabi, line 116: When you bring `yoshi-stylable-dependencies` - doesn;t it bring these?

packages/wix-ui-core/src/components/avatar/avatar.spec.tsx
```diff
@@ -6,7 +6,7 @@ import { ReactDOMTestContainer } from '../../../test/dom-test-container';
 import { Avatar, AvatarProps } from '.';
 import { nameToInitials } from './util';
 import { avatarDriverFactory } from './avatar.uni.driver';
-import styles from './avatar.st.css';
+import * as styles from './avatar.st.css';
```
NitayRabi, line 9: `StylableDOMUtil` take all of the module's import result?? why not maintain similar API change...

packages/wix-ui-core/src/components/dropdown/Dropdown.tsx
```diff
@@ -1,15 +1,14 @@
...
+import onClickOutside, { InjectedOnClickOutProps } from 'react-onclickoutside';
+import { style, classes } from './Dropdown.st.css';
 import { Popover, Placement, PopoverProps } from '../popover';
 import { DropdownContent, IDOMid } from '../dropdown-content';
 import { Option } from '../dropdown-option';
 import { CLICK, HOVER, OPEN_TRIGGER_TYPE } from './constants';
 
 export type DropdownProps = Pick<PopoverProps, 'fixed' | 'flip' | 'moveBy'> & {
+  /** hook for testing purposes */
+  'data-hook'?: string;
```
NitayRabi, line 11: We somtimes have `data-hook` and sometimes `dataHook`, and it's reused and a lot of places - maybe we can have (not in this PR) `DefaultProps` with either `data-hook` or `dataHook`

packages/wix-ui-core/src/components/loadable/Loadable.tsx
```diff
@@ -89,7 +89,7 @@ export class Loadable<LoadableExports> extends React.Component<
         ).slice(0)}" to access exported component property. `
       );
     }
-    return moduleItem[key];
+    return moduleItem[key] || moduleItem;
```
NitayRabi, line 92: Is this an unrelated fix?

packages/wix-ui-core/src/components/popover-next/popover-next.tsx
```diff
@@ -13,13 +13,7 @@ import Loader from './components/Loader';
 import Portal from './components/Portal';
 import { PopperProps } from './components/Popper';
 
-import style from '../popover/Popover.st.css';
-
-import {
-  AttributeMap,
-  attachStylesToNode,
-  detachStylesFromNode,
-} from '../../utils/stylableUtils';
```
NitayRabi, line 22: Whoohoo

packages/wix-ui-core/src/components/popover/Popover.tsx
```diff
@@ -145,6 +141,11 @@ export type PopoverType = PopoverProps & {
   Content?: React.FunctionComponent<ElementProps>;
 };
 
+const attachClasses = (node: HTMLElement, classnames: string) =>
+  node && node.classList.add(...classnames.split(' '));
+const detachClasses = (node: HTMLElement, classnames: string) =>
+  node && node.classList.remove(...classnames.split(' '));
```
NitayRabi, line 147: I saw this in several places - can this be something we achieve with `classNames` ? or reused?

packages/wix-ui-core/src/components/tooltip/Tooltip.tsx
```diff
@@ -52,6 +52,10 @@ export interface TooltipProps {
   minWidth?: number;
   /** content element maxWidth value */
   maxWidth?: number;
+
+  className?: string;
+  /** hook for testing purposes */
+  'data-hook'?: string;
```
NitayRabi, line 58: Maybe `DefaultProps` with `className` & `dataHook` & `data-hook` ?

packages/wix-ui-core/src/types/external-types.d.ts
```diff
@@ -1 +1 @@
-declare module '*.md';
+import 'yoshi/types';
```
NitayRabi, line 1: Isn't it better to put it in `tsconfig`?

packages/wix-ui-core/src/utils/withStylable/withStylable.tsx
```diff
@@ -8,6 +8,38 @@ function isComponentInstance(value: any): value is React.Component {
...
+  className?: string;
+  [attributeName: string]: StateValue | undefined;
+}
+
+interface InheritedAttributes {
+  className?: string;
+  [props: string]: any;
+}
+
+type RuntimeStylesheet = {
```
NitayRabi, line 27: Why are we declaring a separate definition for `RuntimeStylesheet`?

packages/wix-ui-core/package.json
argshook: sure thing, i'm aware of the issues with having this as part of `dependencies`

packages/wix-ui-core/src/components/avatar/avatar.spec.tsx
argshook: > StylableDOMUtil take all of the module's import result?

yep

https://github.com/wix/stylable/blob/master/packages/dom-test-kit/src/stylable-dom-util.ts#L18

packages/wix-ui-core/src/components/popover/Popover.tsx
argshook: it is only related to `Popover` & `popover-next`.

these need to put some `className`s on portalled nodes, which are not
treated as react components. Thus, `classnames` helper wouldn't really
be that helpful, when DOM has same purpose API.

Maybe they portalled node should be considered as react component through some refs, but i don't want to delve into that in this PR

packages/wix-ui-core/src/components/tooltip/Tooltip.tsx
argshook: you mean something like `interface Props extends DefaultProps`?

maybe, but it could become another `WixComponent`. i'd rather do these decisions in separate PR

packages/wix-ui-core/src/components/loadable/Loadable.tsx
argshook: this is a hotfix for Loadable, to remove the assumption that each "loadable" item either has one specific key, or a `default`.

when this component was built it was assumed that it will be used only for dynamically loading components. however, there are cases where `.st.css` is also dynamically loaded.

it worked before well, because `.st.css` would export only one thing and assumption from `Loadable` worked well

since stylable v3 `.st.css` exports an object and multiple things are needed (namely `style` and `classes`) and that assumption no longer holds.

---

the whole `Loadable` implementation is not too good and there already were attempts at fixing it by @mykas but we now have other priorities :(

packages/wix-ui-core/src/components/tooltip/Tooltip.tsx
NitayRabi: I don't like the extend approach but using `&` for type composition is quite nice (and it's not an entire magical base class with magical functionality). But I agree it's a separate topic

packages/wix-ui-core/src/utils/withStylable/withStylable.tsx
argshook: to be honest i'm not entirely sure what to do with this `withStylable` HOC. It should probably be removed entirely but maybe in separate PR

to answer your question: these types were previously defined in
`globals.d.ts` which has now been changed to use what stylable
recommends

however, `withStylable` was written for stylable v1 and with changed
types typescript yells at me. thus a quick copy paste to calm it down.

because we're upgrading to stylable v3 we should probably upgrade
`withStylable` HOC too.

but, maybe it's not needed at all anymore?

but, there are some users who rely on it, so removing it is another
breaking change.

thus, there were too many questions to answer for an already big PR so i
decided to copy-paste and shush `tsc` for now

packages/wix-ui-core/src/types/external-types.d.ts
argshook: i was blindly following yoshi migration guide https://wix.github.io/yoshi/docs/migration/version-4#typescript-definitions-for-yoshi

you think it should live in `tsconfig.json` `typeRoots` or `includes` or somewhere else maybe?

packages/wix-ui-core/package.json
argshook: it brings these 

https://github.com/wix/yoshi/blob/17df7fd37b5b2f10f020fb6efcebe80099f2f44d/packages/yoshi-stylable-dependencies/package.json#L11-L14

and also `@stylable/core` and `@stylable/runtime` but as "phantom dependencies"

i think it's better to be explicit and it shouldn't be harmful, because of `^`, wdyt?

packages/wix-ui-core/src/components/dropdown/Dropdown.tsx
argshook: > We somtimes have `data-hook` and sometimes `dataHook`

yeah that's an issue and it's a topic on its own

ideally, there should be no ambiguity and only one wait to set it,
probably with `data-hook` as it's closer to html.

but practically it can get treated as duplication in certain scenarios

consider this use case with enzyme:

```js
const Best = props =>
  <div data-hook={props['data-hook']}>
    lalala 100s of nested components, whatever
  </div>

const App = () =>
  <div>
    <Best data-hook="batman"/>
  </div>

const wrapper = mount(App)
wrapper.find('[data-hook="batman"]').length === 2
wrapper.debug()

// output of .debug
<div>
  <Best data-hook="batman">
    <div data-hook="batman"/>
  </Best>
</div>
```

to play against this we should probably use `dataHook` when passing the hook between components, but `data-hook` when applying it to actual HTML element.

but this requires good discipline and is additional overhead.

this can become a prolonged discussion which imo should anyway be tackled in separate PR

PR #1837, opened 2020-04-14
[wix-ui-core] <Pagination/> - render empty button when not showing prev/next buttons

jjaskulis:


mykas:
All good, but we need a unit test for this :)

PR #1841, opened 2020-04-19
`<Tooltip/>`- disable onShow and onHide executions when tooltip is disabled

TalZa1810:
Prevent `onShow` and `onHide` executions when `Tooltip` is disabled.

Related issue: https://github.com/wix/wix-style-react/issues/5367


mykas:
perfect, Tal ❤️

PR #1864, opened 2020-06-01
wix-ui-core: fix(Popover): pass `clickOutsideRef`

argshook:
otherwise it will always be `{ current: null }` and `clickOutside`
component will not work properly.

This is already done in `popover-next`, exactly the same way: https://github.com/wix/wix-ui/blob/master/packages/wix-ui-core/src/components/popover-next/popover-next.tsx#L435

required for https://github.com/wix/wix-style-react/pull/5445


argshook:
two visual diffs that are approved. They were caused by newer chrome and it's updated background color of a button (slightly less grey)

PR #1878, opened 2020-07-03
Emoji Icons that can't be part of general icons because they are 30x30px

ZilvinasMaciulis:


ZilvinasMaciulis:
@argshook @shlomitc I meant to upload 30x30 icons to the system folder. Should I reupload?

shlomitc:
It is already in the system folder, I just don't understand why it's 30px size.
aren't all of our icons are 24x24 and 18x18?

TalZa1810:
I see that we have also in 36px:
https://github.com/wix/wix-ui/blob/master/packages/wix-ui-icons-common/src/system/raw/MoveLarge.svg?short_path=bdcc949

https://github.com/wix/wix-ui/blob/master/packages/wix-ui-icons-common/src/system/raw/StarLarge.svg?short_path=80258e2

It's not consistent.

ZilvinasMaciulis:
@shlomitc system icons are designed for specific components and are not always 24x24. Other than that are the General Icons. Those are strictly 24 or 18. 

It's ok to have these 30x30 icons as emoji

shlomitc:
@ZilvinasMaciulis - OK, but I think we should somehow decide on naming conventions because "***Large" is not longer indicative.

TalZa1810:
Should I merge this? Rename this? 
Do we want to rename the existing icons as well? and if so what do you think is the right approach of naming those icons?

shlomitc:
Renaming means a breaking change, so let's not do it for all.
Let's decide on conventions and follow them with these new icons.

TalZa1810:
What do you think about renaming these file with the number of pixles as suffix instead of `Large`?
For example: `FaceDisapointed30.svg `or `FaceDisapointedThirty.svg`

shlomitc:
sounds good

PR #1879, opened 2020-07-05
<FaceNeutral/>- remove fill attribute

TalZa1810:


TalZa1810:
@shlomitc  I didn't get this part from the guidlines. Is it related to UXs? 
`Make sure SVGs are correctly exported from Illustrator/Sketch/Figma, meaning they should merge all layers into one, and apply the masks, which will result in an SVG with a single path.`

shlomitc:
@TalZa1810 - yes, it's the UX responsibility

PR #1885, opened 2020-07-16
New Marketplace icons to Add

ZilvinasMaciulis:


TalZa1810:
Hi @ZilvinasMaciulis ,
Just s tiny question before merging- why do we have the `Wix` prefix for some of the icons?
why not just having a `MarketPlace` icon?

ZilvinasMaciulis:
@TalZa1810  

These icons represent Wix brands only like Wix Chat, Wix Forms.

PR #1892, opened 2020-07-30
Help Large icon to add

ZilvinasMaciulis:
Will be used in close button  inside a new Custom modal


TalZa1810:
I noticed that when the icon suffix is a number, it adds this directly to the "Normal" storybook page. We can take a look with the [Faces icons example](https://wix-wix-ui-icons-common.surge.sh/?selectedKind=Icons%2FSystem&selectedStory=Normal&full=0&addons=0&stories=1&panelRight=0).

I think that we can either decide on an a range approach for icons- meaning `Large` suffix is for icons with height and width of "24x24" till "36x36" (as an example). 

Another approach is going with the approach of the general icons. 
In this case, we can't rename the existing icons (which in some cases are in different sizes than what is defined with the "general icons"). Moreover, we should decide what happens when adding a new icon which should be in a different size approach of the "general icons" (the Faces icons for example).

@shlomitc wdyt?
PM if something is not clear.

shlomitc:
@TalZa1810 - following your number convention, I think it's redundant to categorize the system icons by size since we're anyway going to have a range of it.

Anyway, we're going to do some rework on the entire icons metadata, so I wouldn't mind that much about it at the moment

TalZa1810:
ok so I'll leave it with the number suffix and merge this. thanks.

PR #1895, opened 2020-08-03
Dropdown - Mark onDeselect prop as optional

NitzanSirota:
Mark `onDeselect` prop as optional because it is only relevant for multiselection flow and not for all flows.


NitzanSirota:
Maybe we should add  `callback && callback()` here :https://github.com/wix/wix-ui/pull/1895/commits/cc3b58af0ee6b4b547abb886d01c5b43da834ad2#diff-0e7ed48b3654b37ad53c8528489c3229L276

argshook:
@NitzanSirota 

> Maybe we should add `callback && callback(option)` here :
> 
> https://github.com/wix/wix-ui/blob/cc3b58af0ee6b4b547abb886d01c5b43da834ad2/packages/wix-ui-core/src/components/dropdown/Dropdown.tsx#L276

maybe add both `onSelect` and `onDeselect` to `defaultProps`?

like so

```
Dropdown.defaultProps = {
  onSelect: () => {},
  onDeselect: () => {}
}
```

wdyt?

PR #1896, opened 2020-08-03
Controlled tooltip

ariel7342:


packages/wix-ui-core/src/components/tooltip/Tooltip.tsx
```diff
@@ -53,7 +55,10 @@ export interface TooltipProps {
   minWidth?: number;
   /** content element maxWidth value */
   maxWidth?: number;
-
+  /** whether controlled or not*/
+  controlled?: boolean;
+  /** when controlled enables tooltip show or not*/
+  show?: boolean;
```
argshook, line 61:
```suggestion
  shown?: boolean;
```
let's keep consistency with `Popover`

packages/wix-ui-core/src/components/tooltip/Tooltip.tsx
```diff
@@ -53,7 +55,10 @@ export interface TooltipProps {
   minWidth?: number;
   /** content element maxWidth value */
   maxWidth?: number;
-
+  /** whether controlled or not*/
+  controlled?: boolean;
```
argshook, line 59: no need for additional prop, `shown` alone should indicate that component is controlled

packages/wix-ui-core/src/components/popover/Popover.tsx
```diff
@@ -127,6 +130,7 @@ export interface PopoverProps {
    * When true - onClickOutside will be called only when popover content is shown
    */
   disableClickOutsideWhenClosed?: boolean;
+  controlled?: boolean;
```
argshook, line 133: i don't think you need this prop, Popover, when used with `shown` prop get controlled.

ariel7342: so what should I do in this case? https://github.com/wix/wix-ui/blob/c4e623c6239fc85a7ba8006308cae044332d0851/packages/wix-ui-core/src/components/popover/Popover.tsx#L517

argshook: originally this code is:

```
const shouldRenderPopper = isMounted && (shouldAnimate || shown);
```

the following code:

```
const shouldRenderPopper = controlled
      ? shouldAnimate || shown
      : isMounted && (shouldAnimate || shown);
```

does not seem to change much. The part with `shouldAnimate || shown` is
still in both `if` branches, just without `isMounted`

`isMounted` is set during `componentDidMount`, thus regardless of
whether `controlled` is `true` or `false`, functionality seems to be the
same.

the `controlled` or not idea is "controlled" by `shown` prop.

maybe i'm missing something?

ariel7342: the problem was when I tried to test tooltip component is shown in wix-ui-tpa , it wasn't mounted in first run

argshook: there was likely a different issue, because wix-ui-core has a test for `shown={true}` https://github.com/wix/wix-ui/blob/master/packages/wix-ui-core/src/components/popover/Popover.spec.tsx#L108-L119

unfortunately i can't currently confirm that test is working, due to artifactory issues :(

ariel7342: so the only problem Im facing is that this test is not passing no matter what - https://github.com/wix/wix-ui/blob/df2b7e45e476d523196ef1853384421efef662a5/packages/wix-ui-core/src/components/tooltip/Tooltip.spec.tsx#L55

argshook:
@ariel7342 

thanks for working on this and sorry for my late response. I'm taking over this feature and will finalize implementation.

It will work same as in `Popover`: given `shown={true}` it will become controlled.

the reason  you were not able to pass all tests is because `Tooltip.spec.tsx` runs tests for `Tooltip` and `tooltip-next` - a (hopefully) next generation of the tooltip. Thus, the `shown` prop need to be added to the both.

PR #1900, opened 2020-08-16
refactor radio button in order to support a11y

Arielliv:
In order to enable a11y focus ring in radion button - https://jira.wixpress.com/browse/EE-24790
I'm passing `checkedIcon` to `RadioButton` with 'wixSdkShowFocusOnSibling' class, in order to enable the focus ring of the radio button. 
please review it :) @argshook

### DIFF
--- a/packages/wix-ui-core/src/components/radio-button/RadioButton.tsx
+++ b/packages/wix-ui-core/src/components/radio-button/RadioButton.tsx
@@ -117,27 +117,27 @@ export class RadioButton extends React.Component<
         aria-checked={checked ? checked : false}
         {...filterDataProps(this.props)}
       >
-        <input
-          type="radio"
-          className={classes.hiddenRadio}
-          disabled={disabled}
-          required={required}
-          onFocus={this.onFocus}
-          onBlur={this.onInputBlur}
-          checked={checked}
-          value={value}
-          name={name}
-          tabIndex={tabIndex}
-          onChange={() => null}
-          onKeyDown={this.handleInputKeyDown}
-          ref={radio => (this.radioRef = radio)}
-          aria-label={this.props['aria-label']}
-        />
         <span
           className={classes.icon}
           onMouseEnter={this.onHover}
           onMouseLeave={onIconBlur}
         >
+          <input
+            type="radio"
+            className={classes.hiddenRadio}
+            disabled={disabled}
+            required={required}
+            onFocus={this.onFocus}
+            onBlur={this.onInputBlur}
+            checked={checked}
+            value={value}
+            name={name}
+            tabIndex={tabIndex}
+            onChange={() => null}
+            onKeyDown={this.handleInputKeyDown}
+            ref={radio => (this.radioRef = radio)}
+            aria-label={this.props['aria-label']}
+          />
           {checked ? checkedIcon : uncheckedIcon}
         </span>
         <span className={classes.label}>{label}</span>


packages/wix-ui-core/src/components/radio-button/RadioButton.tsx
```diff
@@ -117,27 +117,27 @@ export class RadioButton extends React.Component<
         aria-checked={checked ? checked : false}
         {...filterDataProps(this.props)}
       >
-        <input
```
argshook, line 120: hey, thanks for contributing!

could you elaborate on how moving `<input>` to be nested in `<span className={classes.icon}>` helps?

jonathanadler:
So it's just switching the position of the `input`?

jonathanadler:
@argshook what do you think?
looks legit

packages/wix-ui-core/src/components/radio-button/RadioButton.tsx
ariel7342: yes, it like I wrote above because the focus is on the `input` element (which is hidden) we don't see the blur focus ring.
a solution for that is using `wixSdkShowFocusOnSibling` class, which will enable focus ring on sibling element of the input.
currently, the sibling element is span, and not the actual Icon which been passed
In order to support a11y focus ring I moved the span above, and that way the `Icon` element is the sibling of the `input` element, and he gets the focus ring :)

argshook: unfortunately i can't allow such change, because:

1. it's very platform specific, wix-ui-core doesn't know (and it
   shouldn't) anything about `wixSdkShowFocusOnSibling`
2. there's no test to check for specific hierarchy in HTML. A developer
   in future might refactor this code without any knowledge about it and
   thus break your usage.
3. i don't use wixSDK so i'm not aware of it's API, but anyway a
   property of `wixSdkShowFocusOnSibling` feels very odd. Is there no
   other way to know when an item should be focused? something like just
   `wixSdkShowFocus`?

ariel7342: ok, so we currently blocked, which means we cannot continue a lot of plans we had, like new templates for subscription plans in stores and etc.
1. how do you support a11y then?
2. I never heard of hierarchy test before, If that what's missing please send me an example and I'll do it
3. I don't know any other way to enable focus ring on *sibling* and not the current focused element, if you do, let's use it then 

we need to push this a11y for a solution, lets work together to achieve it

ariel7342: BTW this is viewer way for supporting a11y in case of siblings, we don't have any other way 
https://github.com/wix-private/thunderbolt/blob/1eadb80676dc71c371a86ee6bdde53ad567197ec/packages/thunderbolt-becky/src/types/css.d.ts#L68

jonathanadler: I'm not really sure using something like `wixSdkShowFocusOnSibling` will be needed.
This component should be able to show the correct focus ring.
I think this means that we'll need to manually control focus for the component.
ie add the correct `tabindex` and `aria-` attrs

jonathanadler: btw @ariel7342 a RadioButton will soon be released in wix-ui-tpa

ariel7342: The focus ring supposes to be on the `Icon` element, the input is hidden, currently, it's not working like this.
What do you mean by `manually control focus for the component`, it will still need to be focused on the input and the focus ring will still have to be on the `Icon`
When does it suppose to be realised ?  @jonathanadler  @argshook

jonathanadler: Like I wrote, manually controlling the focus means adding the correct `tabindex` and `aria-` attributes.

ariel7342: I still don't get how will it solve the focus ring issue if the focus will never be on the `Icon` element
we will need a solution like `wixSdkShowFocusOnSibling`

argshook: > I never heard of hierarchy test before, If that what's missing please send me an example and I'll do it

there's nothing special about "hierarchy test", what i mean is that this
specific hierarchy has to be somehow covered with test. That's because
we would require exactly `span > input` to be in code and without test
no one would know that it's a requirement.

that's not a good approach for a library that calls itself `core`. A
specific order of elements tailored for one platform is not really core
anymore.

---

the current state is (correct me if i'm wrong):

1. it is possible to style the icon of RadioButton in wix-ui-core
   without any changes. Just set a className on `checkedIcon` and voila
2. it is possible to know when RadioButton is focused (through a
   combination of `className` prop and css `:focus`)
3. it is NOT easily possible to get exact style for focus ring?

i would guess number 3 should be accessible from some scss import or
similar. Worst case it can be just hardcoded?

ariel7342: currently, when the input is focused, my `icon` that I pass is covered with another span, so there is no way for me to trigger focus ring on my icon... 
viewer solution for cases like this is `wixSdkShowFocusOnSibling`, but when we have <input/><span>{icon}</span>
the icon is not sibling, which means we cannot do anything
from my code, it wouldn't be possible to do anything for adding focus ring on `icon` when the input in radioButton if focused 
1. `checkedIcon` - check is not the case, I need to cover focus case
2. I don't think its possible in our case
3. I'm not sure I understand the question

I really need a solution because its blocking us right now

ariel7342: please take my project and show me how it's possible today, I really don't see a way right now @argshook  
you can see our usage of `RadioButton`
https://github.com/wix-private/ecom/tree/master/client/wixstores-client/wixstores-client-product-page

argshook: correct approach is to use stylable

```
import { classes } from './your-radio-button-styles.st.css';

// ...later

<RadioButton
  label="Radio Gaga"
  className={classes.radioButton}
  checkedIcon={<span className={classes.icon}>🔘</span>}
  uncheckedIcon={<span className={classes.icon}>⚪</span>}
  tabIndex={0}
/>
```

```
:import {
  -st-from: "wix-ui-core/dist/src/components/radio-button/RadioButton.st.css";
  -st-default: RadioButton;
}

.radioButton {
  -st-extends: RadioButton;
}

.radioButton:focused .icon {
  border: 1px solid blue; // or any style you need
}
```

this way you have full control of the styling, no need for any changes
in `wix-ui-core`. I see your project already uses yoshi and
`wix-ui-core: ^3`, therefore there should be no problem to add this
styling.

ariel7342: We are not supposed to duplicate focus ring class implementation, the viewer has it.
If I'll create a class for it today, and the viewer will change their style for focus ring, I'll break convention 
we are not supposed to do it in our code, but to use viewer focus ring or in other cases `wixSdkShowFocusOnSibling`
@argshook

argshook: > We are not supposed to duplicate focus ring class implementation, the viewer has it.

and is there no possible way to reuse it?

eranshabi: you need to reuse the viewer implementation of the class, I don't think there is a way

jonathanadler: @ariel7342 can you please send a live example?

jonathanadler: The question here is how does the viewer identify which elements to add the focus ring on?
Cause, like @argshook said, changing the dom structure this way can be easily changed in the future, and theoretically, the outside environment shouldn't rely on the dom structure of this component.
The outside environment **should** rely on the behaviour of the component.
So we need the behaviour of the focus in this component to be as expected (which it seems it isn't at the moment)

We need to understand how the viewer's focus ring works, and then adjust the logic of the component accordingly

ariel7342: I cannot upload videos here @jonathanadler, I'll send the examples in `ui-tpa` thread
BTW, why we even need the span?  can't we just do <><input/>{icon}</> ?
who can we talk in order to proceed from the viewer?

jonathanadler: I meant can you send a link for a sure with focus rings on?

ariel7342: you can use this site - https://arielliv.wixsite.com/mysite-2/product-page/product

jonathanadler: Great, 
So as you can see in the image, what we need to do is assign the correct tabindex to the element we want to be focused.
In this case I actually think it should be the wrapper of the radiobutton, and not the circle itself.
To fix this, all we need to do is add `tabIndex="0"` to the wrapper (which I manually did here:)
image

jonathanadler: From what I can see, we also need to accept `space` press as a selection

jonathanadler: btw,
Note that the ColorPicker component above the radiobuttons with the bug, is a wix-ui-tpa component that also uses this wix-ui-core RadioButton component, and a11y works good there.
So @ariel7342 maybe you could look at the ColorPicker's implementation [here](https://github.com/wix/wix-ui-tpa/blob/master/src/components/ColorPicker/ColorPicker.tsx) and see why it's not working in your case.
It might be that the correct implementation should be in wix-ui-tpa/RadioButton, and once you'll use that component, this issue will be resolved

ariel7342: So what do you suggest ? should I wait for `ui-tpa` release and use radioButton from there?

PR #1901, opened 2020-08-17
New directions for chevron large

ZilvinasMaciulis:


TalZa1810:
@ZilvinasMaciulis 
I'm not sure about the naming.. `ChevronDownLargeSmall`? two suffixes?

ZilvinasMaciulis:
Yes. We have several chevron icon sizes for the same artboard. Meaning that both 24 and 18 artboards can contain large and small chevrons.

ZilvinasMaciulis:
image

here is some visual explanation

TalZa1810:
Apparently we have already some icons with the same naming:
- `RectangleLargeSmall` 
- `TriangleLargeSmall`
- `CircleLargeSmall`

 It's very confusing IMO. 

FYI @shlomitc

shlomitc:
I agree it is confusing. let's try to find a better convention

PR #1904, opened 2020-08-20
<NavStep /> - add className prop which is sent by <NavStepper />

OriMarron:
This PR fixes a bug that started with migration to Stalable 3 in the `<NavStepper>` component.

`<NavStepper>` is sending a `className` props (https://github.com/wix/wix-ui/blob/master/packages/wix-ui-core/src/components/nav-stepper/NavStepper.tsx#L37-L38) to its child which is a `<NavStep>`. However, currently the `NavStep` component ignores this prop and does not render it in the `<li>` element.

So I added this prop which is sent to the call to `st(...)`.

I tested locally with `wix-ui-premium-purchase` which extends `NavStepper`. Current behavior is that the child elements have no styling. This change fixes the issue.

### DIFF
--- a/packages/wix-ui-core/src/components/nav-stepper/NavStep.tsx
+++ b/packages/wix-ui-core/src/components/nav-stepper/NavStep.tsx
@@ -5,12 +5,20 @@ import { StepProps } from '../stepper';
 export type ExternalNavStepProps = Partial<StepProps> & {
   disabled?: boolean;
   children: React.ReactNode;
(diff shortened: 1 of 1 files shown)


argshook (dismissed):
Awesome, thanks for investigating!

EgoziE (approved):
Looks good, I'll merge it when

OriMarron:
> Looks good, I'll merge it when CI is green

🙏

PR #1932, opened 2020-09-17
[wix-ui-icons-common] Add icon search to storybook

gendeld:


packages/wix-ui-icons-common/stories/components/search-input/SearchInput.tsx
```diff
@@ -0,0 +1,25 @@
+import React, { ChangeEvent } from "react";
```
shlomitc, line 1: Technically component this should go to `wix-storybook-utils` since it's a UI element.
The only thing that could stay here is the css layout of it (putting it in the right place, fixating the width, etc...)

packages/wix-ui-icons-common/stories/components/category-list/CategoryList.spec.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+const search = async (query: string, searchInput: HTMLInputElement) => {
+  const inputEvent = createEvent.input(searchInput, {
+    target: { value: query },
+  });
+
+  fireEvent(searchInput, inputEvent);
+
+  await waitForDomChange();
+};
```
shlomitc, line 24: this should go into the driver

packages/wix-ui-icons-common/stories/components/category-list/CategoryList.spec.tsx
```diff
@@ -0,0 +1,77 @@
...
+    const driver = new CategoryListDriver(baseElement);
+    const searchInput = driver.getSearchInput();
+
+    expect(driver.getCategoryRows()).toHaveLength(3);
+
+    const [edit] = iconsMetadata;
+    const query = edit.title;
+    await search(query, searchInput);
+
+    expect(driver.getCategoryRows()).toHaveLength(1);
```
shlomitc, line 75: I think it will be stronger to also make sure you get the correctly filtered value, instead of just checking that one row appears

packages/wix-ui-icons-common/tsconfig.json
```diff
@@ -12,7 +12,8 @@
     "jsx": "react",
     "types": ["node", "jest"],
     "lib": ["dom", "esnext"],
-    "resolveJsonModule": true
+    "resolveJsonModule": true,
+    "esModuleInterop": true
```
shlomitc, line 16: just make sure the output of the icons is not changed

packages/wix-ui-icons-common/stories/components/category-list/CategoryList.driver.tsx
```diff
@@ -0,0 +1,27 @@
+import dataHooks from "../../dataHooks";
+
+const getAllByDataHook = (baseElement: HTMLElement, dataHook: string) =>
+  baseElement.querySelectorAll(`[data-hook="${dataHook}"]`);
+
+const getByDataHook = (baseElement: HTMLElement, dataHook: string) =>
+  baseElement.querySelector(`[data-hook="${dataHook}"]`);
+
+export default class CategoryListDriver {
```
shlomitc, line 9: We usually prefer creating drivers that are stateless (simple functions).

shlomitc (changes requested):
Excellent job 👍
I think that's much more clear than the previous draft.
It's really easy to follow and understand, also code is structured nicely.
just a few comments

packages/wix-ui-icons-common/stories/components/category-list/CategoryList.driver.tsx
gendeld: It's not stateful 🤔
The class syntax just makes it easier in this context to use methods within methods given the this keyword.
Otherwise I'd need to define the functions and then return them.
More of a stylistic thing, should I change it to a function anyway?

packages/wix-ui-icons-common/stories/components/category-list/CategoryList.driver.tsx
```diff
@@ -0,0 +1,49 @@
...
+    this.baseElement = baseElement;
+  }
+
+  getSearchInput = () =>
+    getByDataHook(
+      this.baseElement,
+      dataHooks.categorySearchInput
+    ) as HTMLInputElement;
+
+  getCategoryTitles = () =>
```
shlomitc, line 27: We prefer returning primitives instead of elements. In the `unidriver` world, we don't want consumers to interact with the internal elements.

read more here: https://github.com/wix/wix-style-react/blob/master/docs/contribution/TEST_DRIVERS_GUIDELINES.md#drivers-apis

so my suggestion is to return `getAmountOfCategoryTitles()`

packages/wix-ui-icons-common/stories/components/category-list/CategoryList.driver.tsx
```diff
@@ -0,0 +1,49 @@
...
+  baseElement.querySelector(`[data-hook="${dataHook}"]`);
+
+export default class CategoryListDriver {
+  baseElement: HTMLElement;
+
+  constructor(baseElement: HTMLElement) {
+    this.baseElement = baseElement;
+  }
+
+  getSearchInput = () =>
```
shlomitc, line 21: This is internal, so no need to expose it as a class method

packages/wix-ui-icons-common/stories/components/category-list/CategoryList.driver.tsx
shlomitc: in our libraries, we do return an object of functions.
In this case, I agree that it's just a matter of style since this driver is not exposed to consumers, so you can keep it that way

packages/wix-ui-icons-common/stories/components/category-list/CategoryList.driver.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+  search = async (query: string) => {
+    const searchInput = this.getSearchInput();
+    const inputEvent = createEvent.input(searchInput, {
+      target: { value: query },
+    });
+
+    fireEvent(searchInput, inputEvent);
+
+    await waitForDomChange();
```
shlomitc, line 47: This is a deprecated method, you should use `waitFor` instad
https://testing-library.com/docs/dom-testing-library/api-async#waitfor

packages/wix-ui-icons-common/stories/components/search-input/SearchInput.tsx
gendeld: https://github.com/wix/wix-ui/pull/1935

packages/wix-ui-icons-common/tsconfig.json
gendeld: Good catch 😅

shlomitc (approved):

PR #1935, opened 2020-09-21
[wix-storybook-utils] Add search input

gendeld:


packages/wix-storybook-utils/src/ui/search-input.js
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classnames from 'classnames';
+import Input from './input';
+import Search from 'wix-ui-icons-common/Search';
+
+import styles from './styles.scss';
+
+HTMLInputElement;
```
shlomitc, line 9: this can be removed

packages/wix-storybook-utils/src/ui/search-input.js
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classnames from 'classnames';
+import Input from './input';
+import Search from 'wix-ui-icons-common/Search';
```
shlomitc, line 5: this will cause a circular dependency, let's just copy the icon's SVG

packages/wix-storybook-utils/src/ui/styles.scss
```diff
@@ -54,13 +54,28 @@
   outline: 0;
 }
 
+.searchInputContainer {
+  position: relative;
+  margin: 30px 0;
```
shlomitc, line 59: I don't think this should be part of the component. the layout should be done by the consumer

packages/wix-storybook-utils/src/ui/search-input.js
gendeld: Hasn't so far, we use import from wix-ui-icons-common in 7 files, including similar 
./src/ui/dropdown.js
./src/ui/toggle-switch.js

shlomitc: maybe, but we didn't explicitly (or at all) used them from `wix-ui-icons-common` - https://github.com/wix/wix-ui/search?q=path%3Apackages%2Fwix-ui-icons-common%20wix-storybook-utils

In this case however, we're definitely going to consume the search-input directly, so that's an obvious circular dependency. I'd also prefer removing this dependency completely from `wix-storybook-utils`, just to make sure it won't be an issue in the future

shlomitc (approved):
just refactor the svg and we're good to merge

shlomitc (approved):
👍  just verify it looks visually fine and lets merge

PR #1936, opened 2020-09-21
move testId to constructor so each tooltip gets a different id

jonathanadler:


shlomitc (commented):
it looks legit to me.
@mykas - why is this wrapped with `isTestEnv` condition? why shouldn't it always be applied? If I'm thinking about e2e testing, then it will work only in jsdom but not in puppeteer

jonathanadler:
So what do yous say?
@mykas @shlomitc

shlomitc:
I think it's safe to merge, but let's consider this also not in test env

PR #1937, opened 2020-09-22
[wix-ui-icons-common] Add API tab to storybook docs

gendeld:


packages/wix-ui-icons-common/stories/APITab.tsx
```diff
@@ -0,0 +1,15 @@
+import { tab, description } from "wix-storybook-utils/Sections";
+
+const API_Table = `
+| Name | Type | Default Value | Required | Description |
+| --- | --- | --- | --- | --- |
+| className | string |  |   | Set custom class to svg root of icon |
+| size | string |  |   | Set the size of the icon |
```
shlomitc, line 7: We can also suggest passing “width” and “height”. Size prop put both of them for squares icons, but if for some reason different width and height are needed it can help

packages/wix-ui-icons-common/stories/APITab.tsx
```diff
@@ -0,0 +1,15 @@
...
+| style | object |  |   | Set style object to svg root of icon |
+| ***All other Props are passed to the SVG element*** | | | | |
+`;
+
+export default tab({
```
shlomitc, line 12: I think this should be part of the main story, this file should be only the markdown content

shlomitc (dismissed):
cool.
just FYI, in WSR we used in the past to create `md` files and then import them using `raw-loader`. in this case, I think it's good to keep it as `tsx` file since this will be imported without any loaders from the `node_modules`

PR #1945, opened 2020-10-12
<Dropdown /> - Add `dynamicWidth` and `appendTo` props

GuyPerkal:


jonathanadler (commented):
Looks great!
Can you please add a visual test for this

GuyPerkal:
What do you mean by visual test? I can only find the functionality tests in `Dropdown.spec.tsx`

jonathanadler:
> What do you mean by visual test? I can only find the functionality tests in `Dropdown.spec.tsx`

You can find the visual tests here: https://github.com/wix/wix-ui-tpa/blob/master/src/components/Dropdown/Dropdown.visual.tsx

You'll probably need to wrap the dropdown in some `div` in order to simulate a situation where the new props really take affect.

I suppose `dynamicWidth` should be tested.
Testing `appendTo` might be too much of an overkill

GuyPerkal:
But this PR is just for wix-ui-core, I figured I'd add it to wix-ui-tpa in a separate PR.

jonathanadler:
> But this PR is just for wix-ui-core, I figured I'd add it to wix-ui-tpa in a separate PR.

Oh so sorry, my bad 🤦‍♂️  I thought we're in wix-ui-tpa

EgoziE (approved):

packages/wix-ui-core/src/components/dropdown/Dropdown.tsx
```diff
@@ -56,6 +56,10 @@ export type DropdownProps = Pick<PopoverProps, 'fixed' | 'flip' | 'moveBy'> & {
   allowReselect?: boolean;
   /** Options box z-index */
   optionsContainerZIndex?: number;
+  /** Make width fit container */
+  dynamicWidth?: boolean;
+  /** Element to append the dropdown to */
+  appendTo?: AppendTo;
```
argshook, line 62: @jonathanadler  even though these props are just passed to Popover, to
avoid regression in future there should be a test checking that these props exist

jonathanadler: @argshook that's what I thought as well.
@GuyPerkal maybe you can add a simple test just to check that the props are passed, for future proof?

GuyPerkal: I'm not sure how to test it - couldn't figure out to get the Popover's properties using the Dropdown driver.

argshook: driver api is public api so it should be solid and not expose any
internals. For case like this we can use private driver, which is not
exposed to the users.

In there you can extend public driver and add any kind of methods to
check that prop exists

currently `Dropdown` does not have private driver so you would need to
create one.

an example of this can be found in `Popover.private.driver.ts`

https://github.com/wix/wix-ui/blob/master/packages/wix-ui-core/src/components/popover/Popover.private.driver.ts

then use this private driver in test instead of public one, for example
https://github.com/wix/wix-ui/blob/f73c1d7768b088f5bc3b2cb22b6b59ec1164bb7b/packages/wix-ui-core/src/components/popover/Popover.spec.tsx#L11

GuyPerkal: Sorry but I'm still not sure how to test it using the DOM API available in the drivers, I don't know how to check that a React prop was passed to the Popover the Dropdown uses.

argshook: don't check that prop was passed, check that expected behaviour happens
when prop is passed.

you can create `Dropdown.private.driver.ts`, which would extend `Popover.private.driver.ts`

then, in `Dropdown.spec.ts` write a test similar to this one in Popover

https://github.com/wix/wix-ui/blob/f73c1d7768b088f5bc3b2cb22b6b59ec1164bb7b/packages/wix-ui-core/src/components/popover/Popover.spec.tsx#L591-L610

maybe there's a test in Popover for `dynamicWidth`, which could also be
done from `Dropdown.spec.ts`

---

no need to do something extensive and complex. all what matters is that
future dev trying to remove `dynamicWidth` prop from `Dropdown` would be
stopped by a failing test

GuyPerkal: In order to test that dynamicWidth was passed to Popover I would need to spy on Popover's call to createModifiers, which seems wrong to me - not only does it mean testing implementation details, it's the implementation details of a component that has its own tests (Popover). I'm also not sure how to do it technically.

argshook: @GuyPerkal 
ok, this is too much effort, let's do something simple:

you can add this to `Dropdown.spec.tsx`

```
describe('props for Popover', () => {
  /*
  * Some props are passed down to `<Popover>` component used internally.
  * Behaviour of those props are tested in `<Popover>` implementation.
  * Thus, these tests just confirm that props are really passed
  */

  const propsAndValues = {
    placement: 'test',
    appendTo: 'test',
    dynamicWidth: true,
    showArrow: true,
    timeout: 0,
    id: 'test',
    flip: 'test',
    role: 'test',
  };

  Object.entries(propsAndValues).forEach(([prop, value]) => {
    it(`Popover should receive ${prop}=${value} prop`, () => {
      const mounted = mount(createDropdown({ [prop]: value }));
      expect(mounted.find('Popover').props()[prop]).toEqual(value);
    });
  });
});
```

this is a naive check but all i want is some code that makes sure future devs don't just remove some passed prop, so in here i think it's valid

GuyPerkal: Done!

argshook (approved):

argshook:
@GuyPerkal thanks for you patience :)

PR #1947, opened 2020-10-21
#FSM-3969|Fix typing bug when component is in rtl direction

zivpiz:


packages/wix-ui-core/src/components/signature-input/signing-pad/SigningPad.tsx
```diff
@@ -83,6 +85,9 @@ class SigningPadComp extends React.Component<SigningPadProps, SigningPadState> {
     onInit: PropTypes.func,
     /* Is the signature pad disabled */
     disabled: PropTypes.bool,
+    /* Is the signature pad direction rtl or ltr */
+    direction: PropTypes.any,
```
zivpiz, line 89: Tried to use  `direction: PropTypes.oneOf(['rtl', 'ltr']),` but it just treated it as string..
I can also just set direction to be `string` in both places. 
Tell me what you think

PR #1950, opened 2020-10-22
wix-ui-test-utils: feat(puppeteer): support `wrapper` argument

argshook:
This PR adds support for `wrapper` argument in `puppeteerUniTestkitFactoryCreator`.
Similar to how `wrapper` is supported in `enzymeUniTestkitFactoryCreator`

puppeteer unidriver usage before this PR:

```js
import { puppeteer } from 'puppeteer'
import { ButtonTestkit } from 'wix-style-react/dist/testkit/puppeteer'

const browser = await puppeteer.launch();
const page = await browser.newPage();

it('should work', async () => {
  const buttonTestkit = ButtonTestkit({
    page,
    dataHook: 'my-button'
  })
  expect(await buttonTestkit.exists()).toBe(true)
  // what if `page` has more than one button with `my-button` datahook? We can't test them easily :(
})
```

this would look for `[data-hook="my-button"]` and would use first occurence of element for the testkit.

if `page` has multiple elements with `my-button` data-hook, there's no way to test the rest of elements, user is always getting testkit for
first one (because unidriver internally does `page.$(selector)`)

---

this PR adds `wrapper` argument. continuing on example from above

```js
import { puppeteer } from 'puppeteer'
import { ButtonTestkit } from 'wix-style-react/dist/testkit/puppeteer'

const browser = await puppeteer.launch();
const page = await browser.newPage();

it('should work', async () => {
  const buttonTestkit = ButtonTestkit({
    page,
    dataHook: 'my-button',
    wrapper: await page.$('.some-wrapper-selector')
  })
  // testkit of a first `my-button` occurence within `.some-wrapper-selector`
  expect(await buttonTestkit.exists()).toBe(true)
})
```

### DIFF
--- a/packages/wix-ui-test-utils/src/puppeteer/puppeteer.ts
+++ b/packages/wix-ui-test-utils/src/puppeteer/puppeteer.ts
@@ -23,11 +23,19 @@ export function puppeteerUniTestkitFactoryCreator<T extends BaseUniDriver>(
     options: { dataHook: string }
   ) => T
 ) {
-  return async (obj: { dataHook: string; page: Page }) => {
-    const base = pupUniDriver({
-      page: obj.page,
-      selector: `[data-hook='${obj.dataHook}']`,
-    });
+  return async (obj: {
+    dataHook: string;
+    page: Page;
+    wrapper?: ElementHandle;
+  }) => {
+    const {wrapper: element} = obj; // destructuring `wrapper` so that type inference works well
+    const selector = `[data-hook='${obj.dataHook}']`;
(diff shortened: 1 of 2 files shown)


mykas (approved):
Looks good just try to add some unit test for the part that handles either page or wrapper logic. As we talked maybe extracting this logic to a separate function that would return arguments.

argshook:
@mykas i added two tests, please review if you have a moment

mykas:
@argshook if it's green it's green! let's merge this

PR #1951, opened 2020-10-25
#FSM-3969|Support rtl typing in SignatureInput

zivpiz:


NitayRabi:
Could you add tests (either visual or functional)?

NitayRabi (commented):
How does this affect A11y? maybe we should add `dir` ? does it also affect styling in any sort of way?

NitayRabi (approved):

PR #1953, opened 2020-10-26
(StorybookUtils): LiveCodeExample - set overflow to visible

mosheKerbel:
@argshook @shlomitc 

before:
Screen Shot 2020-10-26 at 9 19 57

after:
Screen Shot 2020-10-26 at 9 19 41

### DIFF
--- a/packages/wix-storybook-utils/src/LiveCodeExample/index.scss
+++ b/packages/wix-storybook-utils/src/LiveCodeExample/index.scss
@@ -209,7 +209,6 @@
 
 .preview {
   padding: 30px 20px;
-  overflow: auto;
   @include checkerboard(20px, #eff2f6, #fff);
 
   &.darkPreview {
--- a/packages/wix-storybook-utils/src/Playground/index.tsx
+++ b/packages/wix-storybook-utils/src/Playground/index.tsx
@@ -102,6 +102,7 @@ const Playground: React.FunctionComponent<Props> = ({
             editorCode: formatted,
           });
         }}
+        previewProps={{className: styles.overflowPreview}}
         {...rest}
       />
     ),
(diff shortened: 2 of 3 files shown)


shlomitc:
The reason for the overflow is to make the full-screen playground (like in WSR) to support scrolling.
In the inline code-snippets, bad things happen.

In WSR I just removed all these dashed code examples in favor of the plain background ones (also updated the generator).
Anyway, I think it's better to have only a single type of inline code examples.

argshook:
> The reason for the overflow is to make the full-screen playground (like in WSR) to support scrolling.
> In the inline code-snippets, bad things happen.

maybe `overflow: auto` should be applied only for playground, and all
the rest of code examples don't specify overflow at all (thus, the
default `overflow: visible` would work)?

mosheKerbel:
> The reason for the overflow is to make the full-screen playground (like in WSR) to support scrolling.
> In the inline code-snippets, bad things happen.
> 
> In WSR I just removed all these dashed code examples in favor of the plain background ones (also updated the generator).
> Anyway, I think it's better to have only a single type of inline code examples.

When we started documenting base-ui, @Zohar-Telor requested the bg-color to be transparent for components, so it's working that way for all pages in base-ui.

myabe we can do what Arijus suggested?

shlomitc:
yes, I'm good with this solution as well.

packages/wix-storybook-utils/src/LiveCodeExample/LiveCodeExample.tsx
```diff
@@ -37,6 +37,7 @@ export interface Props {
   autoRender?: boolean;
   darkBackground?: boolean;
   noBackground?: boolean;
+  isOverflow?: boolean;
```
argshook, line 40: @mosheKerbel i don't think you need a dedicated prop, just remove `overflow` css property entirely and add it to playground only

argshook (approved):
:+1:

PR #1957, opened 2020-11-03
[wix-ui-core] <Autocomplete /> fix option select

dtelega:
Can be reproduced in storybook.
Try type value and select some of the dropdown items(for example "value1"). It will not be selected
AR:
Screen Recording 2020-11-03 at 16 16 03 (1)

ER:
Screen Recording 2020-11-03 at 17 11 07

the same fix was done for addressInput component https://github.com/wix/wix-ui/commit/624c5ee6271837e8074eb0991db6c8dbe91e306d


dtelega:
1 eyes test is red because input still in focus after option selected

jonathanadler (commented):
We'll need to add a test for this
Otherwise someone might remove this in the future

jonathanadler (approved):

PR #1961, opened 2020-11-08
[wix-ui-core] `<DropdownContent/>` - Scroll to selected element on mount

GuyPerkal:
I'll add a test, but first I wanted to get the general concept approved, plus the usage of `scrollIntoView`. I could calculate the scroll position myself, but scrollIntoView is nicer to use and I think is supported by all the browser we support - https://caniuse.com/scrollintoview

### DIFF
--- a/packages/wix-ui-core/src/components/dropdown-content/DropdownContent.tsx
+++ b/packages/wix-ui-core/src/components/dropdown-content/DropdownContent.tsx
@@ -56,6 +56,15 @@ export class DropdownContent extends React.PureComponent<
 
   state = { hoveredIndex: NOT_HOVERED_INDEX };
 
+  componentDidMount() {
+    if (this.optionsContainerRef && this.props.selectedIds.length) {
+      const selectedIndex = this.props.options.findIndex((option) => option.id === this.props.selectedIds[0]);
+      const selectedOption = this.optionsContainerRef.childNodes[selectedIndex] as HTMLElement;
+      selectedOption.scrollIntoView({ block: 'center' });
+      this.setHoveredIndex(selectedIndex);
+    }
+  }
+
   setHoveredIndex(index: number) {
     if (this.state.hoveredIndex !== index) {
       this.setState(
--- a/packages/wix-ui-core/src/components/dropdown-content/DropdownContent.tsx
+++ b/packages/wix-ui-core/src/components/dropdown-content/DropdownContent.tsx
@@ -57,10 +57,25 @@ export class DropdownContent extends React.PureComponent<
   state = { hoveredIndex: NOT_HOVERED_INDEX };
 
   componentDidMount() {
-    if (this.optionsContainerRef && this.props.selectedIds.length) {
-      const selectedIndex = this.props.options.findIndex((option) => option.id === this.props.selectedIds[0]);
-      const selectedOption = this.optionsContainerRef.childNodes[selectedIndex] as HTMLElement;
-      selectedOption.scrollIntoView({ block: 'center' });
+    if (this.props.selectedIds.length) {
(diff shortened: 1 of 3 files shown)


GuyPerkal:
I'm also not sure whether we want this to be the default behaviour or if I should add a prop.

packages/wix-ui-core/src/components/dropdown-content/DropdownContent.tsx
```diff
@@ -56,6 +56,15 @@ export class DropdownContent extends React.PureComponent<
 
   state = { hoveredIndex: NOT_HOVERED_INDEX };
 
+  componentDidMount() {
+    if (this.optionsContainerRef && this.props.selectedIds.length) {
+      const selectedIndex = this.props.options.findIndex((option) => option.id === this.props.selectedIds[0]);
+      const selectedOption = this.optionsContainerRef.childNodes[selectedIndex] as HTMLElement;
+      selectedOption.scrollIntoView({ block: 'center' });
```
jonathanadler, line 63: Note that passing the scrollIntoView options isn't supported by all browsers and is considered experimental.
So it should be
`selectedOption.scrollIntoView();`
or
`selectedOption.scrollIntoView(true);`

See https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView#Parameters

jonathanadler (commented):
I'm not really sure what the use case for this is.
Should this really be the desired behaviour?

GuyPerkal:
It's for dropdowns with lots of options, to prevent the user from scrolling to get back to the option they chose before. Could be useful if they click on the option next to the one they meant to click. This is also the default behavior for HTML dropdowns.

packages/wix-ui-core/src/components/dropdown-content/DropdownContent.tsx
GuyPerkal: Actually this is so that the selected option will be in the middle of the options list after scroll, if the options are not supported they will just be ignored and the option will be at the top of the options list.

argshook: last time i used `scrollIntoView` i had different results in different browers. it was a while ago, so maybe now things are different, but i'd probably do something like `element.scrollTop = xxx`

shlomitc: in WSR we use this implementation: https://github.com/wix/wix-style-react/blob/d1a176172bca8ff0065dda20261fabea928b59e0/src/utils/scrollIntoView.js#L23

GuyPerkal:
Before I make any changes and add tests, is the general idea OK with everyone?

jonathanadler:
> Before I make any changes and add tests, is the general idea OK with everyone?

By the responses here I believe so.
Right @shlomitc @argshook ?

shlomitc:
> > Before I make any changes and add tests, is the general idea OK with everyone?
> 
> By the responses here I believe so.
> Right @shlomitc @argshook ?

yes, concept-wise that's cool

GuyPerkal:
@jonathanadler I no longer use scrolledIntoView, added tests.

GuyPerkal:
So it seems like it's not possible to test scrolling in Jest/JSDOM, I marked that test as ignored. It passes when running on Mocha.

GuyPerkal:
Any comments or can this be merged? For me the only current issue is the ignored test, I'm not sure if I should leave it ignored or remove it.

jonathanadler (approved):

GuyPerkal:
Do you have an estimation for when this will be merged?

PR #1964, opened 2020-11-11
infra - use esm official support for stylable files

shlomitc:
Using the best practices to support stylable and esmodules (see example [here](https://github.com/wix/stylable/tree/master/packages/webpack-plugin/test/e2e/projects/browser-field/node_modules/test-components))

1. it's backward compatible for all existing imports.
2. added esm js entries for the `hocs`. put it in a dedicated folder only with the relevant parts.
3. Due to an incorrect architecture of the css API of the `Popover` component, I created a `popover-proxy` entry which symlinks to the `Popover` folder and exposes its internal parts. I wish I could refactor it someday, but it's a big change so it can't be done in this PR. You can follow this comment for more details: https://github.com/wix/wix-style-react/pull/5445#discussion_r433881731.

## Thoughts:
The following change will require a change also in WSR. For example:
```diff
- import { Avatar as CoreAvatar } from 'wix-ui-core/dist/src/components/avatar';
+ import { Avatar as CoreAvatar } from 'wix-ui-core';
```
```diff
- -st-from: 'wix-ui-core/dist/src/components/avatar/avatar.st.css';
- -st-default: Avatar;
+ -st-from: 'wix-ui-core/index.st.css';
+ -st-named: Avatar;
```

Although it's cool to import from a single entry file, this will cause commonjs imports to become even longer since now this will import the entire `wix-ui-core` components.

**Suggestion:**
If symlink is a fine solution - I think we should create an entry per component. That way we can also maintain commonjs cherry-picking + esm imports between projects.

### DIFF
--- a/packages/wix-ui-core/hocs/package.json
+++ b/packages/wix-ui-core/hocs/package.json
@@ -1,0 +1,9 @@
+{
+  "name": "hocs",
+  "main": "../hocs.js",
+  "module": "../hocs.es.js",
+  "browser": {
+    "./index.st.css": "../hocs.es.st.css",
+    "./index.js": "../dist/es/src/index.js"
+  }
+}
--- a/packages/wix-ui-core/hocs/package.json
+++ b/packages/wix-ui-core/hocs/package.json
@@ -1,9 +1,9 @@
 {
   "name": "hocs",
-  "main": "../hocs.js",
-  "module": "../hocs.es.js",
+  "main": "../dist/src/hocs/index.js",
+  "module": "../dist/es/src/hocs/index.js",
   "browser": {
     "./index.st.css": "./index.es.st.css",
-    "./index.js": "../hocs.es.js"
+    "./index.js": "../dist/es/src/hocs/index.js"
   }
 }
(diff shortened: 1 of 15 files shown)


barak007 (commented):
look like a good start. I'm not sure about the popover-proxy but it can probably be fixed later.
I think we we can still generate the index.st.css with the cli as soon as it gets more complex.

packages/wix-ui-core/hocs/package.json
```diff
@@ -0,0 +1,9 @@
+{
+  "name": "hocs",
+  "main": "../dist/src/hocs/index.js",
+  "module": "../dist/es/src/hocs/index.js",
+  "browser": {
+    "./index.st.css": "./index.es.st.css",
```
yavorsky, line 6: `"stylableModule": "../dist/es/src/hocs/Focusable/Focusable.st.css"`
`"styleableMain": "../dist/src/hocs/Focusable/Focusable.st.css"`

---
"module": "../dist/es/src/hocs/index.js", -> ../dist/es/src/hocs/index.st.css

PR #1965, opened 2020-11-18
[wix-ui-icons-common] Expose icon declaration files to consumers

gendeld:


shlomitc (approved):

dht:
@gendeld @shlomitc 
I think this PR effecting [wix-ui-icons-commons@2.0.294](https://www.npmjs.com/package/wix-ui-icons-common/v/2.0.294) is breaking our `my-account-header` build with:
```
AccountMenuTeamTrigger.tsx(33,49): error TS2322: Type 'number' is not assignable to type 'string'
```
When I dived into the `.d.ts` definition files I saw:
```
export interface DropDownArrowProps extends React.SVGAttributes<SVGElement> {
    size?: string;   <===== string instead of number
}
declare const DropDownArrow: React.SFC<DropDownArrowProps>;
export default DropDownArrow;
```

gendeld:
@dht 
Icons have always defined size as string, but up until recently the types weren't exposed.
`wix-ui-icons-common` depends on `svg2react-icon` which creates the react icon components with the type.
Recently in version 3.1.175 we updated the generated types to also support number.
You can check what version `svg2react-icon` is installed by running:
```
npx qnm --why svg2react-icon
```
Reinstalling node_modules should give you the latest version available if there aren't conflicting versions elsewhere in your dependency graph.

dht:
@gendeld 
I see, well our team probably didn't know a string is expected :-)
It is mentioned in the docs that a string is needed, of course.
Anyway, it seems we are in good company and many teams are passing a number:
https://bo.wix.com/columbus/grep?query=ChevronDown+size%3D{&extension=&case=false&algorithm=codesearch
It will probably break each one of those builds, wouldn't it?

I was of course referring to the CI build not the local environment:
http://tc.dev.wixpress.com/viewLog.html?buildId=121284950&buildTypeId=MyAccountHeader_HeaderClient_O
with wix-ui-icons-common on latest

Basically, if I understand correctly, until 2.0.294 it gracefully accepted an invalid type and now because the `d.ts` files are exposed it fails the typescript compilation.

Perhaps we, as in our team, would wait for the next version to support numbers and stick with 2.0.293 for the time being.

gendeld:
@dht Not exactly a wrong type, just not what was expected, we changed it to be more loose to match the React.SVGElement spec.
Since we define the dependency as
```
"svg2react-icon": "^3.0.0",
```
it should install  v3.1.175 with the looser typing when available.
It's weird that it didn't work on your end, I'd be interested to know what 
```
npx qnm --why svg2react-icon
```
prints for you

shlomitc:
@gendeld - `svg2react-icon` is a build tool, so it will not be a dependency of `wix-ui-icons-common`. it generated the output, so I think in this case simply releasing a new version of `wix-ui-icons-common` will do the trick

gendeld:
@shlomitc yes of course 😅 don't know why I didn't connect those dots

PR #1970, opened 2020-11-24
add isRtl to scope in LiveCodeExample

jonathanadler:
So that the content in the live example can use this if needed


argshook (approved):

jonathanadler:
Thanks!

PR #1971, opened 2020-11-24
fix live code example setState issue (after rtl in scope change)

jonathanadler:
dont set state if not needed + default scope to empty object


argshook (approved):

jonathanadler:
Thanks!

PR #1976, opened 2020-12-02
Use correct string format for Stylable for ButtonStyle

ronami:
### Summary

As part of the work on Yoshi v5, we're bumping Stylable to v4. Stylable v4, checks that the CSS it outputs is valid CSS and fails otherwise. The CSS output of the changed file includes invalid CSS syntax which causes it to fail in the browser (I don't have a real site use-case). With Stylable v4, it fails in build time and prevents us from testing Yoshi v5 on more projects.

This PR fixes it by having a single string for the entire font-family variable. This fixes the build time issues with Yoshi v5 and should also fix a production bug we currently have.

cc @barak007

### DIFF
--- a/packages/wix-ui-core/src/components/deprecated/button/ButtonStyle.st.css
+++ b/packages/wix-ui-core/src/components/deprecated/button/ButtonStyle.st.css
@@ -13,7 +13,7 @@
     padding: 0px 23px;
     borderRadius: 0;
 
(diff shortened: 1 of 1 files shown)


barak007:
#### Explanation: 
Stylable `:vars` that wrapped with quotes are treated like "content as is". When we apply the `value()` function we strip the quotes from both ends.
In previous version of stylable webpack we didn't know that the output css was broken. In the new version we moved the general optimization to work form the target css so we parse each css after the transform and we can report if it's broken.

PR #1978, opened 2020-12-07
wix-ui-core: fix stylable patch script

argshook:
previously only first task would be run, because promise would get
resolved before it finishes it's job:

```diff
-const appendFile = util.promisify(fs.appendFileSync);
+const appendFile = util.promisify(fs.appendFile);
```

also, wrapping `addPopoverInternalParts` with an `if`:

```js
if(entryFile) {
  return addPopoverInternalParts({entryFile})
}
```

because there are tasks that don't pass `entryFile` which results in
`undefined.es.st.css` file creation

cc @tomrav

### DIFF
--- a/packages/wix-ui-core/hocs.es.st.css
+++ b/packages/wix-ui-core/hocs.es.st.css
@@ -7,7 +7,3 @@
 .root EllipsedTooltip{}
 :import {-st-from: "./dist/es/src/hocs/EllipsedTooltip/Text.st.css";-st-default:Text;}
 .root Text{}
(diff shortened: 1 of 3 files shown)


packages/wix-ui-core/hocs.es.st.css
```diff
@@ -7,3 +7,7 @@
 .root EllipsedTooltip{}
 :import {-st-from: "./dist/es/src/hocs/EllipsedTooltip/Text.st.css";-st-default:Text;}
 .root Text{}
+:import {-st-from: "./dist/es/src/components/popover/Popover.st.css";-st-named: arrow as Popover_arrow, withArrow as Popover_withArrow, popoverContent as Popover_popoverContent;}
+.root .Popover_arrow {}
+.root .Popover_withArrow {}
+.root .Popover_popoverContent {}
```
shlomitc, line 13: This should appear only on the `index.st.css` file, not in the hocs one

PR #1979, opened 2020-12-13
feat(DropdownContent) - Check the selected option exists before attem…

GuyPerkal:
…pting to scroll to it

I needed to add this check because of the AddressInput, it doesn't reset its selected id when its options change - that's probably not the expected behavior but I didn't want to mess with it.


jonathanadler (approved):

jonathanadler:
Let's merge this tomorrow

PR #1986, opened 2020-12-17
<RadioButton/> - added unidriver and id as a prop

Zivilekin:


packages/wix-ui-core/src/utils/generateDataAttr.ts
```diff
@@ -0,0 +1,13 @@
+export const generateDataAttr = (
+  props: Record<string, unknown>,
+  filter?: string[]
```
mykas, line 3: I don't think this should be optional. Function without second parameter is useless 🤷‍♂️

mykas (approved):

PR #1998, opened 2020-12-28
feat(wix-ui): onFocusByKeyboard

urig-wix:
add onFocusByKeyboard prop


packages/wix-ui-core/src/components/radio-button/RadioButton.e2e.ts
```diff
@@ -26,4 +26,19 @@ describe('RadioButton', () => {
       },
     );
   });
+
+  eyes.it('RadioButton onFocusByKeyboard works', async () => {
```
jonathanadler, line 30: In this case I think we can use a regular, `it` function, instead of `eyes.it`.
The difference is that `eyes.it` takes a screenshot, and `it` is a regular protractor test

jonathanadler (approved):
Awesome 💪 

I'm approving, but there's just left one small comment to fix.

jonathanadler (approved):

jonathanadler (approved):

PR #2000, opened 2021-01-03
feat(Image) - expose `className` prop

nitayneeman:
This PR exposes `className` props for `Image` & `MediaImage`.

See this:
https://github.com/wix/wix-ui-tpa/pull/567/files#r550477196

### DIFF
--- a/packages/wix-ui-core/src/components/image/image.driver.private.ts
+++ b/packages/wix-ui-core/src/components/image/image.driver.private.ts
@@ -9,6 +9,7 @@ import {
 export interface ImageDriver extends ImagePublicDriver {
   getResizeMode(): Promise<string | boolean>;
   getSrcSet(): Promise<string>;
+  hasClass(className: string): Promise<boolean>;
 }
 
 export const imageDriverFactory = (base: UniDriver): ImageDriver => {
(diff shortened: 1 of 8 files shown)


packages/wix-ui-core/src/components/image/image.driver.private.ts
```diff
@@ -9,6 +9,7 @@ import {
 export interface ImageDriver extends ImagePublicDriver {
   getResizeMode(): Promise<string | boolean>;
   getSrcSet(): Promise<string>;
+  hasClass(className: string): Promise<boolean>;
```
jonathanadler, line 12: Do we want to expose this method?

jonathanadler (approved):

packages/wix-ui-core/src/components/image/image.driver.private.ts
nitayneeman: I don't think so, at least for now (unless a consumer will ask it), since it's mainly for internal tests... That's why I placed it in the private driver.

jonathanadler: Oh sorry, I thought this was the public driver.
👍

jonathanadler (dismissed):

shlomitc (approved):

PR #2004, opened 2021-01-05
feat(MediaImage) - pass dimensions to the native image

nitayneeman:
This PR passes the container dimensions to the native image as attributes, if they are provided.

The thing is today we can pass the container (or target) dimensions:
```jsx
const containerWidth = 300;
const containerHeight = 250;
const sourceWidth = 480;
const sourceHeight = 360;

// ...

<MediaImage
  width={containerWidth}
  height={containerHeight}
  mediaPlatformItem={{
    uri: 'c5f754_dd75514d14fa4057b4f4a6cc8ce7add3~mv2.jpg',
    width: sourceWidth,
    height: sourceHeight,
  }}
/>
```

And then we'll get:
image

Basically the image itself might be larger according to the media platform service's calculation, but the container (IMO) should be resized as requested (300x250). In other words, I think that `width` & `height` on the direct component (not those of mediaPlatformItem) should behave like `width` & `height` of a regular native `img` element. In this way, we render the dimensions that we actually requested and not different sizes (even though the resource itself might arrive larger depends on the media service)...

So after the change, the same code above will produce:
image

Besides that it's more correct (you get the container/target you request), it also has a matter of [performance](https://www.smashingmagazine.com/2020/03/setting-height-width-images-important-again/).

The only concern is this usage:
https://github.com/wix-private/scheduler/blob/master/statics/viewer/bookings-service-details-widget/src/components/BookingServicePage/Widget/Header/Header.tsx#L59-L60

It might affect since in the past they render the dimensions which return and calculated by media service as are, and now the image would be restricted by `headerDimensions` (according to code).  @liorMar - WDYT? If it's problematic, we can pass `nativeProps` prop or something like that, but I think this change is correct in general...

### DIFF
--- a/packages/wix-ui-core/src/components/media-image/media-image.browser.spec.tsx
+++ b/packages/wix-ui-core/src/components/media-image/media-image.browser.spec.tsx
@@ -105,6 +105,15 @@ describe('MediaImage', () => {
     expect(await mediaImageDriver.getSrc()).toEqual(SRC);
   });
 
+  it('should set the container dimensions as attributes when provided', async () => {
+    const mediaImageDriver = await createDriver(
+      <MediaImage mediaPlatformItem={mediaPlatformItem} width={WIDTH} height={HEIGHT} />,
+    );
+
+    expect(await mediaImageDriver.getWidthAttribute()).toEqual(WIDTH);
+    expect(await mediaImageDriver.getWidthAttribute()).toEqual(HEIGHT);
+  });
+
   it('displays an alt prop', async () => {
     imageClientSDK.getScaleToFillImageURL.mockReturnValue(BROKEN_SRC);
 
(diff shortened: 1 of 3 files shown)


jonathanadler:
Looks good, though like you've noticed, it might break existing apps.
Is booking's app the only one using this?

nitayneeman:
Well, it looks the same - before:
image

After:
image

liorMar (approved):
Looks good

nitayneeman:
@jonathanadler These are the usages I could find:
https://github.com/search?q=org%3Awix+wix-ui-core+mediaimage&type=code
https://github.com/search?q=org%3Awix+wix-ui-core+mediaimage&type=code

Only Bookings and Restaurants as...

@shlomitc do you see any concern regarding merging this PR?

nitayneeman:
@jonathanadler can you approve? let's merge it? 🙏

jonathanadler (approved):

PR #2011, opened 2021-01-21
#FSM-4334|AddressInput - Support Atlas Service

zivpiz:
Instead of googleMaps.
provider and new client will be sent from wix-ui-santa.


packages/wix-ui-core/src/clients/Atlas/AtlasBasicClient.ts
```diff
@@ -0,0 +1,112 @@
...
+import {
+  Address,
+} from './types';
+import { MapsClient } from '../GoogleMaps/types'
+import { Omit } from 'type-zoo'
+import { metaSiteInstaceMock } from './testUtils'
+
+const ATLAS_WEB_BASE_URL = '/api/wix-atlas-service-web';
+const BASE_LINGUIST_HEADER =
+  '|en-us|false|4e8a573a-6b3e-426f-9d2f-5285b7dc90f8';
```
NitayRabi, line 16: 4e8a573a-6b3e-426f-9d2f-5285b7dc90f8 ?

packages/wix-ui-core/src/clients/Atlas/AtlasBasicClient.ts
```diff
@@ -0,0 +1,112 @@
...
+  }))
+
+const toSuggestions = (predictions: Array<V2Prediction>): Array<Address> =>
+  predictions.map(prediction => ({
+    place_id: prediction.searchId || '',
+    description: prediction.description || '',
+    types: []
+  }));
+
+export class AtlasBasicClient implements Omit<MapsClient, 'placeDetails' | 'useClientId'> {
```
NitayRabi, line 44: Maybe Instead of `Omit` we should support a BaseClient, and GoogleClient which has `placeDetails` & `useClientId ` ?

packages/wix-ui-core/src/clients/Atlas/AtlasBasicClient.ts
```diff
@@ -0,0 +1,112 @@
...
+      }).predict;
+    }
+
+    if (!this._getPlace) {
+      this._getPlace = PlacesServiceV2()({
+        'Authorization': instance,
+        'x-wix-linguist': `${lang}${BASE_LINGUIST_HEADER}`,
+      }).getPlace;
+    }
+  }
```
NitayRabi, line 62: Why do we need the `_initServices` approach? does it manipulate the DOM? adds some script tag?

NitayRabi (approved):
LGTM
- So consumers of these new clients should opt-in to replace their GoogleBasicClient right?

packages/wix-ui-core/src/clients/Atlas/AtlasBasicClient.ts
zivpiz: We just need a 'valid' value there, it can be constant (verified with platformization team)

packages/wix-ui-core/src/clients/Atlas/AtlasBasicClient.ts
zivpiz: Sounds good, I'll change

packages/wix-ui-core/src/clients/Atlas/AtlasBasicClient.ts
zivpiz: Not really, no. just to save work and not re-define the headers every time.
Should I change it?

NitayRabi: So why not do it once in the constructor? or create a "memoized" getter for the methods?

zivpiz:
`So consumers of these new clients should opt-in to replace their GoogleBasicClient right?`
Basically yes, restaurants will still have the GoogleBasicClient (If they still use this component), and the viewer component will start using Atlas based on an experiment we'll be opening (Logic can be found here - https://github.com/wix-private/wix-ui-santa/pull/1812/files#diff-2dcad2816fd6690b0c3e33413ddb472f2047249068245d57019098a64087da7cR132)

shlomitc:
@gendeld - if you have any inputs on this implementation based on your experience with Atlas that would be great

packages/wix-ui-core/src/clients/Atlas/AtlasBasicClient.ts
zivpiz: Changed 👍

packages/wix-ui-core/src/clients/Atlas/AtlasBasicClient.ts
```diff
@@ -0,0 +1,113 @@
...
+      }).predict;
+    }
+    return this._predictWithHeaders(request)
+  }
+
+  private _getPlace(request: any, lang: string, instance: string) {
+    if (!this._getPlaceWithHeaders) {
+      this._getPlaceWithHeaders = PlacesServiceV2()({
+        'Authorization': instance,
+        'x-wix-linguist': `${lang}${BASE_LINGUIST_HEADER}`,
```
gendeld, line 60: Are we sure this is needed? As long as there is a `wixLanguage` field in the cookie, Atlas provides predictions in the appropriate language

packages/wix-ui-core/src/clients/Atlas/AtlasBasicClient.ts
gendeld: You can use the app's instance as the last part of the field, rather than using a random BO instance

packages/wix-ui-core/src/clients/Atlas/AtlasBasicClient.ts
```diff
@@ -0,0 +1,113 @@
...
+  V2GetPlaceResponse,
+  V2Prediction,
+  WixAtlasServiceWeb,
+} from '@wix/ambassador-wix-atlas-service-web/http'
+import {
+  Address,
+} from './types';
+import { BaseMapsClient } from '../GoogleMaps/types'
+
+const ATLAS_WEB_BASE_URL = '/api/wix-atlas-service-web';
```
gendeld, line 12: I've used '/wix-atlas-service-web', do we know what is the difference? are these the same? or is this right for the viewer?

packages/wix-ui-core/src/clients/Atlas/AtlasBasicClient.ts
zivpiz: Yep, it wouldn't work in the viewer if I didn't provide this header.
This is the solution that was discussed with platformization team

packages/wix-ui-core/src/clients/Atlas/AtlasBasicClient.ts
zivpiz: Not sure what's the difference to be honest, but it looks like it's the right url for my usecase 🤔

packages/wix-ui-core/src/clients/Atlas/AtlasBasicClient.ts
zivpiz: The same instance I use for Auth?

packages/wix-ui-core/src/clients/Atlas/AtlasBasicClient.ts
```diff
@@ -0,0 +1,113 @@
...
+const BASE_LINGUIST_HEADER =
+  '|en-us|false|4e8a573a-6b3e-426f-9d2f-5285b7dc90f8';
+const { AutocompleteServiceV2, PlacesServiceV2 } = WixAtlasServiceWeb(ATLAS_WEB_BASE_URL);
+
+const serializeResult = (results: CommonAddress[]) =>
```
gendeld, line 17:
```suggestion
const serializeGeocodeResult = (results: CommonAddress[]) =>
```
Maybe better to differentiate so it's easier to understand we're doing something different here.

packages/wix-ui-core/src/clients/Atlas/AtlasBasicClient.ts
```diff
@@ -0,0 +1,113 @@
...
+  results.map(atlasResponse => ({
+    formatted: atlasResponse.formattedAddress,
+    streetAddress: atlasResponse.streetAddress,
+    subdivision: atlasResponse.subdivision,
+    city: atlasResponse.city,
+    country: atlasResponse.country,
+    postalCode: atlasResponse.postalCode,
+    ...(atlasResponse.geocode
+      ? {
+        location: {
```
gendeld, line 27: https://en.wikipedia.org/wiki/Null_Island
0, 0 is a place on the map, maybe it's not a good default location

packages/wix-ui-core/src/clients/Atlas/AtlasBasicClient.ts
```diff
@@ -0,0 +1,113 @@
...
+    description: prediction.description || '',
+    types: []
+  }));
+
+export class AtlasBasicClient implements BaseMapsClient {
+  private _predictWithHeaders;
+  private _getPlaceWithHeaders;
+
+  private _predict(request: any, lang: string, instance: string) {
+    if (!this._predictWithHeaders) {
```
gendeld, line 47: I think it would look cleaner if this was inside the constructor function.

packages/wix-ui-core/src/clients/Atlas/AtlasBasicClient.ts
```diff
@@ -0,0 +1,113 @@
...
+      this._predictWithHeaders = AutocompleteServiceV2()({
+        'Authorization': instance,
+        'x-wix-linguist': `${lang}${BASE_LINGUIST_HEADER}`,
+      }).predict;
+    }
+    return this._predictWithHeaders(request)
+  }
+
+  private _getPlace(request: any, lang: string, instance: string) {
+    if (!this._getPlaceWithHeaders) {
```
gendeld, line 57: I think it would look cleaner if this was inside the constructor function.

packages/wix-ui-core/src/clients/Atlas/AtlasBasicClient.ts
```diff
@@ -0,0 +1,113 @@
...
+    place_id: prediction.searchId || '',
+    description: prediction.description || '',
+    types: []
+  }));
+
+export class AtlasBasicClient implements BaseMapsClient {
+  private _predictWithHeaders;
+  private _getPlaceWithHeaders;
+
+  private _predict(request: any, lang: string, instance: string) {
```
gendeld, line 46:
```suggestion
  private _predict(request: PredictRequest, lang: string, instance: string) {
```

packages/wix-ui-core/src/clients/Atlas/AtlasBasicClient.ts
```diff
@@ -0,0 +1,113 @@
...
+    if (!this._predictWithHeaders) {
+      this._predictWithHeaders = AutocompleteServiceV2()({
+        'Authorization': instance,
+        'x-wix-linguist': `${lang}${BASE_LINGUIST_HEADER}`,
+      }).predict;
+    }
+    return this._predictWithHeaders(request)
+  }
+
+  private _getPlace(request: any, lang: string, instance: string) {
```
gendeld, line 56:
```suggestion
  private _getPlace(request: V2GetPlaceRequest, lang: string, instance: string) {
```

packages/wix-ui-core/src/clients/Atlas/types.ts
```diff
@@ -0,0 +1,5 @@
+export interface Address {
```
gendeld, line 1:
```suggestion
export interface Suggestion {
```

packages/wix-ui-core/src/clients/Atlas/AtlasBasicClient.ts
```diff
@@ -0,0 +1,113 @@
...
+      }).getPlace;
+    }
+
+    return this._getPlaceWithHeaders(request)
+  }
+
+  async autocomplete(
+    clientId: string,
+    lang: string,
+    request: any,
```
gendeld, line 70: Create a ClientRequest type

packages/wix-ui-core/src/clients/Atlas/AtlasBasicClient.ts
```diff
@@ -0,0 +1,113 @@
...
+    instance: string
+  ): Promise<Address[]> {
+
+    const predictRequest = {
+      input: typeof request === 'string' ? request: request.input,
```
gendeld, line 75:
```suggestion
      input: typeof request === 'string' ? request : request.input,
```

gendeld: How would it be string? You're generating it in the `createAutocompleteRequest` function no?

packages/wix-ui-core/src/clients/Atlas/AtlasBasicClient.ts
```diff
@@ -0,0 +1,113 @@
...
+        ? {
+          countryCodes: [request.componentRestrictions.country],
+        }
+        : {}),
+
+    }
+    try {
+      const predictResponse = await this._predict(predictRequest, lang, instance)
+      if (predictResponse.predictions && predictResponse.predictions.length) {
+        return toSuggestions(predictResponse.predictions || []);
```
gendeld, line 86:
```suggestion
        return toSuggestions(predictResponse.predictions);
```

packages/wix-ui-core/src/clients/Atlas/AtlasBasicClient.ts
```diff
@@ -0,0 +1,113 @@
...
+      }
+      return Promise.reject('ZERO_RESULTS');
+    }
+    catch (e) {
+      return Promise.reject(e)
```
gendeld, line 91: Why catch an error to throw it back again?

packages/wix-ui-core/src/clients/Atlas/AtlasBasicClient.ts
```diff
@@ -0,0 +1,113 @@
...
+    }
+  }
+
+  async geocode(
+    clientId: string,
+    lang: string,
+    request: any,
```
gendeld, line 98: Add ClientRequest type

packages/wix-ui-core/src/clients/Atlas/AtlasBasicClient.ts
zivpiz: No real reason, changing

packages/wix-ui-core/src/clients/Atlas/AtlasBasicClient.ts
```diff
@@ -0,0 +1,113 @@
...
+    instance: string
+  ) {
+    const getPlaceRequest = {
+      searchId: request.placeId
+    }
+
+    try {
+      const result: V2GetPlaceResponse = await this._getPlace(getPlaceRequest, lang, instance);
+      return serializeResult([result?.place?.address] || [{}]) as any;
```
gendeld, line 107: Why all this casting?
If result can be empty (?) check for it and return an error
otherwise return mapped result

packages/wix-ui-core/src/components/address-input/AddressInput.tsx
```diff
@@ -41,6 +42,8 @@ export type AddressInputProps = Pick<
   'data-hook'?: string;
   /** Maps client, should implement autocomplete, geocode and placeDetails methods */
   Client: MapsClientConstructor;
+  /** client identifier - GoogleMaps or Wix's Atlas */
```
gendeld, line 45:
```suggestion
  /** Which location provider you want to use, can be either 'google' or 'atlas' (default : 'google') */
```

packages/wix-ui-core/src/clients/Atlas/AtlasBasicClient.ts
zivpiz: The casting is because `geocode` expects a return type which is a bit different, that's why I have different functions in AddressInput.tsx as well
But I'll return an error if no address is found

packages/wix-ui-core/src/components/address-input/AddressInput.tsx
```diff
@@ -41,6 +42,8 @@ export type AddressInputProps = Pick<
...
+  providerName?: string;
```
gendeld, line 46:
```suggestion
  providerName?: 'google' | 'atlas';
```
You should also provide a default value for this in defaultProps.
I suggest using  'google' as default, because in order to use 'atlas' you must pass `instance` prop.
So using Atlas by default would be a breaking change for people already using the component

zivpiz: I agree, adding

packages/wix-ui-core/src/clients/Atlas/AtlasBasicClient.ts
```diff
@@ -0,0 +1,96 @@
...
+      ...(typeof request !== 'string' && request.componentRestrictions && request.componentRestrictions.country
+        ? {
+          countryCodes: [request.componentRestrictions.country],
+        }
+        : {}),
+
+    }
+    const predictResponse = await this._predict(predictRequest)
+    if (predictResponse.predictions && predictResponse.predictions.length) {
+      return toSuggestions(predictResponse.predictions || []);
```
gendeld, line 76:
```suggestion
      return toSuggestions(predictResponse.predictions);
```
You are already checking it's a valid array in the previous line

packages/wix-ui-core/src/clients/Atlas/AtlasBasicClient.ts
```diff
@@ -0,0 +1,96 @@
...
+  }
+
+  async autocomplete(
+    clientId: string,
+    lang: string,
+    request: ClientAutocompleteRequest,
+  ): Promise<Suggestion[]> {
+
+    const predictRequest = {
+      input: typeof request === 'string' ? request : request.input,
```
gendeld, line 66: How can request be string?

packages/wix-ui-core/__mocks__/@wix/ambassador-wix-atlas-service-web/http/index.ts
```diff
@@ -0,0 +1,50 @@
...
+						}
+            return getPlaceSucceeded ? Promise.resolve(successfulGetPlaceResponseMock) : Promise.reject('GetPlaceFailed')
+					}),
+				}
+			})
+		),
+	}
+}
+
+export { WixAtlasServiceWeb }
```
gendeld, line 50: Linting?

packages/wix-ui-core/package.json
```diff
@@ -83,6 +83,7 @@
   },
   "dependencies": {
     "@loadable/component": "^5.14.1",
+    "@wix/ambassador-wix-atlas-service-web": "^1.0.148",
```
gendeld, line 86: Does `wix-ui-core` have any non-wix consumers? If so, this line will break their builds

packages/wix-ui-core/src/clients/Atlas/AtlasBasicClient.spec.ts
```diff
@@ -0,0 +1,106 @@
...
+          country: 'some-other country'
+        }
+      }
+      try {
+        result = await client.autocomplete(clientId, lang, predictRequest, metaSiteInstaceMock)
+      }
+      catch (e) {
+        result = e
+      }
+      expect(result).toEqual('PredictFailed')
```
gendeld, line 54: Why are we testing the failed result here?

packages/wix-ui-core/src/clients/Atlas/AtlasBasicClient.ts
```diff
@@ -0,0 +1,96 @@
...
+    lang: string,
+    request: ClientGeocodeRequest,
+  ) {
+    const getPlaceRequest = {
+      searchId: request.placeId
+    }
+
+    const result: V2GetPlaceResponse = await this._getPlace(getPlaceRequest);
+    if (result?.place?.address) {
+      return serializeGeocodeResult([result.place.address]) as any;
```
gendeld, line 92: I think it's bad to cast as any
TypeScript is a contract with the library's consumer.
If we are saying `geocode` returns type X, we should make `serializeGeocodeResult` return X.
If this is not possible, we should change the type. 
Otherwise, we are just creating possible errors on our users' sites

packages/wix-ui-core/package.json
zivpiz: @shlomitc Do you maybe know?

packages/wix-ui-core/src/clients/Atlas/AtlasBasicClient.ts
zivpiz: Not sure but I saw the GoogleMaps supports it - https://github.com/wix/wix-ui/blob/master/packages/wix-ui-core/src/clients/GoogleMaps/GoogleMapsBasicClient.ts#L90
Didnt want to miss anything

packages/wix-ui-core/package.json
shlomitc: `wix-ui-core` should be publically installable, meaning `npm i wix-ui-core` should work from `npmjs`. If you introduce private dependencies, it will fail to install.

We had a similar problem in `wix-style-react` and we ended up with some compromises:
1. all atlas dependencies are `peerDependencies` and `devDependencies` while the consuming project installs it as well (either as devdep or regular one). this means you can still install `wix-ui-core`.
2. In order not to fail the build (typescript), the `index` file should not export this adapter. I don't think you're exposing the adapter from there, so that's fine, but need to be aware of it.

shlomitc: Check the doc page of `AtlasAddresInput` in WSR
https://www.wix-style-react.com/?activeTab=Description&path=%2Fstory%2Fcomponents-api-work-in-progress--atlasaddressinput

zivpiz: I see, thanks!

@gendeld Do I have to install `@wix/ambassador` to get atlas to work though? (I see in the docs that `@wix/ambassador` should be installed)
I had it here for a while but I deleted it, I wonder if it should've failed.

packages/wix-ui-core/src/clients/Atlas/AtlasBasicClient.spec.ts
zivpiz: It's a bit weird but I was having hard time mocking and testing it, in my mock I failed the request if country exists and isn't IL, so in this test I was checking that we actually send the non-IL country.
I've added a comment to the test

packages/wix-ui-core/src/clients/Atlas/AtlasBasicClient.ts
zivpiz: I don't really want to have `serializeGeocodeResult` return X because it will be a lot of work to manipulate the date just to manipulate it back in the component itself.
I can change the type to be all optional values, is that OK by you?

packages/wix-ui-core/src/clients/Atlas/AtlasBasicClient.spec.ts
gendeld: But does it test this statement: 'should set the country filter to atlas autocomplete request'?

gendeld (dismissed):

gendeld (approved):

PR #2016, opened 2021-01-27
remove role=radio and opacity from input in radiobutton component

jonathanadler:
According to review from @NeilWix


jonathanadler:
@mykas @EgoziE 
I was afraid that removing the `role` attribute might cause someone to break, so I added a prop for this.
I wasn't sure how to call the prop actually.
We can change it, I'm not married to it 😄

EgoziE (approved):
Fine...

PR #2019, opened 2021-01-28
Guild Week Task - Classic Editor Icon Library

zivpiz:


shlomitc:
for some reason the CI build is not displayed here but it is failing for visual tests diff
https://pullrequest-tc.dev.wixpress.com/viewLog.html?buildId=15098697&buildTypeId=FedInfra_Package_WixUiIconsCommon&tab=buildLog

Please adjust also `allIcons.e2e.js` to point to the right files (If you can refactor it to import a constant variable that could be great)

shlomitc:
please also notice the `create-aliases` script in `package.json`. This is the part that is responsible for exporting the icons so they can be consumed from the root

mosheKerbel:
I accepted the diffs in AppliTools, rerunning build now.

mosheKerbel (approved):

PR #2027, opened 2021-02-02
<Dropdown/> reset hover state when on mouse leave

jjaskulis:
Currently dropdown options remains hovered even if the mouse is no longer hovered on option.
The hovered state should disappear if the mouse is out of dropdown component bounds.
This PR add the mouse leave handler which resets the hovered state.

### DIFF
--- a/packages/wix-ui-core/src/components/dropdown-content/DropdownContent.spec.tsx
+++ b/packages/wix-ui-core/src/components/dropdown-content/DropdownContent.spec.tsx
@@ -45,7 +45,7 @@ describe('DropdownContent', () => {
     it('should be hovered on mouse enter and not on mouse leave', () => {
       const driver = createDriver(createDropdownContent({ options }));
       driver.optionAt(0).mouseEnter();
-      expect(driver.optionAt(0).isHovered()).toBeTruthy();
+      expect(driver.optionAt(0).isHovered()).toBe(true);;
       driver.optionAt(0).mouseLeave();
       expect(driver.optionAt(0).isHovered()).toBeFalsy();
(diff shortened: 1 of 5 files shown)


packages/wix-ui-core/src/components/dropdown-content/DropdownContent.spec.tsx
```diff
@@ -42,6 +42,14 @@ describe('DropdownContent', () => {
       expect(onOptionClick).toHaveBeenCalledWith(options[4]);
     });
 
+    it('should be hovered on mouse enter and not on mouse leave', () => {
+      const driver = createDriver(createDropdownContent({ options }));
+      driver.optionAt(0).mouseEnter();
+      expect(driver.optionAt(0).isHovered()).toBeTruthy();
```
sivanharel, line 48:
```suggestion
      expect(driver.optionAt(0).isHovered()).toBe(true);;
```

packages/wix-ui-core/src/components/dropdown-content/DropdownContent.spec.tsx
```diff
@@ -42,6 +42,14 @@ describe('DropdownContent', () => {
...
+      driver.optionAt(0).mouseLeave();
+      expect(driver.optionAt(0).isHovered()).toBeFalsy();
```
sivanharel, line 50:
```suggestion
      expect(driver.optionAt(0).isHovered()).toBe(false);
```

sivanharel (approved):
Awesome!

sivanharel (approved):

PR #2031, opened 2021-02-05
<Avatar /> - remove grapheme-splitter dep

jjaskulis:
Removing not needed feature and dependency to decrease bundle size.

https://jira.wixpress.com/browse/DSM-1062

### DIFF
--- a/packages/wix-ui-core/package.json
+++ b/packages/wix-ui-core/package.json
@@ -85,7 +85,6 @@
     "@loadable/component": "^5.14.1",
     "classnames": "^2.2.5",
     "eventemitter3": "^3.1.0",
-    "grapheme-splitter": "^1.0.4",
     "hoist-non-react-methods": "^1.1.0",
     "image-client-api": "^1.3581.0",
     "loadjs": "^3.5.4",
--- a/packages/wix-ui-core/src/components/avatar/avatar.spec.tsx
+++ b/packages/wix-ui-core/src/components/avatar/avatar.spec.tsx
@@ -124,16 +124,6 @@ describe('Avatar', () => {
       expect(await driver.getTextContent()).toBe('ИТ');
     });
 
-    it('should preserve unicode graphemes (emoji)', async () => {
-      const driver = await createDriver(<Avatar name="🏳️‍🌈" />);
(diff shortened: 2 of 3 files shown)


mykas:
Wait, I specifically remembered that we had a user wanting this feature 🤔

mykas:
Can you double check with the task owner if this is still relevant? https://github.com/wix/wix-ui-tpa/issues/145

jjaskulis:
> Can you double check with the task owner if this is still relevant? [wix/wix-ui-tpa#145](https://github.com/wix/wix-ui-tpa/issues/145)

@mykas I see, thanks for bringing attention to this. Will discuss this over with the task owner.

jonathanadler:
@mykas 
We talked about this here: https://github.com/wix/wix-ui/pull/1510

The issue opened in wix-ui-tpa was just something that the developers thought it would be nice to have.
It did not come from a user's request, or a bug.

At the moment we don't see any reason to have this feature, since there isn't a real case for it, and the 3rd party package weighs too much.

I think it's ok to remove, furthermore, I think the tests should still pass (except the test with the emoji - which @jjaskulis you should probably remove in this PR)

jjaskulis:
> @mykas
> We talked about this here: #1510
> 
> The issue opened in wix-ui-tpa was just something that the developers thought it would be nice to have.
> It did not come from a user's request, or a bug.
> 
> At the moment we don't see any reason to have this feature, since there isn't a real case for it, and the 3rd party package weighs too much.
> 
> I think it's ok to remove, furthermore, I think the tests should still pass (except the test with the emoji - which @jjaskulis you should probably remove in this PR)

@jonathanadler updated the test spec

mykas:
@jonathanadler fine by me! As long as there isn't more dependencies like wix-ui-santa (God help us)

jonathanadler:
I see that the build is failing on lint issues.
But I also think they're not related to the PR itself 🤔

jonathanadler (approved):

jjaskulis:
@jonathanadler did a rebase, could you re-approve?

jonathanadler (approved):
Awesome thanks!

PR #2032, opened 2021-02-08
Fix prettier - remove double semicolons

jonathanadler:


EgoziE (approved):
Thanks for the contribution @jonathanadler.
Let's schedule a meeting to have a face-to-face review.

itaytay (commented):
I'm not sure this is the right approach... 🤔 
Check [this](http://json.schemastore.org/prettierrc) out for more details

NitayRabi (commented):
I think I would go with triple instead of 1 or 2. expected something else of you @jonathanadler ...

mosheKerbel:
> I think I would go with triple instead of 1 or 2. expected something else of you @jonathanadler ...

Totally agree with Nitay here.

jonathanadler:
😅  what a zubur

PR #2035, opened 2021-02-14
fix(address-input): #IE-141|disable browser autocomplete for autocomp…

IdoWeitzman:
In address component (which uses InputWithOptions), browser autocomplete should be disabled.
If not disabled, it causes weird behaviour: 
when user start typing, both browser autocomplete and input options open (browser autocomplete gets the focus)
this causes situation where pressing down/up key doesn't navigate through the input options as expected, but navigates through browser autocomplete options.

### DIFF
--- a/packages/wix-ui-core/src/components/address-input/AddressInput.spec.tsx
+++ b/packages/wix-ui-core/src/components/address-input/AddressInput.spec.tsx
@@ -850,7 +850,7 @@ describe('AddressInput', () => {
       init({});
       const browserAutoComplete = driver.getAutoComplete();
 
-      expect(browserAutoComplete).toBe('no-autocomplete');
+      expect(browserAutoComplete).toBe('off');
     });
 
     it('Should handle onChange event', () => {
--- a/packages/wix-ui-core/src/components/address-input/AddressInput.spec.tsx
+++ b/packages/wix-ui-core/src/components/address-input/AddressInput.spec.tsx
@@ -846,6 +846,13 @@ describe('AddressInput', () => {
       expect(driver.isDisabled()).toBeFalsy();
     });
 
+    it('Should pass autocomplete prop (no-autocomplete)', () => {
+      init({});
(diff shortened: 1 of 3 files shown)


zivpiz (approved):

packages/wix-ui-core/src/components/address-input/AddressInput.spec.tsx
```diff
@@ -846,6 +846,13 @@ describe('AddressInput', () => {
       expect(driver.isDisabled()).toBeFalsy();
     });
 
+    it('Should pass autocomplete prop (no-autocomplete)', () => {
+      init({});
```
giladsegal, line 850: You should have a test for the default, and a test for setting it as a prop

packages/wix-ui-core/src/components/address-input/AddressInput.tsx
```diff
@@ -541,6 +541,7 @@ export class AddressInput extends React.PureComponent<
       style: inputStyle,
       inputClassName,
       required: this.props.required,
+      autoComplete: 'no-autocomplete',
```
giladsegal, line 544: you should also update the `AddressInputProps` type

giladsegal:
@shlomitc do you think this is a desired change for WSR as well? 
If not, maybe it should be opt in instead of opt out - I mean that we can set the `autoComplete` field to the browser default and than in WUS we can override it specifically to `no-autocomplete`

giladsegal (approved):

packages/wix-ui-core/src/components/address-input/AddressInput.spec.tsx
IdoWeitzman: @giladsegal with current implementation the input doesn't get `autoComplete` value from props, but passes a fixed value `no-autocomplete` to `InputWithOptions`
should it accept autoComplete from props, and if not provided pass `no-autocomplete` ?

giladsegal (approved):

PR #2038, opened 2021-02-17
Adding onFocusByKeyboard prop to <Checkbox/>

barboaron:


packages/wix-ui-core/src/components/checkbox/Checkbox.tsx
```diff
@@ -12,6 +12,8 @@ export interface CheckboxProps extends React.InputHTMLAttributes<HTMLElement> {
   'data-hook'?: string;
   /** The onChange function will be called with a new checked value */
   onChange?: React.EventHandler<OnChangeEvent>;
+  /** A callback to invoke on focus fired from keyboard event */
+  onFocusByKeyboard? (event: React.FocusEvent<HTMLInputElement>): void;
```
shlomitc, line 16: why is this needed? what's the difference between it and the regular focus event in terms of usage?

barboaron: We need that the focus ring to appear only when it is focused by the keyboard - and not by mouse (like the RadioButton component)

shlomitc: we have a `FocusableHOC` component for that - @jonathanadler is this synced with you?

jonathanadler: It's the same issue we had with the RadioButton component

jonathanadler (commented):
Looks good 👍 

But you'll need to also add some tests, so that we ensure this is working correctly.

You can take a look at the RadioButton tests added here:
https://github.com/wix/wix-ui/commit/469424d2176a94d582ddf7f6aad64c7fd5fe533c

jonathanadler:
Also note that the build is failing over a perfer degredation (it's not related to your change, probably just a dependency).

So you should probably `pull --rebase origin master` and also bump the `video` bundle file threshold in the `perfer.config` file (according to this https://github.com/wix/wix-ui/pull/2038/checks?check_run_id=1918309704)

packages/wix-ui-core/src/components/checkbox/Checkbox.e2e.ts
```diff
@@ -60,4 +60,28 @@ describe('Checkbox', () => {
...
+          Key.TAB,
+          Key.ARROW_DOWN,
+          Key.ARROW_DOWN,
+          Key.ARROW_DOWN,
+          Key.ARROW_DOWN
+        )
+        .perform();
+      expect(await checkboxSecond.isChecked()).toBeTruthy();
+    });
+  });
```
jonathanadler, line 86: So it looks good,
however, the test won't really work.
You'll need to change 2 things:

1. Note that the tests are running on the checkbox storypage and no the checkbox **test** page
You'll need to separate your test from the others and add this `before` statement, instead of the existing one
```
const storyUrl = createStoryUrl({
    kind: Category.TESTS,
    story: 'Checkbox',
  });

  beforeEach(() => browser.get(storyUrl));
```
2. Checkbox navigation doesn't work with the arrow keys (RadioButton does).
So you'll need to change the navigation to just tabs (ie tab forward and then backwards using shift-tab)

jonathanadler (commented):
Nice!
Left just one comment

packages/wix-ui-core/src/components/checkbox/Checkbox.e2e.ts
barboaron: done

packages/wix-ui-core/src/components/checkbox/Checkbox.e2e.ts
```diff
@@ -42,22 +41,46 @@ describe('Checkbox', () => {
...
+    ).then(async () => {
+      expect(await checkboxSecond.isChecked()).toBeFalsy();
+      browser
+        .actions()
+        .sendKeys(
+          Key.TAB,
+          Key.chord(Key.SHIFT, Key.TAB),
+          Key.TAB,
+          Key.chord(Key.SHIFT, Key.TAB),
+          Key.TAB
```
jonathanadler, line 80: I think you can remove these three lines

jonathanadler (commented):
Last comment

packages/wix-ui-core/src/components/checkbox/Checkbox.e2e.ts
barboaron: I removed lines 79-80 because the second checkbox should be checked at the end of the test.

barboaron: done

jonathanadler (approved):

jonathanadler:
After the build will pass we'll merge

PR #2041, opened 2021-02-21
feat(Image&MediaImage): expose `nativeRef` prop

nitayneeman:
This PR exposes a `nativeRef` prop for Image & MediaImage instead of https://github.com/wix/wix-ui/pull/2037 which is problematic to verify and might affect current applications.

We also thought about exposing `onComplete` that "opt-in"s the logic but it's still hard to verify and actually might be confusing with the loading status... Because we need to expose the case that the image is completed but the loading status is still "loading".

Besides, I think it makes sense to expose a ref for such an atom component - similar to the ButtonNext case.

### DIFF
--- a/packages/wix-ui-core/src/components/image/image.tsx
+++ b/packages/wix-ui-core/src/components/image/image.tsx
@@ -20,7 +20,7 @@ export interface ImageState {
   src?: string;
   status: ImageStatus;
 }
-class ImageComponent extends React.PureComponent<ImageProps, ImageState> {
+export class Image extends React.PureComponent<ImageProps, ImageState> {
   private readonly getSrc = (): string =>
     this.props.src ? this.props.src : this.getSrcSet();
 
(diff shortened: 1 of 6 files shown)


packages/wix-ui-core/src/components/image/image.tsx
```diff
@@ -4,6 +4,8 @@ import { ImageStatus, FALLBACK_IMAGE } from './consts';
 
 export interface ImageProps {
   nativeProps?: React.ImgHTMLAttributes<HTMLImageElement>;
+  /** A reference to be passed to the native image element */
+  nativeRef?: React.Ref<HTMLImageElement>;
```
nitayneeman, line 8: I thought about using `forwardRef` but I think that `nativeRef` is semantically better than just `ref` in this case

jonathanadler (dismissed):
IMO it looks good 👍

PR #2043, opened 2021-02-23
Add popover content classname

jonathanadler:
🙏


mykas (approved):

shlomitc:
Why is this needed? If you -st-extend the component you can target the built in class called “popoverContent”

PR #2050, opened 2021-03-07
Adding OnBlur prop to <Checkbox/>

barboaron:


nitayneeman (commented):
Looks good, but we also need to test this logic.

We can do something similar to this:
https://github.com/wix/wix-ui/blob/master/packages/wix-ui-core/src/components/radio-button/RadioButton.spec.tsx#L167-L174

packages/wix-ui-core/src/components/checkbox/Checkbox.spec.tsx
```diff
@@ -230,5 +230,14 @@ describe('Checkbox', () => {
...
     });
+
+    it('invokes callback for onBlur with the correct value', () => {
+      const onBlur = jest.fn();
+      const checkbox = createDriver(<Checkbox onBlur={onBlur} />);
+
+      checkbox.click();
+      checkbox.blur();
+      expect(onBlur.mock.calls.length).toEqual(1);
+    });
```
nitayneeman, line 241: I would extract it to a different describe like `Blur` since it's the opposite of `Focus` 🙃

barboaron: done

PR #2051, opened 2021-03-08
Scope packages in install

mykas:


packages/wix-ui-core/package.json
```diff
@@ -66,7 +66,7 @@
     "storybook-start": "npm run storybook",
     "storybook:visual": "start-storybook -c ./.storybook/visual -p 6007",
     "start": "run-p yoshi-start storybook-start",
-    "release": "npx yoshi release && npx teamcity-surge-autorelease@^1.0.0",
+    "release": "npx @wix/yoshi release && npx teamcity-surge-autorelease@^1.0.0",
```
shlomitc, line 69: shouldn't we just remove the `npx` and keep it `yoshi`?

mykas: I remember there was some problem with that in publish build, but maybe it would work now

PR #2071, opened 2021-04-09
wix-ui-framework: mark changed components with `dirty` flag

frontned:
@argshook edit: hijacking top comment to include description

---

This PR introduces few changes for `wix-ui-framework`, focusing on `wuf update` command.

running `wuf update` creates `components.json` file, which has a structure like so:

```
{
  ComponentName: {
    path: 'path/to/compnent',
  }
}
```

this PR enhances `components.json` to also include a `dirty: true` flag for components.

`dirty: true` indicates that some component has changed as compared to `master` branch and it needs to be tested.

this flag is useful for `wuf export-testkits` command. With it we can generate files that are needed only for changed components.

for example, to run visual tests only on changed components, we need to have a file like so:

```
require('changed-component.visual.js'),
require('changed-component2.visual.js')
```

this is easy to do when having `dirty: true` flag, as can be seen here https://github.com/wix/wix-style-react/pull/7042

---

in addition, since this feature should be testable for future maintainability, this PR also includes some test utilities, like `GitTestkit`. It's usage can be seen here: https://github.com/wix/wix-ui/pull/2071/files#diff-a546f47860aa1b73b0684e0505ad3fb2ec641181eefbb68e73f5a09be71134aaR205


argshook:
This PR originally started during a pair programming guild activity.
Thanks for all the help @frontned!

No need to continue working on this if you don't want to or lack the
time, i'll take care of the rest.
unless you do want to work on it, then by all means go ahead! :)

PR #2072, opened 2021-04-12
<Playground /> - change react-live editor to codemirror editor

egidijusja:


packages/wix-storybook-utils/src/LiveCodeExample/LiveCodeExample.tsx
```diff
@@ -50,8 +44,8 @@ interface State {
   isEditorOpened: boolean;
   parseError: object | null;
   renderPreview: boolean;
+  editorInstance: any;
```
mykas, line 47: lets use real types from codemirror

egidijusja: Added codemirror types package and used it here.

PR #2073, opened 2021-04-12
wix-storybook-utils: import codemirror style globally

argshook:
using `.global.scss` extension disables css modules


egidijusja:
Thanks, this solution works as expected. Will close this PR and use this as a reference in my own PR.

PR #2075, opened 2021-04-14
[wix-storybook-utils] - new demo section

Zivilekin:


packages/wix-storybook-utils/src/Sections/views/demo.tsx
```diff
@@ -0,0 +1,7 @@
+import * as React from 'react';
+import { DemoSection } from '../../typings/story-section';
+import styles from '../styles.scss';
+
+export const demo: (props: DemoSection) => React.ReactNode = ({ demo }) => (
+  <div className={styles.demo}>{demo}</div>
```
mykas, line 6: lets make it accept title too. Use this here with sectionWithSiblings just like we did in story page section

PR #2077, opened 2021-04-15
Dropdown keyboard Accessibility - open with space key

aarnoldaas:
We got request from @NeilWix to improve dropdown accessibility - open dropdown options with space key

### DIFF
--- a/packages/wix-ui-core/src/components/dropdown/Dropdown.tsx
+++ b/packages/wix-ui-core/src/components/dropdown/Dropdown.tsx
@@ -188,7 +188,7 @@ export class DropdownComponent extends React.PureComponent<
   }
 
   isClosingKey(key) {
-    return key === 'Tab' || key === 'Enter' || key === 'Escape';
+    return key === 'Tab' || key === 'Enter' || key === 'Escape' || key === ' ';
   }
 
   onKeyDown(evt: React.KeyboardEvent<HTMLElement>) {
@@ -203,8 +203,7 @@ export class DropdownComponent extends React.PureComponent<
         this.dropdownContentRef.onKeyDown(eventKey, evt);
 
       switch (eventKey) {
-        case 'Enter':
-        case ' ': {
+        case 'Enter': {
           this.onKeyboardSelect();
           const { multi } = this.props;
           !multi && this.close();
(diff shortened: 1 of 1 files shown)


sivanharel:
@aarnoldaas - Thanks for the contribution!

The builds are failed since your commit is not signed.
In order to contribute to the library, you need to first sign your commits. If you haven’t set-up your GitHub account for signed commits, please follow this guide: https://github.com/wix/wix-style-react/blob/master/packages/wix-style-react/docs/contribution/CREATE_PR.md#sign-commits

packages/wix-ui-core/src/components/dropdown/Dropdown.tsx
```diff
@@ -203,7 +203,8 @@ export class DropdownComponent extends React.PureComponent<
         this.dropdownContentRef.onKeyDown(eventKey, evt);
 
       switch (eventKey) {
-        case 'Enter': {
+        case 'Enter':
+        case ' ': {
```
sivanharel, line 207: Enter key is also closing the list. Should the `space` also close the dropdown list or only open it?

NeilWix: both enter and space keys should toggle (expand, collapse) the options menu

sivanharel: Thanks @NeilWix!
@aarnoldaas please add this functionality as well 🙏

aarnoldaas: Clicking spacekey trigger poppover onClick method, I just need to mark space as closing key so it will not trigger open/close multiple times.

sivanharel (changes requested):
Now the `space` key opens and closes the list but doesn't select the new option.
You should also keep the `case ' ':` and then it would work as expected :)

sivanharel:
You could test it on the storybook playground tab :)

aarnoldaas:
@NeilWix should space key select option or just toggle the dropdown?

aarnoldaas:
Checked how native Select works. It selects option with space key. Made it to work same way.

sivanharel (approved):
In ord

PR #2078, opened 2021-04-18
[wix-storybook-utils] Introduce global loader config for playground components

mykas:
This PR is part of improved stories effort. This will introduce global configuration to get playground components loaded for every code section. 

You will only need to introduce path to the file which imports all components and loader will add it to every story file to be loaded.


packages/wix-storybook-utils/src/loader/index.js
```diff
@@ -4,7 +4,8 @@ const path = require('path');
 const pathFinder = require('react-autodocs-utils/src/path-finder');
 const gatherAll = require('react-autodocs-utils/src/gather-all');
 const metadataMerger = require('react-autodocs-utils/src/metadata-merger');
-const prepareStory = require('react-autodocs-utils/src/prepare-story'); // TODO: should be part of wix-storybook-utils
```
mykas, line 7: @argshook I am making this comment to actual realisation. For now I just moved `prepareStory` function near here with specific improvements for playground components, but I would like to improve it more by getting rid of `storiesOf` function etc. So my next PR will be about this.

packages/wix-storybook-utils/.storybook/webpack.config.js
```diff
@@ -39,6 +39,7 @@ module.exports = ({ config }) => {
         },
         repoBaseURL:
           'https://github.com/wix/wix-ui/tree/master/packages/wix-storybook-utils/src/components/',
+        playgroundComponentsPath: path.resolve(__dirname, '..', 'src', 'allComponents'),
```
mykas, line 42: @shlomitc @EgoziE @jonathanadler what do you think about this? Does this name make sense to you? 

What it would mean is that we will able to add this path to webpack config for storybook loader and forget about it.

packages/wix-storybook-utils/src/loader/index.js
mykas: Fun fact: I was thinking now what would Arijus say in this PR when checking the code. 

Guess what! No tests! So of course tests will be moved here too from react-autodos-utils and updated with changes that got introduced for playground components.

packages/wix-storybook-utils/.storybook/webpack.config.js
shlomitc: I think we should have a single configuration file to point at that the loader will read from.
In this file, you will define the components scope

jonathanadler: Awesome!
A single config file would be a nice approach 👌

mykas: But what does it mean a single file? Do you mean moving all these params to one file and then in this webpack config reference it?

mykas: Or you mean that loader will expect certain file to be in the same folder and try to read it?

shlomitc:
```js
/* .storybook/wsu.config.js */
module.exports = {
  playgroundComponentsPath: path.resolve(__dirname, '..', 'src', 'allComponents'),
}
```

or maybe you can already import the `allComponents` file in the config file

packages/wix-storybook-utils/src/loader/prepareStory.js
```diff
@@ -0,0 +1,122 @@
...
+          const storiesOfProperty = types.objectProperty(types.identifier('storiesOf'), types.identifier('storiesOf'));
+          const playgroundComponentsProperty = types.objectProperty(types.identifier('playgroundComponents'), types.identifier('playgroundComponents'));
+
+          path.node.properties.push(storiesOfProperty);
+          storyConfig.playgroundComponentsPath && path.node.properties.push(playgroundComponentsProperty);
+
+          configProperties = path.node.properties;
+          path.stop();
+        },
+      });
```
argshook, line 66:
```suggestion
      if (isES5) {
        // add requires
        ast.program.body.unshift(parse('const { storiesOf } = require("@storybook/react")'));
        ast.program.body.unshift(parse('const story = require("wix-storybook-utils/Story").default'));
      } else {
        // add imports
        ast.program.body.unshift(parse('import { storiesOf } from "@storybook/react"'));
        ast.program.body.unshift(parse('import story from "wix-storybook-utils/Story"'));
      }

      return ast;
    })
    .then(ast => {
      // TODO: this is not too good, unfortunatelly, i cant return
      // rejected promise from within visitor, babylon complains
      let error = null;

      const configAST = parse(`(${JSON.stringify(storyConfig)})`);
      let configProperties;

      visit(configAST)({
        ObjectExpression(path) {
          const storiesOfProperty = types.objectProperty(types.identifier('storiesOf'), types.identifier('storiesOf'));
          path.node.properties.push(storiesOfProperty);

          if (storyConfig.playgroundComponentsPath) {
            const playgroundComponentsProperty = types.objectProperty(
              types.identifier('playgroundComponents'),
              parse(`require("${path.relative(sourcePath, storyConfig.playgroundComponentsPath)}")`)
            );

            path.node.properties.push(playgroundComponentsProperty);
          }

          configProperties = path.node.properties;
          path.stop();
        },
      });
```

argshook: github shows a big diff, but the idea is not big: don't do `import`/`require` at the top, do one `require` right into the `storyConfig`

packages/wix-storybook-utils/.storybook/webpack.config.js
argshook: not sure we need to aim at this. storybook anyway needs a `webpack.config.js` which is where we keep `wix-storybook-utils` config.

if we extract the config of `wix-storybook-utils` from `webpack.config.js` to a separate file, we'll need to deal with at least two config files instead of one, which is opposite of what this comment thread is about?

mykas: I did some research on importing files already to webpack config and its problematic due the way webpack works only with common js. For now passing a path is one way we can approach this.

mykas: And about your example: do you expect it to be loaded inside webpack.config.js or you expect webpack to try to find such file?

packages/wix-storybook-utils/src/loader/prepareStory.js
mykas: oh thats nice 🤔

packages/wix-storybook-utils/.storybook/webpack.config.js
shlomitc: I'm ok with keeping it in the `webpack.config.js` file.
The default value of `playgroundComponentsPath` here should point to the `index.js` file which makes sense to have in every project

mykas: Are we sure we want default value? We have different patterns across libraries too. 
tpa: src/components
wsr: src/
wuc: src/components

shlomitc: `src/index` is the most common in libraries, but you can try to resolve both paths

mykas: one more thing. Its not just about component but also icons. We would need to have the file anyway to import icons. Or we want to assume icons and load them for every lib?

shlomitc: So looks like we must have a separate entry file for the playground context. Let’s just find a better name than the allComponents and not place it in src since it shouldn’t go to dist

packages/wix-storybook-utils/.storybook/playground.scope.js
```diff
@@ -0,0 +1,4 @@
+import * as Icons from 'wix-ui-icons-common';
```
mykas, line 1: @shlomitc what do you think? `playground.scope.js` ? any other name you have in mind?

packages/wix-storybook-utils/src/loader/prepareStory/prepareStory.spec.js
```diff
@@ -0,0 +1,320 @@
+/* global describe it expect */
```
mykas, line 1: @argshook all hail to unit tests!

packages/wix-storybook-utils/src/loader/prepareStory/prepareStory.js
```diff
@@ -0,0 +1,119 @@
...
+      let configProperties;
+
+      visit(configAST)({
+        ObjectExpression(path) {
+          const storiesOfProperty = types.objectProperty(types.identifier('storiesOf'), types.identifier('storiesOf'));
+
+          if(storyConfig.playgroundComponentsPath) {
+            const playgroundComponentsPath = pathLib.relative(sourcePath, storyConfig.playgroundComponentsPath);
+            const playgroundComponentsProperty = types.objectProperty(types.identifier('playgroundComponents'), types.identifier(`require('${playgroundComponentsPath}').default`));
+            storyConfig.playgroundComponentsPath && path.node.properties.push(playgroundComponentsProperty);
```
argshook, line 55:
```suggestion
            path.node.properties.push(playgroundComponentsProperty);
```

packages/wix-storybook-utils/src/loader/prepareStory/prepareStory.js
```diff
@@ -0,0 +1,119 @@
...
+      const configAST = parse(`(${JSON.stringify(storyConfig)})`);
+      let configProperties;
+
+      visit(configAST)({
+        ObjectExpression(path) {
+          const storiesOfProperty = types.objectProperty(types.identifier('storiesOf'), types.identifier('storiesOf'));
+
+          if(storyConfig.playgroundComponentsPath) {
+            const playgroundComponentsPath = pathLib.relative(sourcePath, storyConfig.playgroundComponentsPath);
+            const playgroundComponentsProperty = types.objectProperty(types.identifier('playgroundComponents'), types.identifier(`require('${playgroundComponentsPath}').default`));
```
argshook, line 54:
```suggestion
            const playgroundComponentsProperty = types.objectProperty(types.identifier('playgroundComponents'), parse(`require('${playgroundComponentsPath}').default`));
```

the value of a property is not an identifier, it's a function call. an easy way to get ast of that is to use `parse`

packages/wix-storybook-utils/src/loader/prepareStory/prepareStory.spec.js
```diff
@@ -0,0 +1,320 @@
...
+    });
+
+
+    it('should inject playgroundComponents require when given playgroundComponentsPath through storyConfig', () => {
+      const source = `
+      export default {};
+      `;
+
+      const config = {
+        playgroundComponentsPath: '/file'
```
argshook, line 302: `/file` would be a `file` in the root of a filesystem, that's not what we want. let's try to imitate a more realistic scenario.
```suggestion
        playgroundComponentsPath: 'playground.scope.js'
```

`playgroundComponentsPath` should probably expect a path which is relative to `.storybook/webpack.config.js`

packages/wix-storybook-utils/src/loader/prepareStory/prepareStory.spec.js
```diff
@@ -0,0 +1,320 @@
...
+
+export default story({
+  _config: {
+    "playgroundComponentsPath": "/file",
+    playgroundComponents: require('../../file').default,
+    storiesOf: storiesOf
+  }
+});`;
+
+      return expect(prepareStory(config, '/kazkas/kazkas')(source)).resolves.toEqual(expectation);
```
argshook, line 317: maybe there's a way to have a more realistic test? it's not very clear how `/kazkas/kazkas` becomes `../../`, especially when `playgroundComponentsPath` points to `/file` - a file on root

also let's use english :)
```suggestion
      return expect(prepareStory(config, '/something/something')(source)).resolves.toEqual(expectation);
```

argshook (changes requested):

packages/wix-storybook-utils/src/loader/prepareStory/prepareStory.spec.js
mykas: oh no, I left lithuanian! its our secret language! how dare me

packages/wix-storybook-utils/.storybook/playground.scope.js
shlomitc: `playgroundScope.js` could be enough. not sure we need the mid-name of `.scope` since it's not a repeating convention in the same library

PR #2079, opened 2021-04-20
[wix-storybook-utils] `<Playground/>` - autocomplete

egidijusja:


packages/wix-storybook-utils/src/LiveCodeExample/Editor.tsx
```diff
@@ -0,0 +1,154 @@
...
+              const indent = codemirror.getOption('indentUnit');
+              const spaces = Array(indent + 1).join(' ');
+              codemirror.replaceSelection(spaces);
+            }
+          },
+          "'<'": completeAfter,
+          'Cmd-I': completeIfInTag,
+          'Ctrl-I': completeIfInTag,
+          "'='": completeIfInTag,
+          "' '": completeIfInTagNewAttribute,
```
mykas, line 146: Lets see if there is a parameter called `Space` instead this

mykas:
```suggestion
          Space:completeIfInTagNewAttribute,
```

maybe?

packages/wix-storybook-utils/src/LiveCodeExample/Editor.tsx
```diff
@@ -0,0 +1,154 @@
...
+        autoCloseTags: true,
+        autoCloseBrackets: true,
+        theme: 'wsr',
+        viewportMargin: 50,
+        lineNumbers: true,
+        hintOptions: {
+          schemaInfo: hints,
+        } as CodeMirror.ShowHintOptions,
+        extraKeys: {
+          Tab: codemirror => {
```
mykas, line 133: Lest extract thus function too and call it: `handleIndentation`

packages/wix-storybook-utils/src/Playground/index.tsx
```diff
@@ -103,6 +104,7 @@ const Playground: React.FunctionComponent<Props> = ({
           });
         }}
         previewProps={{ className: styles.overflowPreview }}
+        hints={getHints(rest.scope)}
```
mykas, line 107: Lets move this logic inside LiveCodeExample instead so every code editor in storybook will benefit

packages/wix-storybook-utils/src/LiveCodeExample/Editor.tsx
egidijusja: Yeah it exists, I changed it to `Space`

PR #2088, opened 2021-04-26
Dropdown - toggle menu on `space` key only when the prop value of `en…

sivanharel:
…ableToggleMenuOnSpace`is true

### DIFF
--- a/packages/wix-ui-core/src/components/dropdown/Dropdown.spec.tsx
+++ b/packages/wix-ui-core/src/components/dropdown/Dropdown.spec.tsx
@@ -62,6 +62,36 @@ describe('Dropdown', () => {
       );
     });
 
+    it('should show content on `Enter`', () => {
+      const driver = createDriver(createDropdown({ options }));
+
+      Simulate.keyDown(document.querySelector(
+          '[data-hook="open-dropdown-button"]'),
+          { keyCode: 13 },
+          );
+      expect(driver.isContentElementExists()).toBeTruthy();
+    });
+
+    it('should show content on `Space` when `enableSpaceToggleMenu` is true', () => {
+      const driver = createDriver(createDropdown({ options, enableToggleMenuOnSpace: true, }));
+
+      Simulate.keyDown(document.querySelector(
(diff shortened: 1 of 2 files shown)


packages/wix-ui-core/src/components/dropdown/Dropdown.spec.tsx
```diff
@@ -62,6 +62,36 @@ describe('Dropdown', () => {
...
+    });
+
+    it('should not show content on `Space`  when `enableSpaceToggleMenu` is false', () => {
+      const driver = createDriver(createDropdown({ options, enableToggleMenuOnSpace: false, }));
+
+      Simulate.keyDown(document.querySelector(
+          '[data-hook="open-dropdown-button"]'),
+          { keyCode: 32 },
+          );
+      expect(driver.isContentElementExists()).toBeTruthy();
```
jonathanadler, line 92: Should this be
```suggestion
      expect(driver.isContentElementExists()). toBeFalsy();
```
?

packages/wix-ui-core/src/components/dropdown/Dropdown.tsx
```diff
@@ -188,11 +193,24 @@ export class DropdownComponent extends React.PureComponent<
   }
 
   isClosingKey(key) {
-    return key === 'Tab' || key === 'Enter' || key === 'Escape' || key === ' ';
+    const { enableToggleMenuOnSpace } = this.props;
+
+    return key === 'Tab' || key === 'Enter' || key === 'Escape' || (enableToggleMenuOnSpace && key === ' ');
```
jonathanadler, line 198: 😮  I didn't know `' '` works as space

jonathanadler (approved):
Looks good!

I left some tiny comment

packages/wix-ui-core/src/components/dropdown/Dropdown.tsx
sivanharel: I also didn't know it works up until Arnoldas from Events contributed it.
https://keycode.info/

PR #2105, opened 2021-05-07
fix all lint warnings + fix conflict with prettier

jonathanadler:
All I did here was to change the tslint.json file a bit (left a comment there) and ran `npm run lint -- --fix`


packages/wix-ui-core/tslint.json
```diff
@@ -1,46 +1,3 @@
...
-      "jsx-double"
-    ],
-    "no-floating-promises": {
-      "severity": "warn"
-    },
-    "restrict-plus-operands": {
-      "severity": "warn"
-    }
-  }
+  "extends": ["./tslint_base.json", "tslint-config-prettier"]
```
jonathanadler, line 2: This is added in order to resolve the known conflicts between tslint and prettier.

argshook: what's the purpose of splitting config to `tslint_base`? is it used somewhere else? if not, i'd just keep it here

argshook: > Actually, this is a bit more readable, don't you think?

not for me, cause this way i need to check two files and two `extends`. I'd rather have only one.

> Also, from what I read in the past, this is the way to prevent prettier and tslint colliding.

that's about adding `tslint-config-prettier` isn't? adding it shouldn't require to split into a separate file?

jonathanadler: I can't find it now, but I read it somewhere

jonathanadler: ok, so I tried combining the json's locally, and yeah, it doesn't work.
I don't really understand why, but this is the only way I managed to make it work 🤷‍♂️ .

argshook (approved):

PR #2109, opened 2021-05-18
wix-ui-core: add `--force` to `postinstall` of `@stylable/dom-test-kit`

argshook:
This PR adds `--force` to ensure `@stylable/dom-test-kit@3` is installed in users `node_modules`.

it is to ensure that at the root level of users `node_modules` they get `@stylable/dom-test-kit` version 3, and not some other version.

### DIFF
--- a/packages/wix-ui-core/package.json
+++ b/packages/wix-ui-core/package.json
@@ -40,7 +40,7 @@
     "url": "git+https://github.com/wix/wix-ui.git"
   },
   "scripts": {
-    "postinstall": "npm install @stylable/dom-test-kit@3 --no-save",
+    "postinstall": "npm install @stylable/dom-test-kit@3 --no-save --force",
     "prebuild": "npm run update-components && npm run update-components-standalone",
     "build": "npm run transpile && npm run import-path && npm run build:named-exports && build-storybook && npm run build-standalone && npm run build-puppeteer-testkits",
     "transpile": "npm run build:named-exports && yoshi build && npm run patch-stylable && npm run transpile-mixins",


barak007:
Any package that uses `@stylable/dom-test-kit` should add it as dependency. this line should be removed...it's not nice to override someone else `node_modules`

argshook:
@barak007 
> Any package that uses `@stylable/dom-test-kit` should add it as dependency. this line should be removed...it's not nice to override someone else `node_modules`

i heard a saying "should is a name of a fish" and so although i totally agree, the reality is (at least was) that users don't always do it, or they have transitive packages that don't always do it.

maybe today it's no longer the case but i am not 100% sure, therefore the "fix" is also not sure.

unfortunately i do not have capacity to investigate. Perhaps you or someone you know could help with this?

barak007:
You can change the `should` to `must`. it's not the responsibility of this library to solve issues in other projects.
`@stylable/dom-test-kit` is no different then any other dependency. so I'm not sure what do you want to investigate.

argshook:
@barak007

`postinstall` was introduced, because right after migration of stylable 2 -> 3 there were many broken builds with unclear errors like `cannot read property 'x' of undefined`.
We spent a great deal of time until it was clear that either project itself or some transitive dependency has a mix of different stylable versions and thus their build is failing. It was not trivial to just upgrade stylable in those projects.

wix-ui-core doesn't solve issues on other projects but it absolutely should not break other projects, it's a core library ending up in many places and so, changes like that must be taken with care.

what i'm afraid of is that after we remove `postinstall` some builds will start to fail but it will take time until it reaches our attention, users wont be happy, potentially some deployments might get delayed.

however, we had a brief discussion with @shlomitc  and we'll remove `postinstall` after all

https://github.com/wix/wix-ui/pull/2110

thus, closing this PR.

PR #2110, opened 2021-05-19
wix-ui-core: remove `postinstall` script

argshook:
this was added to ensure correct version of `@stylable/dom-test-kit` to
be present in the root of `node_modules` at the user site.

hopefully by now all users have updated their stylable versions and this
is no longer needed

### DIFF
--- a/packages/wix-ui-core/package.json
+++ b/packages/wix-ui-core/package.json
@@ -40,7 +40,6 @@
     "url": "git+https://github.com/wix/wix-ui.git"
   },
   "scripts": {
-    "postinstall": "npm install @stylable/dom-test-kit@3 --no-save",
     "prebuild": "npm run update-components && npm run update-components-standalone",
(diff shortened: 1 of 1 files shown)


AviVahl:
If `@stylable/dom-test-kit` it used in code imported by consumers (drivers, for example), then it should be specified in "dependencies". Otherwise, that code could:
1. blow up not finding a needed dep (best case scenario)
2. pick up an unrelated `@stylable/dom-test-kit` version (worst case scenario). this can easily lead to unexpected results.

argshook:
> If `@stylable/dom-test-kit` is used in code imported by consumers (drivers, for example), then it should be specified in "dependencies". Otherwise, that code could:
> 
> 1. blow up not finding a needed dep (best case scenario)
> 2. pick up an unrelated `@stylable/dom-test-kit` version (worst case scenario). this can easily lead to unexpected results.

i can't recall exactly why, but all of  these trick were done for a reason.
hopefully those tricks aren't needed anymore. I checked some projects that use `wix-ui-core` and those that use `@stylable/dom-test-kit` install it on their own.

regardless, here's a PR https://github.com/wix/wix-ui/pull/2111

PR #2112, opened 2021-05-24
<Checkbox/>, <Radio/> add autoFocus and focus() support

carmelc:
SCHED-21937

### DIFF
--- a/packages/wix-ui-core/src/components/radio-button/RadioButton.tsx
+++ b/packages/wix-ui-core/src/components/radio-button/RadioButton.tsx
@@ -70,6 +70,8 @@ export interface RadioButtonProps {
   className?: string;
   /** Inline style */
   style?: object;
+  /** define whether the native input radio should be auto focused */
+  autoFocus?: boolean;
 }
 
 export interface RadioButtonState {
@@ -95,6 +97,10 @@ export class RadioButton extends React.Component<
     onKeyDown: noop,
     onHover: noop,
     onBlur: noop,
+  };
+
+  focus() {
+    this.radioRef?.focus();
   };
 
   render() {
@@ -154,6 +160,7 @@ export class RadioButton extends React.Component<
           ref={(radio) => (this.radioRef = radio)}
           aria-label={this.props['aria-label']}
           aria-describedby={this.props['aria-describedby']}
+          autoFocus={this.props.autoFocus}
         />
         <span
           className={classes.icon}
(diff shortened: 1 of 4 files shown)


packages/wix-ui-core/src/components/radio-button/RadioButton.tsx
```diff
@@ -70,6 +70,8 @@ export interface RadioButtonProps {
   className?: string;
   /** Inline style */
   style?: object;
+  /** define whether the native input radio should be auto focused */
+  autoFocus?: boolean;
```
shlomitc, line 74: `autoFocus` is a bad practice since it doesn't behave consistently in case you use it in multiple components.

I suggest using only the `ref` method

carmelc: That is true, but since Checkbox already have it, I wanted the API to be consistent, and removing it from Checkbox is a breaking change, so I am adding `focus` method to both, but also making sure that autoFocus is supported in both,

shlomitc: Bad practices we intend to deprecate and remove, so I wouldn’t dog the hole deeper just because we already have a hole

nitayneeman (approved):
Looks great.

Regarding `autoFocus` - after a discussion we decided to expose it mainly to align with Checkbox and treat to that like other native props (`checked`, `disabled`, etc.)

jonathanadler:
> Looks great.
> 
> Regarding `autoFocus` - after a discussion we decided to expose it mainly to align with Checkbox and treat to that like other native props (`checked`, `disabled`, etc.)

The `autofocus` attribute is used for focusing an element on a [page load](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#htmlattrdefautofocus), so it won't work correctly for dynamic content.

Also, if in theory it would work on each render of the input, it would cause focus issues for consumers using it.

So I agree with @shlomitc that this shouldn't be used here

carmelc:
> discussion

@jonathanadler I think that you are confusing native `autofocus` with react's `autoFocus`
when you set `autoFocus` in JSX + React, react actually implements it properly by calling `focus()` when the component mounts.
You can see some explanation [here](https://blog.maisie.ink/react-ref-autofocus/).
React imitates the autofocus behavior but in a more cross-browser manner.

BTW, native `autofocus` also works when you add an element (or even just the prop) to the DOM in most browsers.

PR #2114, opened 2021-05-27
[wix-storybook-utils] Playground - auto save feature

egidijusja:


packages/wix-storybook-utils/src/LiveCodeExample/LiveCodeExample.tsx
```diff
@@ -122,7 +144,9 @@ export default class LiveCodeExample extends React.PureComponent<Props, State> {
   };
 
   onEditorChange = code => {
-    this.setState({ code }, () => this.props.onChange(this.state.code));
+    const { title, onChange, storyName } = this.props;
+    this.setState({ code }, () => onChange(this.state.code));
+    window.sessionStorage.setItem(`${storyName}-${title}`, code);
```
shlomitc, line 149: why not localStorage? it's the most popular usage I'm familiar and one that we use in Wix

egidijusja: We were thinking about `localStorage` and `sessionStorage` while implementing this solution. We chose `sessionStorage`, because it seems that it fits our use case better.

localStorage doesn't have expiry date, so it means that modified code will be saved and loaded to the storybook examples for indefinite time. If user modifies some examples and opens storybook after days or weeks he'll see the same modified code (even our updated code examples during that time wouldn't be shown until user press `reset code` button.

sessionStorage is cleared when browser tab is closed, but it remains if user refreshes the tab.

shlomitc: 1. why do we care if this snippet will always exist? I don't think that's that bad
2. I see now that you will save the snippet change for each component story. I think that's not needed, do this only in the main playground where snippet sharing is what makes sense. the examples in the stories should always be kept the original ones when you refresh or move between pages

egidijusja: The survey about WSR playground (conducted by UX) revealed that one of the major issues is that users want to check API table and play around with our examples at the same time. Examples and API table are on different tabs and on tab switch code is reset, so users usually have two opened tabs at the same time (one for API, one for examples) as a workaround.

egidijusja: Maybe we could have localStorage for playground and sessionStorage for components stories. Playground page has dummy example anyway and users might be building more complex prototypes longer, so session storage here might not be the best fit as the code is reset on tab close.

shlomitc: As far as I understand the complain is moving between component API table and the full playground page, not the story page.

mykas: Actually not just the main playground but also story playgrounds. The main playground is yes most popular place to work, but story playgrounds are played around too and some unexpected refresh or tab change destroys the prototyping experience.

mykas: I agree with @egidijusja that story playgrounds should be done still with local sessions. The amount of times I had bad exp in even migrating the story and playing around in the story playground is too much.

mykas: And also it does not hurt the experience if we introduce local storage solution for main playground and local session per story

shlomitc: So in the case of an example in a specific story page, moving to the API tab should keep the example, but once the user moved to another story page, it is unexpected for the modified snippet to be saved. It "ruins" the original example and not common in any documentation system I've seen

mykas: Yes that could work, but I am not sure how to solve this because our pages are rendered new when you switch to different tabs. You either need to save it or not rerender it. Maybe keeping in some local state this could be solved, but I am not sure

egidijusja: This could be solved by removing session storage data about previous story when navigating to a new one.

mykas: @shlomitc WDYT? ☝️

shlomitc: it will still preserve it on refresh, so maybe it's better to clean it when you're entering the story page and not leaving it.

egidijusja: Yes, it makes sense.

So to sum up:
1. We will have local storage on main playground
2. We will have session storage on components stories and it will exist till user goes to another story or refreshes the page.

right?

shlomitc: sounds great

mykas (approved):

PR #2118, opened 2021-05-31
Dropdown  - add additional dropdown behaviour for better A11y support

jjaskulis:
Main issue was that trying to open dropdown using space key, would open it and automatically close it immediately. This is due to not preventing default behaviour when handling key down, so the space key would also trigger onClick handler, which controls state of dropdown..

Also Removed callback in open method, it was used only in onKeyDown method and it's usage doesn't really depend on running after setting the open state, but it just depends on the state being open. So made that explicit.


sivanharel (approved):
Great job!!

PR #2126, opened 2021-06-17
skipPropsWithoutDoc

kobiburnley:
We use `react-storybook-utils` with TypeScript, and we would like to filter out props that don't have a description

### DIFF
--- a/packages/react-autodocs-utils/src/parser/react-docgen-parse.js
+++ b/packages/react-autodocs-utils/src/parser/react-docgen-parse.js
@@ -7,11 +7,11 @@ const componentResolve = require('./component-resolve');
 
 const ensurePropsKey = object => ({ props: {}, ...object });
 
-const parseTypescript = path => ensurePropsKey(typescriptParser.parse(path)[0] || {
+const parseTypescript = path => ensurePropsKey(typescriptParser.parse(path, {
   propFilter: {
     skipPropsWithoutDoc: true,
   }
-}); // react-docgen-typescript returns array, so
+})[0] || {}); // react-docgen-typescript returns array, so
 
 const parseJavascript = (source = '', sourcePath = '') => {
   let parsed;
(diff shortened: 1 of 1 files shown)


argshook:
Hey there @kobiburnley, pardon for such a late response, i had a whole
month of vacation without any javascript!

Unfortunately, i cannot accept this PR, because it would break existing
users who depend on getting undocumented props in the output. This is
for sure the case in wix-design-systems, for example. It's also
indicated by red tests in the build of this PR.

To prevent breaking existing users, this functionality should be
configurable. Here's a PR addressing exactly that: https://github.com/wix/wix-ui/pull/2142

With it, you'd be able to use `gatherAll(path, { skipPropsWithoutDoc: true })`

Let me know if this would work for you and i'll release that change
under new version.

Thanks and sorry for such a long delay!

kobiburnley:
@argshook is there a way to pass this option via a story using `wix-storybook-utils` webpack plugin?

argshook:
> @argshook is there a way to pass this option via a story config using `wix-storybook-utils` webpack plugin?

ah, you said you use `react-autodocs-utils`, so i assumed you use it directly.

At the moment no, there is no way, but with the aforementioned PR it's trivial to implement

kobiburnley:
> We use `react-storybook-utils` with TypeScript, and we would like to filter out props that don't have a description

`react-storybook-utils`

argshook:
hey @kobiburnley, i have released new versions of both
`react-autodocs-utils` and `wix-storybook-utils`.

You should be able to use now `skipPropsWithoutDoc` setting through the
`WixStorybookWebpackPlugin` like so:

```js
new WixStorybookWebpackPlugin({
  // ...
  parserOptions: {
    skipPropsWithoutDoc: true,
  },
  // ...
}),
```

make sure to use latest version of `wix-storybook-utils`: 4.0.30

PR #2133, opened 2021-06-29
expose AnchorTitled component

TalZa1810:
Expose `<AnchorTitled/>` component so it can be used within the cheatsheet


packages/wix-storybook-utils/src/AnchoredTitle/index.tsx
```diff
@@ -0,0 +1,81 @@
...
+    };
+
+    const isChildren = React.Children.count(children) > 0;
+
+    const onAnchorClicked = event => isChildren? onCopy(event): event.preventDefault();
+
+    return (
+        <div
+            className={styles.titleContainer}
+            onClick={!isChildren ? onCopy : null}
```
TalZa1810, line 45: this was changed

packages/wix-storybook-utils/src/AnchoredTitle/index.tsx
```diff
@@ -0,0 +1,81 @@
...
+                data-tip
+                className={styles.link}
+                id={id}
+                href={`#${id}`}
+                target="_self"
+                onClick={onAnchorClicked}
+            >
+                <LinkSmall />
+            </a>
+            { isChildren ?
```
TalZa1810, line 61: this was changed

packages/wix-storybook-utils/src/AnchoredTitle/index.tsx
```diff
@@ -0,0 +1,81 @@
...
+    const onMouseEnter = () => ReactTooltip.show(copyLinkRef.current);
+
+    const onMouseLeave = () => {
+        ReactTooltip.hide(copyLinkRef.current);
+        setCopied(false);
+    };
+
+    const isChildren = React.Children.count(children) > 0;
+
+    const onAnchorClicked = event => isChildren? onCopy(event): event.preventDefault();
```
TalZa1810, line 40: adding a check here `isChildren`

TalZa1810:
Hey @mykas,
I've added an example to the components section within storybook.
I wasn't sure whether I should add unit tests for this component.
Also, do I need to update types definitions? I saw this file but wasn't sure where should I update the types:
`packages/wix-storybook-utils/src/typings/story-section.ts`.


CC: @milkyfruit 
I just wanted to make sure, UX wise, that it's ok to use this component within the cheatsheet so that users can share a link to a specific example within the cheatsheet as well like we have with regular storybook examples.

packages/wix-storybook-utils/src/AnchoredTitle/index.tsx
mykas:
```suggestion
            onClick={!isChildren && onCopy}
```

mykas (changes requested):

packages/wix-storybook-utils/src/AnchoredTitle/index.tsx
```diff
@@ -0,0 +1,81 @@
...
+import Markdown from '../Markdown';
+import styles from './styles.scss';
+import sectionWithSiblingsStyles from '../Sections/section-with-siblings/styles.scss'
+
+const translations = {
+    copyLink: 'Copy Link',
+    linkCopiedToClipboard: 'Copied',
+};
+
+export const AnchoredTitle = ({ title, children }) => {
```
mykas, line 16: I think you missed to update the place where this AnchoredTitle is used in Section folder.

mykas: https://github.com/wix/wix-ui/blob/ef8346bb7aecef383e7ddd8582d5f44bc185a13b/packages/wix-storybook-utils/src/Sections/section-with-siblings/index.tsx#L11

mykas:
> packages/wix-storybook-utils/src/typings/

Typings there are for storybook sections so there is no need for that. Regarding unit tests this component did not have any, but it would be great if you could add some knowing that we have now another place that uses it.

mykas (approved):

PR #2136, opened 2021-07-01
<Dropdown/> restrict prevent default only to certain keys

jjaskulis:


jonathanadler (commented):
Looks legit 🥇 

I wonder if we can maybe add a test that simulates the issue that we encountered in the first place, and then make sure that this will fix it

jonathanadler:
> Looks legit 🥇
> 
> I wonder if we can maybe add a test that simulates the issue that we encountered in the first place, and then make sure that this will fix it

Also, we need to wait for the build to end to see that all the tests pass

jjaskulis:
@jonathanadler Yeah agree that test would be nice, but I'm not sure how add that test quickly and can't spare the time currently

jonathanadler:
It's kind of a stupid test, but you can add this:

```js
describe('dropdownA11yFixes', () => {
    it('should tab out of dropdown when dropdownA11yFixes is true', () => {
      const preventDefaultSpy = jest.fn();
      const driver = createDriver(
        createDropdown({ options, dropdownA11yFixes: true }, true),
      );

      Simulate.keyDown(driver.getTargetElement(), {
        key: 'Tab',
        preventDefault: preventDefaultSpy,
      });

      expect(preventDefaultSpy).not.toBeCalled();
    });
  });
```

It should fail with the old code, and pass with the new.
Can you try it?

jjaskulis:
@jonathanadler added it with minor change, fails with old code and passes with this fix.

jonathanadler:
> @jonathanadler added it with minor change, fails with old code and passes with this fix.

Yes!
👑

jonathanadler (approved):

PR #2137, opened 2021-07-04
Update to ts-lib 2.3.0

mosheKerbel:
Align versions with document-services-json-schemas that is using ts-lib 2.3.0
we are integrating with it in lean components and wsr gets the current ts-lib version of wix-ui-core.


mosheKerbel:
FYI @shlomitc 
@tompere

PR #2144, opened 2021-07-07
Fix dropdown trigger with space and enter keys

jonathanadler:


packages/wix-ui-core/src/components/dropdown/Dropdown.tsx
```diff
@@ -144,7 +153,9 @@ export class DropdownComponent extends React.PureComponent<DropdownProps, Dropdo
     this.setState({ isOpen: true }, () => this._onExpandedChange());
   }
 
-  onPopoverClick() {
+  onPopoverClick(e) {
+    const { dropdownA11yFixes } = this.props;
+
```
sivanharel, line 158: I think you can revert it

sivanharel (approved):
Looks good!

PR #2149, opened 2021-07-08
AutoTestkit - use importTestkitPath prop

EgoziE:
Currently "wix-style-react/dist" is hardcoded.
So I created importTestkitPath prop to solve taht.

### DIFF
--- a/packages/wix-storybook-utils/src/AutoTestkit/unified-testkit-documentation.tsx
+++ b/packages/wix-storybook-utils/src/AutoTestkit/unified-testkit-documentation.tsx
@@ -5,10 +5,12 @@ import { makeImportCode } from './make-import-code';
 import { Metadata } from '../typings/metadata';
 import { FieldsDocumentation } from './fields-documentation';
 import { Descriptor } from './typings';
+import { StoryConfig } from '../typings/story-config';
 
 interface Props {
   dataHook?: string;
   metadata: Metadata;
+  storyConfig: StoryConfig;
 }
 
 const extractNested = (descriptors: Descriptor[]) =>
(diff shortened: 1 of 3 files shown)


packages/wix-storybook-utils/src/AutoTestkit/unified-testkit-documentation.tsx
```diff
@@ -48,9 +51,9 @@ export const UnifiedTestkitDocumentation: React.FunctionComponent<Props> = ({
       <Code dataHook="auto-testkit-driver-import-code">
         {makeImportCode({
           testkit: {
-            template: `import { <%= component.displayName %>Testkit } from 'wix-style-react/dist/testkit';
-import { <%= component.displayName %>Testkit } from 'wix-style-react/dist/testkit/enzyme';
-import { <%= component.displayName %>Testkit } from 'wix-style-react/dist/testkit/puppeteer';`,
+            template: `import { <%= component.displayName %>Testkit } from '${storyConfig.config.importTestkitPath}/testkit';
```
argshook, line 54: i thought the idea is to add `importTestkitFormat`, so that the whole import code block can be added through webpack config?

EgoziE: I know, but I like it as `importTestkitPath`, it looks good.
Don't worry we can change it later because it's internal, I'll merge in the meantime 😍

argshook: alright, but please open a task for it so that we don't forget

PR #2157, opened 2021-07-20
Input - add input ref props

Shakedtu:
In order to get the native input ref from component using wix-ui-tpa Input field


jonathanadler (approved):
Looks great!

There's just an issue with the build.
I opened a ticket about it (https://jira.wixpress.com/browse/DEVEX-18170), hopefully it will be resolved and we can merge

PR #2159, opened 2021-07-26
<Input /> - change input ref prop to function

Shakedtu:
In order to be aligned with wix-style-react, I changed the input ref to be a function that gets the input. It makes the ref to make flexible, since now we can assign multiple inputRefs on the same react.ref


jonathanadler (approved):

jonathanadler:
Theoretically this is a breaking change.
But since we just released this feature I guess it would be ok 😄

PR #2175, opened 2021-08-01
[Address-Input] Adding appendTo prop for the use case that an AddressInput component inside of a modal, and the options element should overflow out of the modal

barboaron:
…


env-github-token:
### 🤖 Hi, I'm #Janet!

Project Name: **wix-ui-core**

* https://wix-pages.com/wix-ui-core.pr_2175/

jonathanadler (commented):
Looks good 👌 

But I think it would be a good idea to add a test ensuring this prop exists.
Otherwise someone might remove it and break your app.

barboaron:
> Looks good 👌
> 
> But I think it would be a good idea to add a test ensuring this prop exists.
> Otherwise someone might remove it and break your app.

Added :) @jonathanadler

jonathanadler:
> > Looks good 👌
> > But I think it would be a good idea to add a test ensuring this prop exists.
> > Otherwise someone might remove it and break your app.
> 
> Added :) @jonathanadler

Maybe add to `AddressInput` as well 😬

jonathanadler (approved):
Nice!

jonathanadler:
Let's just wait for all the tests to run, and then we can merge :)

PR #2178, opened 2021-08-02
Revert "Fix publish do revert (#2177)"

jonathanadler:
This reverts commit b1655e4b9d9c486cf122711483c77fb670fce282.

So I had to remove Janet in order to get the master build work again.
I wasn't able to find an easier solution.

It seems that the release script is running without node_modules and without the build result, so both `build-storybook` and `npx build-storybook` didn't work in this stage.
We had a similar issue when we moved to the monorepo (@argshook should probably remember, cause I negged him a lot about it 😛 ).

So I created this PR to try and bring Janet back.
We'll probably need to do some changes with this script, maybe even talk to CI.

btw, we can of course `npm i && npm run build` but it will take a long time, and this build already takes a long time.


env-github-token:
### 🤖 Hi, I'm #Janet!

Project Name: **wix-ui-core**

* https://wix-pages.com/wix-ui-core.pr_2178/

packages/wix-ui-core/package.json
```diff
@@ -65,7 +65,7 @@
     "storybook-start": "npm run storybook",
     "storybook:visual": "start-storybook -c ./.storybook/visual -p 6007",
     "start": "run-p yoshi-start storybook-start",
-    "release": "npx @wix/yoshi@4 release && npx teamcity-surge-autorelease@^1.0.0",
+    "release": "npx @wix/yoshi@4 release && npx build-storybook && npx janet deploy --ci",
```
argshook, line 68:
```suggestion
    "release": "npx @wix/yoshi@4 release && npx build-storybook --package @storybook/react@5 && npx janet deploy --ci",
```

executing `npx build-storybook` would try to install `build-storybook`
package and run `build-storybook` from `node_modules/.bin`. But in this
case `build-storybook` is coming from `@storybook/react` package.

with `--package` flag we can set which package to get this binary from.

argshook (changes requested):

argshook:
we might need to add `npm install` to `release` script, cause i'm not sure if only `build-storybook` will be enough

packages/wix-ui-core/package.json
jonathanadler: Nice!

It's true, cause I tried it with `npx build-storybook` and it failed with
`npm ERR! 404 Not Found - GET http://npm.dev.wixpress.com/build-storybook - not_found`

jonathanadler:
> we might need to add `npm install` to `release` script, cause i'm not sure if only `build-storybook` will be enough

@argshook `npm install` and `npm build` as well?

If I'm not mistaken, the release process does come with the `dist` folder, right?
Maybe we can have a storybook configuration that uses the `dist` components?
And then we won't need to `npm install` and `npm build`?

PR #2180, opened 2021-08-04
add css var in order to change the previewExampleFont

sivanharel:


packages/wix-storybook-utils/src/ui/typography.scss
```diff
@@ -76,6 +76,11 @@ $font: (
   }
 }
 
+/** This mixin is used to define the preview examples' font */
+@mixin previewExampleFont($type: madeFor) {
```
jonathanadler, line 80: I think the argument can be removed
```suggestion
@mixin previewExampleFont() {
```

jonathanadler (approved):
Looks good to me 👍 
Just left one small comment.

Regarding the name of the mixin, this is also fine but I also thought of some other options:
`codePreviewFont`
`examplePreviewFont` (switched the example and the preview 😁 )
or maybe just `exampleFont`? or `previewFont`?

@argshook wdyt?

argshook (commented):
This is a valid solution but i'm not very into it, because it introduces additional config that needs to be maintained.

Ideally imo, there should be no font to override, so before going this
approach i want to first check how much impact there is with removing
some fonts from wix-storybook-utils. For example removing `@include font` from `.tab` in sections

sivanharel:
@argshook - I agree that we might need to remove the @include font from ".tab" in sections and apply this mixin only on the specific elements. For example, we would like to use MadeFor for the "import part" and for the "edit button".
But this change requires us to check that we are not breaking any WSR or BaseUI component.

Screen Shot 2021-08-05 at 11 00 28

argshook:
> But this change requires us to check that we are not breaking any WSR or BaseUI component.

yes i'm aware, that's what i'm doing now :)

argshook:
Here's a PR https://github.com/wix/wix-ui/pull/2182

argshook:
I'm merging https://github.com/wix/wix-ui/pull/2182
thus, this PR should not be required anymore.
wix-storybook-utils will self-release in couple of minutes.

It should work well, but @sivanharel let me know if there are some issues

jonathanadler:
@argshook but wasn't that the whole purpose for adding the madefor font to the storybook?
So that all the storybook text will be using Madefor?

If not, then why do we need madefor inside storybook-utils at all?

sivanharel:
Thanks @argshook !
@jonathanadler - The storybook items already have the madeFor font specified :)

argshook:
> @argshook but wasn't that the whole purpose for adding the madefor font to the storybook?
> So that all the storybook text will be using Madefor?
> 
> If not, then why do we need madefor inside storybook-utils at all?

@jonathanadler 

madefor is set in wix-storybook-utils on non-user components like titles, texts, buttons and similar. But there was an issue that font was leaking into LiveCodeExample preview. Users of wix-storybook-utils, like wix-ui-tpa, doesn't need madefor rendered within LiveCodeExample preview and #2182 fixes that.

PR #2188, opened 2021-08-12
wix-ui-framework/custom-module-resolve

argshook:
This PR adds a helper function `resolveRequire`.

The purpose is to support resolving paths to files that are considered
as entry files to component.

for example, user might have component folder structure like so:

```
.
└── src
    └── components
        └── SomeComponent
            └── SomeComponent.tsx
```

note that there's no `index.ts`. To support such case, we have an
assumption, that file with the same name as folder name is the entry
file.

This assumption stands true only when there's no `index.js`, `index.ts`, `index.jsx` or `index.tsx` files.


argshook:
in addition, the helper in this PR is already used in `get-dirty-components` (for `wuf update --verbose-output ...`) and `include-metadata` (for `wuf make` to work across libraries)

PR #2191, opened 2021-08-13
wix-base-ui/destruct-imports

EgoziE:
Transform this:
```
import UI from 'baseUILib';

() => <UI.Thumbnails />;
```

to this:
```
import { Thumbnails } from '@wix/wix-base-ui';

() => <Thumbnails />;
```


packages/wix-ui-codemod/src/wix-base-ui/destruct-imports.ts
```diff
@@ -0,0 +1,88 @@
...
+        name: { object: { object: { name: importName } } },
+      });
+
+      openingTagsLevelTwo.forEach((node) => removeLevelTwo(node, j));
+      closingTagsLevelTwo.forEach((node) => removeLevelTwo(node, j));
+    });
+
+  // Replace imports
+  importNodes.replaceWith((_, i) => {
+    if (i === 0) {
```
EgoziE, line 72: @argshook this is the only line I don't like.
replaceWith works on every node (import statement) so I get this line twice.
I tried using jscodeshift-add-imports but got weird errors

argshook: maybe instead of replacing, first remove all old and then add a single new import declaration?

EgoziE: I tried it but `importNodes` remembers 2 iterables.
Anyway it doesn't bother me too much :)

EgoziE:
I also added the case of using BaseUI import as members:

From:
```
console.log(UI.Divider);

React.createElement(baseUILib.Divider, {
  className: 'divider',
});
```

To:
```
console.log(Divider);

React.createElement(Divider, {
  className: 'divider',
});
```

packages/wix-ui-codemod/src/wix-base-ui/destruct-imports.ts
argshook: it remembers 2, because tests fixture has two, maybe that's not the case in real usecases. anyway, not a big deal, can stay as is until (if ever) we face issues :)

packages/wix-ui-codemod/src/wix-base-ui/destruct-imports.ts
```diff
@@ -0,0 +1,60 @@
+import { Transform } from 'jscodeshift';
+
+function transform(file, api): Transform {
```
argshook, line 3:
```suggestion
const transform: Transform = (file, api) => {
```

packages/wix-ui-codemod/src/wix-base-ui/destruct-imports.ts
```diff
@@ -0,0 +1,60 @@
...
+  const namedImports = new Set();
```
argshook, line 4:
```suggestion
  const namedImports: Set<string> = new Set();
```

packages/wix-ui-codemod/src/wix-base-ui/destruct-imports.ts
```diff
@@ -0,0 +1,60 @@
...
+  const j = api.jscodeshift;
+  const root = j(file.source);
+
+  const removeMembers = (node) => {
+    const memberName = node.value.property.name;
```
argshook, line 9:
```suggestion
  const removeMembers = (node: ASTPath<MemberExpression>) => {
    const memberName = (node.value.property as Identifier).name;
```

packages/wix-ui-codemod/src/wix-base-ui/destruct-imports.ts
```diff
@@ -0,0 +1,60 @@
+import { Transform } from 'jscodeshift';
```
argshook, line 1:
```suggestion
import { Transform, ASTPath, MemberExpression, Identifier } from 'jscodeshift';
```

argshook (approved):
imo let's merge and start using.

We'll probably see issues in real projects and will have to adapt, but that can be done in next PRs.

i added some suggestions to silence typescript warnings, but they're not
required for a merge, imo

PR #2193, opened 2021-08-17
wix-ui-framework: import path for plugins

argshook:
This PR uses `package.json` `exports` field to specify exports path for `plugins`.

Users should be able to, for example, `require('wix-ui-framework/plugins/include-metadata)`. Such `require` worked before this PR, but those files possibly could require other files through wrong relative path. That could be avoided by introducing a proxy file. However, all that can be avoided by the use of native node `exports` feature https://nodejs.org/docs/latest-v14.x/api/packages.html#packages_package_entry_points


shlomitc:
Nice!

PR #2210, opened 2021-08-29
RadioButton - add blur method, aria-labelledby

gendeld:
Required for new Editor Elements RadioButton component


packages/wix-ui-core/src/components/radio-button/RadioButton.tsx
```diff
@@ -160,6 +166,7 @@ export class RadioButton extends React.Component<
           ref={(radio) => (this.radioRef = radio)}
           aria-label={this.props['aria-label']}
           aria-describedby={this.props['aria-describedby']}
+          aria-labelledby={this.props['aria-labelledby']}
```
jonathanadler, line 169: Maybe we should add a simple test for the `aria-labelledby` prop, just for coverage 🤔 ?

jonathanadler (commented):
Cool 👌 

I think it would be nice to add a test for the `aria-labelledby` prop, just in case someone tries to remove it in the future

jonathanadler (approved):

jonathanadler (approved):
💪

jonathanadler (approved):

PR #2240, opened 2021-10-11
Create Stylable testkit

sivanharel:


packages/wix-ui-codemod/src/wix-ui-tpa/__tests__/import-stylesheet-via-index-file.spec.ts
```diff
@@ -0,0 +1,33 @@
+import { expect } from 'chai';
```
shlomitc, line 1: use `jest` expectations instead of `chai`

packages/wix-ui-codemod/package.json
```diff
@@ -46,6 +46,8 @@
     "@types/jscodeshift": "^0.6.3",
     "@types/node": "^12.20.27",
     "@wix/yoshi": "^5.0.0",
+    "chai": "^4.3.4",
```
shlomitc, line 49:
```suggestion

```

packages/wix-ui-codemod/src/wix-ui-tpa/__tests__/import-stylesheet-via-index-file.spec.ts
shlomitc, line 1:
```suggestion

```
use jest's built-in expect instead

shlomitc (changes requested):
please add also the first test that uses this infrastructure

shlomitc (approved):

PR #2273, opened 2021-11-04
WSR 10: codemod improvements

egidijusja:
Changes: 

1. Add support for components renaming:
    - TimeInputNext --> TimeInput
    - SortableList --> SortableListBase
    - StyledNestableList --> NestableList

2. Before applying changes to component verify that it is from **wix-style-react** library.


packages/wix-ui-codemod/src/wix-style-react/migrate-wsr10.ts
```diff
@@ -11,19 +16,126 @@ type RemoveInterface = {
...
 
-  const findJSXElementThatContains = (name: string) =>
-    root.find(j.JSXElement, {
-      openingElement: {
-        name: { name },
-      },
-    });
+  const findOpeningTagPaths = (name: string) =>
+    findWSRComponentImport(name).length
+      ? root.find(j.JSXOpeningElement, { name: { name } })
```
argshook, line 137:
```suggestion
      root.find(j.JSXOpeningElement, { name: { name } }).filter(path => findWSRComponentImport(name).length)
```

maybe something like this to secure ourselves from accidentally changing user components that have the same name as some wsr component, but is not imported from wsr?

for example, `<Input/>` in the following case should be unchanged

```diff
import { Input } from 'wix-ui-tpa/dist/src/components/Input';
-import { TimeInputNext } from 'wix-style-react';
+import { TimeInput } from 'wix-style-react';

// ...

<Input size="normal"/>
-<TimeInputNext />
+<TimeInput />
```

argshook (approved):
just an idea about preventing accidental changes, otherwise nice work!

mykas (approved):

PR #2275, opened 2021-11-05
wix-ui-codemod: suggestion for how to handle FontUpgrade removal

argshook:
- remove unnecesary attributes
- use `as="something"` prop, if one exists


egidijusja (approved):
Good job! One more edge case will be covered with this PR.

PR #2278, opened 2021-11-08
WSR 10: support backwards compatible migration

egidijusja:
added command line option for migration scripts:
```
"--backwards-compatible-only" - change code to new major version API which is backwards compatible with previous major version as well.
```
Please check option name and description and let me know if you have better naming suggestions @argshook @mykas


argshook (approved):
i don't have a reason why this shouldn't be added. Naming is good imo

mykas (approved):
Great work! I like the naming now too

PR #2280, opened 2021-11-09
Relaced old icons with newer version and updated two icon names

Zohar-Telor:
Replaced "AppMarketNew" & "BlogNew" with AppMarket" & "Blog", updated all left menu icons to be 30px and updated all the icons to the new versions.


env-github-token:
### 🤖 Hi, I'm #Janet!

Project Name: **wix-ui-icons-common**
Commit: **714395174977924091d176db662abe1e0dc89594**

* https://bo.wix.com/pages/wix-ui-icons-common.pr_2280/

packages/wix-ui-icons-common/src/classic-editor/system/metadata.ts
```diff
@@ -304,16 +304,16 @@ const systemIconsMetadata: Array<IconMetadata> = [
     description: '',
     tags: [],
     sizes: {
-      '32': 'Background',
+      '30': 'Background',
     },
   },
   {
-    title: 'BlogNew',
+    title: 'Blog',
```
TalZa1810, line 311: this means renaming the component. This could break existing users so we can't do this.

packages/wix-ui-icons-common/src/classic-editor/system/metadata.ts
```diff
@@ -277,16 +277,16 @@ const systemIconsMetadata: Array<IconMetadata> = [
     description: '',
     tags: [],
     sizes: {
-      '32': 'AppManager',
+      '30': 'AppManager',
     },
   },
   {
-    title: 'AppMarketNew',
+    title: 'AppMarket',
```
TalZa1810, line 284: this could break existing users so we can't do this.

packages/wix-ui-icons-common/src/classic-editor/system/metadata.ts
```diff
@@ -589,28 +589,28 @@ const systemIconsMetadata: Array<IconMetadata> = [
...
-    sizes: { "26": "BusinessApps" },
+    sizes: { "30": "BusinessApps" },
   },
   {
     title: "Elements",
     category: 'GFPP',
     description: '',
     tags: [],
-    sizes: { "26": "Elements" },
+    sizes: { "30": "Elements" },
```
TalZa1810, line 613: I dont see a change with this file. it's not in the PR.

TalZa1810 (changes requested):
Please review my comments. 
Also, please let's split this to multiple PRs because it makes this harder to track and make sure we don't break anyone when there are too many files.
Thanks 🙏

packages/wix-ui-icons-common/src/classic-editor/system/metadata.ts
Zohar-Telor: @TalZa1810 Can you check if someone is using it? because this icon is for a very specific use-case which was not using baseui.
They are now starting to replace the existing icons there, so better change the name before.

packages/wix-ui-icons-common/src/classic-editor/system/metadata.ts
Zohar-Telor: @TalZa1810 Same thing.

packages/wix-ui-icons-common/src/classic-editor/system/metadata.ts
Zohar-Telor: Should be 26, like all the other icons in it's category, but yes it's not related to this PR, i guess it's something i changed back when uploaded it and didn't push?

Zohar-Telor:
I actually asked the guys in the office if it would be better to split it into different PRs or to a big one and was told to do one because it's easier to review :\

PR #2287, opened 2021-11-17
[wix-ui-codemod] create codemod for esm imports

bamai:
https://jira.wixpress.com/browse/DSM-3120


packages/wix-ui-codemod/src/wix-ui-tpa/__testfixtures__/import-components-via-index-file.output.js
```diff
@@ -0,0 +1,20 @@
...
+  Skins as s,
+  Counter as counter,
+  CounterSize as cs,
+  Dropdown,
+  DropdownProps as dp,
+  DROPDOWN_ALIGNMENT,
+} from 'wix-ui-tpa';
+
+import { component as c2 } from 'some-other-library'; //unrelated import
+import { component as c3 } from 'some-other-new-library'; //unrelated import
```
jonathanadler, line 18: Just curious about the order here.

Two things that come to mind:
1. Do we want to leave the empty lines? This might not be a code style our consumers would like to have and their linters support 
2. What is the order of the final imports here? Some linters might also scream about this as well. Maybe we need to add the new import in the line of the first `wix-ui-tpa` import we found, and remove the lines of the rest?

packages/wix-ui-codemod/src/wix-ui-tpa/import-components-via-index-file.ts
```diff
@@ -0,0 +1,60 @@
+import { Transform, ImportSpecifier } from 'jscodeshift';
+
+const transformComponentsImports: Transform = (fileInfo, api) => {
+  const j = api.jscodeshift;
+  const namedImports: Set<string> = new Set();
+  const root = j(fileInfo.source);
+  const importMatch = /wix-ui-tpa/;
+  const testkitsImportMatch = /wix-ui-tpa\/dist\/src\/testkit/;
+  const stylesImportMatch = /st.css/;
+  const driversImportMatch = /.driver/;
```
jonathanadler, line 10: Note that in this regex the `.` means "any character" and not a real `.`.
This should probably be escaped:
```suggestion
  const driversImportMatch = /\.driver/;
```

jonathanadler (commented):
Nice job!

Left some comments
@sivanharel would love to hear what you think as well

packages/wix-ui-codemod/src/wix-ui-tpa/__testfixtures__/import-components-via-index-file.input.js
```diff
@@ -0,0 +1,10 @@
+import { component } from 'some-library'; //unrelated import
+import { Button, PRIORITY } from 'wix-ui-tpa/Button';
+import { Badge, BADGE_PRIORITY } from 'wix-ui-tpa/Badge';
+import { component as c2 } from 'some-other-library'; //unrelated import
+import { IconButton as I, Skins as s } from 'wix-ui-tpa/IconBtton';
+import { Counter as counter, CounterSize as cs } from 'wix-ui-tpa/Counter';
+import { Dropdown, DropdownProps as dp, DROPDOWN_ALIGNMENT } from 'wix-ui-tpa/Dropdown';
+import { component as c3 } from 'some-other-new-library'; //unrelated import
```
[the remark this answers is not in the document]
sivanharel: Yes, actually it would be great to add these use cases.

packages/wix-ui-codemod/src/wix-ui-tpa/__testfixtures__/import-components-via-index-file.output.js
sivanharel: 1. I think we shouldn't add empty lines but it is not so critical.
2. It seems like it keeps the original order. It adds the import in the line of the first wix-ui-tpa import it finds.

bamai: 1. I’m not sure why it happens, maybe its the linter that we have in wix-ui-codemod (when I tested the codemod on bookings it didn't always add empty lines)
2. Yes, it keeps the original order of the imports and also the order of the imported names. But I see that in base-ui they sort the imported names before they create the new import, do we want to do this here as well?

packages/wix-ui-codemod/src/wix-ui-tpa/import-components-via-index-file.ts
```diff
@@ -0,0 +1,60 @@
+import { Transform, ImportSpecifier } from 'jscodeshift';
+
+const transformComponentsImports: Transform = (fileInfo, api) => {
+  const j = api.jscodeshift;
+  const namedImports: Set<string> = new Set();
+  const root = j(fileInfo.source);
+  const importMatch = /wix-ui-tpa/;
+  const testkitsImportMatch = /wix-ui-tpa\/dist\/src\/testkit/;
+  const stylesImportMatch = /st\.css/;
```
sivanharel, line 9:
```suggestion
  const stylesImportMatch = /\.st\.css/;
```

sivanharel (approved):
Looks good!

packages/wix-ui-codemod/src/wix-ui-tpa/__testfixtures__/import-components-via-index-file.output.js
sivanharel: Regarding the sort, I think it's a nice feature but it's not a must

sivanharel (approved):

PR #2288, opened 2021-11-17
examplesPreviewWrapper- new story feature (storybook-utils)

TalZa1810:


packages/wix-storybook-utils/src/LiveCodeExample/LiveCodeExample.tsx
```diff
@@ -35,6 +35,7 @@ export interface Props {
   previewWarning?({ onConfirm: Function }): React.ReactNode | null;
   hints?: ComponentsHints;
   forceInitialCode?: boolean;
+  playgroundWrapper?: Function;
```
argshook, line 38:
```suggestion
  playgroundWrapper?: StoryConfig['story']['playgroundWrapper'];
```

let's reuse the type from `StoryConfig`, you can get it from here:

```ts
import { StoryConfig } from '../typings/story-config';
```

packages/wix-storybook-utils/src/Sections/views/story-page.tsx
```diff
@@ -24,6 +26,7 @@ const examples = (
       source: examplesCode[item.example],
       compact: true,
       wide: item.wide,
+      playgroundWrapper: examplesConfig?.playgroundWrapper,
```
argshook, line 29: i think it would be nice to be able to configure `playgroundWrapper` also per-example and not only for the whole `.story.js` file.

For example, with this PR, if we ever have a case where 5 out of 6
playgrounds need a wrapper, there would be no way to remove it for
that one example.

Ability to configure per-example can be implemented in separate PR, but
maybe this idea requires a clearer name of `playgroundWrapper`?

Because `playgroundWrapper` is defined once for all code example
preview, maybe it should be named `codePreviewsWrapper` or something
like that?

wdyt?

argshook (commented):
Thanks for contributing!

packages/wix-storybook-utils/src/Sections/views/story-page.tsx
argshook: no need to implement it now, i'm just thinking if we ever find a need
for that, will then `playgroundWrapper` at the top level still make sense.

maybe it should be `previewWrapper`, because playground can be confused
with `playground()` section, which has its own `componentWrapper`, wdyt?

argshook (dismissed):
just the idea to rename `playgroundWrapper` to `previewWrapper`.

Otherwise all good, thanks for contributing!

packages/wix-storybook-utils/src/Sections/views/story-page.tsx
TalZa1810: another option is `examplesWrapper`. Than if we implement wrapper per example it can be named `exampleWrapper`. 
@mykas wdyt?

TalZa1810: ok, I've decided to change this to `examplesPreviewWrapper` to better emphasis what it does

env-github-token:
### 🤖 Hi, I'm #Janet!

Commit: **fa24de13028ed8d4337cda9cd56a126fac474e05**

* https://wix-pages.com/wix-storybook-utils.pr_2288/

PR #2290, opened 2021-11-22
wix-storybook-utils: Ability to hide import statement section from testkit tab

kobiburnley:


env-github-token:
### 🤖 Hi, I'm #Janet!

Commit: **e6c2e0673a103ff846711bb4494d9b0a8a65bcb6**

* https://wix-pages.com/wix-storybook-utils.pr_2290/

argshook (approved):
Personally i prefer state flags that are positive, so instead of
`hideImport` i'd do `showImport` (or rather more descriptive `showImportExample`) which has default value `true`.

This is almost purely a matter of taste, with one exception - `if`s are positive and arguably easier to grasp, for example:

```diff
-{ testkitConfig && !hideImport && ( ... ) }
+{ testkitConfig && showImportExample && ( ... ) }
```

wdyt?

kobiburnley:
> Personally i prefer state flags that are positive, so instead of `hideImport` i'd do `showImport` (or rather more descriptive `showImportExample`) which has default value `true`.
> 
> This is almost purely a matter of taste, with one exception - `if`s are positive and arguably easier to grasp, for example:
> 
> ```diff
> -{ testkitConfig && !hideImport && ( ... ) }
> +{ testkitConfig && showImportExample && ( ... ) }
> ```
> 
> wdyt?

I like to think about the default behavior first, and then name the flag according to the non-default behavior.

So instead of `showImport={false}` with default value `true`, I prefer flags that their default value is `false`, in this case `hideImport={true}` and default value is `false` (or undefined)

argshook (approved):
@kobiburnley thanks for contributing!

Let me know if i should merge or you'd like to (and have time) to
explore render function (or other) ideas.

kobiburnley:
> @kobiburnley thanks for contributing!
> 
> Let me know if i should merge or you'd like to (and have time) to explore render function (or other) ideas.

 I think we can merge and probably document the possible improvement in a github issue or something similar

PR #2292, opened 2021-11-23
wix-storybook-utils: Prettier babel issue

kobiburnley:
`prettier` format [started breaking](https://pullrequest-tc.dev.wixpress.com/viewLog.html?buildId=21421289&buildTypeId=FedInfra_Package_WixStorybookUtils&tab=buildLog&_focus=734#_state=609) on latest babel version. See this [test failure](https://pullrequest-tc.dev.wixpress.com/viewLog.html?buildId=21421289&buildTypeId=FedInfra_Package_WixStorybookUtils&tab=buildLog&_focus=734#_state=609)

I didn't dive into the problem but I found that prettier 2 comes with built in parsers that we can use.

This PR is a quick fix that uses the built-in prettier babel parser

### Relevant Test Log
```
[13:48:33]	[npm run test] Summary of all failing tests
[13:48:33]	[npm run test] FAIL src/LiveCodeExample/doctor-code.spec.ts
[13:48:33]	[npm run test]   ● formatCode › should format import statement
[13:48:33]	[npm run test] 
[13:48:33]	[npm run test]     expect(received).toBe(expected) // Object.is equality
[13:48:33]	[npm run test] 
[13:48:33]	[npm run test]     - Expected
[13:48:33]	[npm run test]     + Received
[13:48:33]	[npm run test] 
[13:48:33]	[npm run test]     - import { Thing } from 'thong';
[13:48:33]	[npm run test]     + import { value Thing } from 'thong';
[13:48:33]	[npm run test]       ↵
[13:48:33]	[npm run test] 
[13:48:33]	[npm run test]       55 |   it('should format import statement', () => {
[13:48:33]	[npm run test]       56 |     const code = formatCode("import { Thing } from 'thong';\n");
[13:48:33]	[npm run test]     > 57 |     expect(code).toBe("import { Thing } from 'thong';\n");
[13:48:33]	[npm run test]          |                  ^
[13:48:33]	[npm run test]       58 |   });
[13:48:33]	[npm run test]       59 | });
[13:48:33]	[npm run test]       60 |
[13:48:33]	[npm run test] 
[13:48:33]	[npm run test]       at Object.<anonymous> (src/LiveCodeExample/doctor-code.spec.ts:57:18)
```


env-github-token:
### 🤖 Hi, I'm #Janet!

Commit: **5521cb71f0fab9ffdfdf8e87941e364994a5f8a8**

* https://wix-pages.com/wix-storybook-utils.pr_2292/

packages/wix-storybook-utils/src/LiveCodeExample/doctor-code.tsx
```diff
@@ -55,14 +56,7 @@ export const formatCode = (code: string) => {
...
         {
-          parsers: {
-            babel: {
-              astFormat: 'estree',
-              parse: babelParse,
-              locStart,
-              locEnd,
-            },
-          },
+          parsers,
```
argshook, line 59: it's been a while so my memory is rusty, but i think the reason for specific parser config (instead of using `prettier/parser-babel` was to prevent babel appearing in the bundle twice.

Maybe today things are different and they're treeshakeable, but it wasn't the case back when i added these lines.
The functions in `doctor-code.tsx` end up in the browser and bundling the whole babel really bloats the bundle. So we should be careful and try to avoid bundling more than we need

kobiburnley: I still see a bug that should be fixed, maybe not using my approach which is just a suggestion, see the bug here: https://pullrequest-tc.dev.wixpress.com/viewLog.html?buildId=21421289&buildTypeId=FedInfra_Package_WixStorybookUtils&tab=buildLog&_focus=734#_state=609

argshook: Thanks, i'm aware of the issue, just want to point out, that using parser like this might have negative impact

kobiburnley: Since this breaks our documentation website we currently use this solution  (prettier built-in parser)
We'll keep using it until the fix, do you have some plans / ideas for fixing it?

argshook: hey, i tried naively to just upgrade puppeteer and not change anything about parser and it seems to be working 

https://github.com/wix/wix-ui/pull/2296 build fails on another thing, but `doctor-code.spec.ts` passes.

maybe there's something i'm missing, some other use case that breaks?

you say it breaks your documentation site, could you show me an example?

kobiburnley: prettier 1 is currently broken, so I guess bumping it in your PR solved it

argshook: yep, upgrading prettier is not an issue and i'd like that to happen, just want to confirm that changing parser is not required for your use case

PR #2294, opened 2021-11-24
<LinearProgressBar/> - make it clickable

Lukermelesh:


packages/wix-ui-core/src/components/linear-progress-bar/LinearProgressBar.tsx
```diff
@@ -144,9 +158,10 @@ export const LinearProgressBar: React.FunctionComponent<LinearProgressBarProps>
         {...getAriaAttributes(props)}
         data-min={_props.min}
         data-error={error}
-        role="progressbar"
+        role={onClick ? '' : 'progressbar'}
```
Lukermelesh, line 161: Maybe I should make it a prop instead?

packages/wix-ui-core/src/components/linear-progress-bar/LinearProgressBar.tsx
```diff
@@ -66,12 +70,16 @@ const resolveIndicationElement = (props: LinearProgressBarProps) => {
...
+  isButton: boolean = false,
+) => {
   const progressWidth = { width: `${value}%` };
+  const Component = isButton ? 'button' : 'div';
   return (
-    <div
+    <Component
       data-hook={ProgressBarDataHooks.container}
-      className={classes.barContainer}
+      className={st(classes.barContainer, { isButton })}
```
jonathanadler, line 82: It's actually best to use stylable states only on the root element, it might cause some issues otherwise.
In any case, I think it might not be needed in this case

packages/wix-ui-core/src/components/linear-progress-bar/LinearProgressBar.st.css
```diff
@@ -18,11 +18,17 @@
...
   width: 100%;
   position: relative;
   height: 100%;
 }
 
+.barContainer:isButton {
+  border: none;
+  padding: 0;
+}
+
```
jonathanadler, line 31: Actually I would just use a native `button` selector 😊 
Like this:
```suggestion
button.barContainer {
  border: none;
  padding: 0;
}

```

packages/wix-ui-core/src/components/linear-progress-bar/LinearProgressBar.tsx
```diff
@@ -144,9 +158,10 @@ export const LinearProgressBar: React.FunctionComponent<LinearProgressBarProps>
         {...getAriaAttributes(props)}
         data-min={_props.min}
         data-error={error}
-        role="progressbar"
+        role={onClick ? '' : 'progressbar'}
         className={st(classes.root, { error, success }, className)}
         {...filterDataProps(props)}
+        onClick={onClick}
```
jonathanadler, line 164: The `onClick` has to be on the `button` element for a11y reasons.
So either the root element should be a button, or the `onClick` should be moved to the button.

packages/wix-ui-core/stories/LinearProgressBar/index.story.tsx
```diff
@@ -16,9 +16,10 @@ export default {
     value: 10,
     showProgressIndication: false,
     error: false,
+    onClick: () => alert('I was clicked!'),
   },
   exampleProps: {
-    customSuffixIndicationText: [{label: 'suffix text', value: '35'}],
-    prefixIndication: [{label: 'prefix', value: '5 stars'}],
+    customSuffixIndicationText: [{ label: 'suffix text', value: '35' }],
+    prefixIndication: [{ label: 'prefix', value: '5 stars' }],
```
jonathanadler, line 23: I think we might need to allow both types of usages, so for it should be something like this:

```ts
const onClickOptions = [
  { value: () => alert('I was clicked!'), label: 'with onClick handler' },
  { value: undefined, label: 'without an onClick handler' },
];
...
componentProps: {
  ...
  onClick: onClickOptions[0].value,
},
exampleProps: {
  ...
  onClick: onClickOptions,
  ...
}
```

jonathanadler (commented):
Looks pretty good 💪 

I left some comments, and I still need to check with @NeilWix regarding the a11y of this feature

packages/wix-ui-core/src/components/linear-progress-bar/LinearProgressBar.st.css
Lukermelesh: Makes sense. I'll change it

packages/wix-ui-core/src/components/linear-progress-bar/LinearProgressBar.tsx
Lukermelesh: Why does it make a difference for accessibility? The only difference here is the click event gets propagated, but it shouldn't be any different to the end user?
Making the root element a button is tricky, because I have to use `display: contents`, in ReviewSummary to get the layout we want.
Adding `onClick` only on the button makes only the progress bar clickable, which might be too small of an area.

Maybe the solution could be to add onClick on both the button and the root element?

packages/wix-ui-core/stories/LinearProgressBar/index.story.tsx
Lukermelesh: It doesn't seem to be working this way. Example where it is done in exactly the same way: [storybook](https://www.wix-pages.com/wix-ui-tpa/?activeTab=Playground&path=%2Fstory%2Fcomponents-components--event) [github](https://github.com/wix-private/wix-design-systems/blob/5c7294aa4ec3efd2a1f1ea35c800917d6dc3ef78/packages/wix-ui-tpa/src/components/Event/docs/index.story.tsx#L44)

packages/wix-ui-core/src/components/linear-progress-bar/LinearProgressBar.tsx
Lukermelesh: For the record: we decided to leave it as is

packages/wix-ui-core/src/components/linear-progress-bar/LinearProgressBar.driver.ts
```diff
@@ -106,6 +113,18 @@ export const linearProgressBarDriverFactory: DriverFactory<LinearProgressBarDriv
...
+        getElement(ProgressBarDataHooks.progressIndicator).hasAttribute(
+          ProgressBarInnerAriaKeys.hidden,
+        ),
+      getAriaLabel: () =>
+        getElement(ProgressBarDataHooks.container).getAttribute(
+          ProgressBarInnerAriaKeys.label,
+        ),
       click: () => {
         const el = getElement(ProgressBarDataHooks.container) as HTMLElement;
         el.click();
```
sivanharel: After you moved the `onclick` event to the root wrapper of the component you could use the `BaseDriver` default `onclick` method.

sivanharel (commented):
It looks really good!
I just wrote 2 minor comments

packages/wix-ui-core/src/components/linear-progress-bar/LinearProgressBar.tsx
```diff
@@ -66,12 +73,18 @@ const resolveIndicationElement = (props: LinearProgressBarProps) => {
   );
 };
 
-const renderBarSection = (value: number | string) => {
+const renderBarSection = (
+  value: number | string,
+  isButton: boolean = false,
+  label?: string,
```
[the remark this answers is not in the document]
Lukermelesh: This is an internal function parameter. Maybe you mean `ariaLabel`?

packages/wix-ui-core/src/components/linear-progress-bar/LinearProgressBar.driver.ts
Lukermelesh: Not sure what you mean. `BaseDriver` is an interface with no `onclick` method.

Lukermelesh: Maybe you mean I should use `click: () => eventTrigger.click(element)`?

packages/wix-ui-core/src/components/linear-progress-bar/LinearProgressBar.tsx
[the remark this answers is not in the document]
sivanharel: ah you are right so yeh, `ariaLabel`

sivanharel (approved):
💪

PR #2298, opened 2021-12-04
wix-stortbook-utils: hideCodeEditor & componentsLinks

TalZa1810:
This feature is used to hide code editor and only displaying the examples themselves, without the code editor next to it.


packages/wix-storybook-utils/stories/StoryStructure/StoryStructure.story.js
```diff
@@ -40,6 +40,7 @@ export default {
     ],
     playgroundComponents: (globalComponents) => ({...globalComponents,  StyleComponent}),
     examplesPreviewWrapper: ({ component }) => <div className="story-class">{component}</div>,
+    hideCodeEditor: true,
```
TalZa1810, line 43: move this per example (content js file)

PR #2300, opened 2021-12-05
wix-ui-tpa script to run all migration codemods

shachar-almagor:


packages/wix-ui-codemod/src/wix-ui-tpa/migrate-wut-3.ts
```diff
@@ -1,14 +1,14 @@
...
 
-const wut3Codemods = {
+const wut3MigrationCodemods = {
   'replace-old-icon-imports': 'wix-ui-tpa/replace-old-icon-imports',
   // TODO: 'types-rename-migration': 'wix-ui-tpa/types-rename-migration',
   'import-components-via-index-file':
     'wix-ui-tpa/import-components-via-index-file',
 };
 
 export const executeWut3MigrationCodemods = (pathToBeMigrated: string) => {
```
sivanharel, line 10: Should we also support getting options?
Such as -p (print) snd -d (dry)?
I am not sure we need I am just raising it up :)

sivanharel (dismissed):
Looks good!
Just one question..

packages/wix-ui-codemod/src/wix-ui-tpa/migrate-wut-3.ts
shachar-almagor: Good question..I think that if we will provide options, we will need to document it / explain how to use them

PR #2307, opened 2021-12-08
Simplify wix-ui-icons-common build script

bamai:
https://jira.wixpress.com/browse/DSM-3423


packages/wix-ui-icons-common/scripts/build-icons.js
```diff
@@ -0,0 +1,32 @@
...
+  try {
+    spawnSync('svg2react-icon', ['--typescript', keepIconsColors, fromPath, toPath] );
+  }
+  catch(err){
+    error = true;
+    console.log(err);
+  }
+  if(!error){
+    console.log("finished icons build for " + libraryPath);
+  }
```
jonathanadler, line 18: Just a small code style remark.
This can be written like this:
```suggestion
  try {
    spawnSync('svg2react-icon', ['--typescript', keepIconsColors, fromPath, toPath] );
    console.log("finished icons build for " + libraryPath);
  }
  catch(err){
    console.log(err);
  }
```
just to spare the addition of the `error` variable 😁

packages/wix-ui-icons-common/scripts/build-icons.js
```diff
@@ -0,0 +1,32 @@
+const spawnSync = require('child_process').spawnSync;
+const path = require('path');
+
+const buildIconsFromLibrary = (libraryPath, keepColors) => {
+  const fromPath = path.join(libraryPath, '/raw');
+  const toPath = path.join(libraryPath, '/dist');
+  const keepIconsColors = keepColors ? '--keep-colors' : '--monochrome';
+  let error = false;
+  try {
+    spawnSync('svg2react-icon', ['--typescript', keepIconsColors, fromPath, toPath] );
```
jonathanadler, line 10: Just a thought, maybe we can use `spawn` instead of `spawnSync` then the build will be done in parallel, which will make it much faster.
The caveat is that we then need to deal with async processes, and stuff like that.
For instance we need to make sure the next npm script called, will be only after all spawned processes end, and perhaps end all processes when a developer presses `ctrl + c` in the middle of the process, etc.

btw, in order for the developer that will run build to see the output of this script we'll need to pass as a third argument the stdio option like this: `{ stdio: 'inherit' }`

packages/wix-ui-icons-common/scripts/create-aliases.js
```diff
@@ -1,35 +1,49 @@
...
-  fse.removeSync(to);
-  fse.ensureDirSync(to);
-}
-fse
-  .readdirSync(from)
-  .filter(file => file.endsWith(".js"))
-  .forEach(file => {
-    const fileName = file.split(".").shift();
-    const tsFile = `${fileName}.d.ts`;
+const createAliasesFromLybrary = (from, to, indexPath) => {
```
jonathanadler, line 4: 😊 
```suggestion
const createAliasesFromLibrary = (from, to, indexPath) => {
```

packages/wix-ui-icons-common/scripts/create-aliases.js
```diff
@@ -1,35 +1,49 @@
...
+  { "from": "./dist/src/classic-editor/system/dist/components", "to": "classic-editor/system", "indexPath": "./dist/src/classic-editor/system/dist" },
+];
 
-if (indexPath) {
-  fse.writeFileSync(
-    `${to}/index.js`,
-    `module.exports = require('./${path.relative(to, indexPath)}');\n`
-  );
-}
+iconLibraries.map(library => createAliasesFromLybrary(library.from, library.to, library.indexPath));
```
jonathanadler, line 49: Same comments are relevant here as well

jonathanadler (commented):
Awesome!!! ⭐ 

I left some minor comments, but in general it looks very good.

Great work!

nitayneeman (approved):
Great job!

PR #2322, opened 2021-12-19
wix-ui-core - add perdeepsmeta for optional dependencies

mykas:
This PR is adding ambassador dependencies and react to optional peer depenendencies. Should solve issues for users that use `npm > 6`


mykas:
@jonathanadler this should solve the whole chain not working for open sources users that uses npm > 6. I also added react to peer deps as it was not there 🤔

jonathanadler:
Looks good I guess

What issues where users getting?

shlomitc (approved):

mykas:
> Looks good I guess
> 
> What issues where users getting?

So basically if you use npm > 6 then npm will install all peerDependencies regardless if you wanted them to be optional or not. And thats bad for our open source people because they cant access those ambassador deps. 

 You have to do this meta thing to bring back the optionality.

PR #2324, opened 2021-12-21
fix(Checkbox): Update the `srOnly` styles according to Neil guidelines and use it from the checkbox component

sivanharel:
Update the `srOnly` styles according to Neil guidelines and use it from the checkbox component

See this task: https://jira.wixpress.com/browse/DSM-3280


jonathanadler (approved):
This has the potential to break although it's super correct

PR #2354, opened 2022-02-09
Keep listening on `onMouseLeave` when `Tooltip` is `disabled`

amitdahan:
# Summary

This PR fixes an issue where a `<Tooltip />` changes from `disabled={true}` to `disabled={false}` when the mouse is already outside of the tooltip's trigger - it'll re-appear as `shown` when enabled.

This was due to `onMouseLeave` not being handled when the `Tooltip` is `disabled`.


mykas (dismissed):
Nicely done!

PR #2356, opened 2022-02-09
Fix `@wix/ambassador-wix-atlas-service-web/http has no exported member 'V2GetPlaceResponse'`

jonathanadler:
Update type name to the new name `GetPlaceResponse`


EgoziE (dismissed):

EgoziE:
I thought we fixed that already....

EgoziE (approved):

